PR #148, opened 2023-10-25
Chore: Config Updates

AlexZorkin:


prv-proton (approved):
is it fine to run prettier to align common coding styles applied across? For some reason I get a blank screen though components are loaded in the login screen probably css issue which can be fixed later in the implementation ticket.

AlexZorkin:
> is it fine to run prettier to align common coding styles applied across? For some reason I get a blank screen, though components are loaded in the login screen probably css issue which can be fixed later in the implementation ticket.

yeah its just scss, I will clean up the existing site over the next few days.

Issue #194, opened 2023-11-10
LCFS - Add BCeID and IDIR roles to the corresponding database table(s)


justin-lepitzki:
**Describe the task**
Add BCeID and IDIR roles to the corresponding database table(s). 

**IDIR (Government) Roles:** https://preview.uxpin.com/59145260c7d8cc48969ac79401cc03d84843c3d9#/pages/164316070/simulate/no-panels

Administrator — can add/edit IDIR users and assign roles and add/edit organizations, BCeID users and assign roles
Analyst — can make recommendations on transfers, transactions and compliance reports, manage file submissions and add/edit fuel codes
Compliance Manager — can make recommendations on compliance reports
Director — can assess compliance reports and approve transactions

**BCeID Roles:** https://preview.uxpin.com/59145260c7d8cc48969ac79401cc03d84843c3d9#/pages/164314016/simulate/no-panels

Manage Users — can add/edit BCeID users and assign roles
Transfer — can create/save transfers and submit files
Compliance Reporting — can create/save compliance reports and submit files
Signing Authority — can sign and submit compliance reports to government and transfers to trade partners/government
Read Only — can view transactions, compliance reports and files

**Purpose**
Roles are fundamental to access control within the system

**Acceptance Criteria**
- [ ] first
- [ ] second
- [ ] third

**Additional context**
- Add any other context about the task here.
- Or here

AlexZorkin:
Hey team! Please [add your planning poker estimate](https://app.zenhub.com/workspaces/Low-Carbon-Fuel-Standard-644c2db3be098a001977ad0b/issues/bcgov/lcfs/194?planning-poker) with Zenhub @hamed-valiollahi @jig-patel @kevin-hashimoto @prv-proton

justin-lepitzki:
Nothing for PO to test

Issue #185, opened 2023-11-09
LCFS - Implement Data Fixtures for Local Development Environment


AlexZorkin:
**Describe the task**
Implement a system for loading data fixtures into the application's database on startup when running in a local development environment. This should be triggered either via the `prestart.sh` script or as part of the FastAPI application's startup event. The fixtures will populate the database with a predefined set of data that will assist developers in local testing and development.

**Purpose**
The absence of initial data in a fresh instance of the application complicates the development process, as developers need to manually insert data to perform local testing. Automating this process by introducing data fixtures will streamline the setup of a local development environment, reduce time spent on repetitive tasks, and improve overall developer productivity. It also ensures consistency of test data across different local environments.

**Acceptance Criteria**
- [ ] Data fixtures are defined in a Python script or module with pre-defined records for each required model.
- [ ] The fixture-loading functionality is triggered automatically during the application's startup when the database is empty.
- [ ] The fixture-loading process is idempotent, ensuring it can run multiple times without duplicating data.
- [ ] Integration of the fixture-loading script into the `prestart.sh` script or the FastAPI startup command, guarded by an environment variable that specifies the application is in a development mode.
- [ ] Documentation is updated to reflect the new process for initializing the development environment with data fixtures.

**Additional context**
- Consider using FastAPI's event handlers like `startup` to check and execute the data loading script.
- Ensure that loading data fixtures is only enabled for local development, not in production settings.
- Here is a simple Python code example using SQLAlchemy models and FastAPI's event handling:

```python
from fastapi import FastAPI, Depends
from sqlalchemy.orm import Session
from . import models
from .database import SessionLocal, engine

models.Base.metadata.create_all(bind=engine)

app = FastAPI()

def get_db():
    db = SessionLocal()
    try:
        yield db
    finally:
        db.close()

def init_db(db: Session = Depends(get_db)):
    # Check if the database is empty, and if so, insert data fixtures
    if db.query(models.YourModel).count() == 0:
        # Assuming YourModel is a SQLAlchemy model and you have a list of dicts with the data
        db.bulk_insert_mappings(models.YourModel, [
            {'name': 'Item1', 'description': 'Description of item 1'},
            {'name': 'Item2', 'description': 'Description of item 2'},
            # ... more data
        ])
        db.commit()

@app.on_event("startup")
async def startup_event():
    with SessionLocal() as db:
        init_db(db)

# ... rest of your FastAPI app code

prv-proton:
[...]

PR #205, opened 2023-11-16
Feat: user roles data model update #194

kevin-hashimoto:
- adds IDIR and bceid user roles

note: changed down revision id on one of the migration files so that autogenerate can run.

not sure if we should include a seed script to populate the DB but here is the sql script:

```
INSERT INTO role (name, description, is_government_role, display_order)
VALUES
    ('Administrator', 'Can add/edit IDIR users and assign roles, add/edit organizations, BCeID users, and assign roles', true, 1),
    ('Analyst', 'Can make recommendations on transfers, transactions, and compliance reports, manage file submissions, and add/edit fuel codes', true, 2),
    ('Compliance Manager', 'Can make recommendations on compliance reports', true, 3),
    ('Director', 'Can assess compliance reports and approve transactions', true, 4),
    ('Manage Users', 'Can add/edit BCeID users and assign roles', false, 5),
    ('Transfer', 'Can create/save transfers and submit files', false, 6),
    ('Compliance Reporting', 'Can create/save compliance reports and submit files', false, 7),
    ('Signing Authority', 'Can sign and submit compliance reports to government and transfers to trade partners/government', false, 8),
    ('Read Only', 'Can view transactions, compliance reports, and files', false, 9);
```


AlexZorkin:
Manually added changes from this pr to avoid conflicts.

AlexZorkin:
We will add the insert statement into the fixtures ticket #185

PR #234, opened 2023-11-28
Feat: edit existinig users 178

kevin-hashimoto:
- Edit users page @ `/users/1/edit` route
- `crumb` param on `Layout` component - Can now opt in/out of using crumbs while using `Layout` component universally

notes: 
- we should probably determine url structure at some point (eg.  if the /users route is shared between IDIR and BCeID users or if they get separate routes: /org/users and /admin/users etc. )


hamed-valiollahi:
Nice work, Kevin! Alex asked me to review this PR. Reviewing your code was a great learning experience for me. Overall, it's excellent, but I have a few minor suggestions that I'll add here shortly. Feel free to consider them or not!

frontend/src/views/editUser/EditUser.jsx
```diff
@@ -0,0 +1,402 @@
...
+                      control={<Radio />}
+                      label="Director — can assess compliance reports and approve transactions"
+                    />
+                  </RadioGroup>
+                </>
+              ) : (
+                <Stack spacing={1}>
+                  <FormControlLabel
+                    control={<Checkbox />}
+                    label="Manage Users — can add/edit BCeID users and assign roles"
```
hamed-valiollahi, line 361: In the wireframe, the role titles are bold, but their descriptions are not. For example:

> **Manage Users** — can add/edit BCeID users and assign roles

frontend/src/views/editUser/EditUser.jsx
```diff
@@ -0,0 +1,402 @@
...
+                    onChange={handleChange}
+                    value={formData.email}
+                  />
+                  {dummy.errors.email && (
+                    <FormHelperText error>{dummy.errors.email}</FormHelperText>
+                  )}
+                </div>
+                <div>
+                  <Typography fontSize={16}>
+                    Alternate email for notifications (optional)
```
hamed-valiollahi, line 207: In the wireframe, the field titles are bold, but "(Optional)" is not. For example:

> **Alternate email for notifications** (optional)

frontend/src/views/editUser/EditUser.jsx
```diff
@@ -0,0 +1,402 @@
...
+                    checked={formData.complianceReporting}
+                  />
+                  <FormControlLabel
+                    control={<Checkbox />}
+                    label="Signing Authority — can sign and submit compliance reports to government and transfers to trade partners/government"
+                    onChange={handleCheckbox}
+                    name="signingAuthority"
+                    checked={formData.signingAuthority}
+                  />
+                  <FormControlLabel
```
hamed-valiollahi, line 387: I understand it can be confusing! In the wireframe, it is actually a Radio Button. Clicking on it will uncheck all the checkboxes above.

Please review the wireframe, click on the Radio Button, and then follow the arrows.

frontend/src/views/editUser/EditUser.jsx
kevin-hashimoto: Mui limitation

frontend/src/App.js
```diff
@@ -66,18 +62,36 @@ const router = createBrowserRouter([
...
             handle: {
-              crumb: data => `Users`, // data from loader is passed into our crumb function so we can manipulate the output
+              crumb: () => `Users`, // data from loader is passed into our crumb function so we can manipulate the output
             },
           },
         ],
       },
     ],
   },
+  {
```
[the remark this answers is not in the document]
kevin-hashimoto: edit users will not have breadcrumbs so it can stay here. the above route group is for layouts that need breadcrumbs as denoted by the `crumb` parameter for `<Layout/>`.

The route will change ( idir/bceid ) according to your other comment, we will add those routes here.

frontend/src/App.js
```diff
@@ -8,6 +8,7 @@ import { ViewUsers } from './views/viewUsers';
 import { Paper } from '@mui/material';
 import { BCTypography } from 'components/BCTypography';
 import UserTabPanel from 'layouts/admin/UserTabPanel';
+import { EditUser } from './views/editUser';
```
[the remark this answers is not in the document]
kevin-hashimoto: It doesnt make much sense to repeat our work ( the layout is the same other than a couple fields ). It would be more beneficial in my opinion set a flag for what fields to show as is already done in `EditUser.jsx`. 

The folders dont necessarily have to reflect the route structure, as that is already done via the routes object that gets passed into react router. This is one of the questions i was planning to ask the devs for our upcoming tooling and standards update.

For now as we already have the pieces to switch between idir and bceid user views within the page component, I think we can leave this for now unless there are bigger conerns.

Thoughts?

frontend/src/layouts/Layout.jsx
```diff
@@ -42,24 +45,30 @@ const Layout = () => {
...
-              </Breadcrumbs>
-            </Paper>
+            {crumbs && (
+              <Paper
+                p={2}
+                elevation={5}
+                sx={{ padding: '1rem', minHeight: '5vh' }}
+              >
+                <Breadcrumbs separator={<NavigateNextIcon fontSize="small" />}>
+                  {breadcrumbs.map((crumb, index) =>
```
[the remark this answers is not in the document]
kevin-hashimoto: As we set up and add the Organizations and [Organization] pages to the routes, the breadcrumbs should be updated accordingly.

Since we dont have those pages yet, and that these crumbs will be added as those pages come, I think we can leave this for now and adjust if needed when those pages are made.

thoughts?

frontend/src/views/editUser/EditUser.jsx
```diff
@@ -0,0 +1,402 @@
...
+            <BCButton
+              variant="outlined"
+              color="dark"
+              style={{
+                background: 'white',
+                alignItems: 'center',
+                justifyContent: 'space-between',
+                gap: 8,
+              }}
+              onClick={() => navigate(-1)}
```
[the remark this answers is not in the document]
kevin-hashimoto: either or would work i believe. if we copy this page component to the two routes ( idir and bceid) like you suggested above, we can add the route to the `onClick`. If we decide to use the page component for both routes, we can have a `handleBackClick` function that will check what type of user they are before routing them to the proper route

frontend/src/views/editUser/EditUser.jsx
hamed-valiollahi: It should work:

```html
<FormControlLabel
  control={<Checkbox />}
  label={
    <span>
      <strong>Manage Users</strong> — can add/edit BCeID users and assign roles
    </span>
  }
  onChange={handleCheckbox}
  name="manageUsers"
  checked={formData.manageUsers}
/>
````

kevin-hashimoto: good find 👍

hamed-valiollahi (approved):

PR #277, opened 2023-12-18
Feat: Add login control for inactive users - 191

### LINKED ISSUE #191: LCFS - Inactive users are unable to login
**Description:**
Users that are set as inactive (done via the edit user page) are not able to login to the application. 

**Wireframe:** <link>

User access control for BCeID and IDIR users

**Acceptance Criteria:**

- [ ] Given I am a <role>, When I am <doing something>, then <this happens>
- [ ] Given I am a <role>, When I am <doing something>, then <this happens>

**Development Checklist:**
- [ ] A
- [ ] B
- [ ] C

**Notes:**

hamed-valiollahi:
This PR introduces a new feature that allows the application to control login access for users marked as inactive.

Closes #191


AlexZorkin (changes requested):
Code looks good but we need to present an error to the user who is trying to login. Currently the app will just return to the login screen but it would be nice to show an error alert on that page when login fails.

- "Given I am an inactive BCeID user, When I am attempting to log in, then I am presented with an error message and will not be able to log in"

hamed-valiollahi:
> Code looks good but we need to present an error to the user who is trying to login. Currently the app will just return to the login screen but it would be nice to show an error alert on that page when login fails.
> 
> * "Given I am an inactive BCeID user, When I am attempting to log in, then I am presented with an error message and will not be able to log in"

Thank you, Alex, for your feedback. I'll create a new PR for the requested changes.

PR #284, opened 2023-12-21
feat: Enhancing Data Management: Server-Side Pagination, Filtering, a…

prv-proton:
Enhancing Data Management: Server-Side Pagination, Filtering, and Loading for AG-Grid
- added pagination, sorting and filtering ability to admin/users endpoint at the backend
- generic component to render the table from server end using ag-grid.
- input text field, checkbutton and other theme & styling changes 

note: created some demo components to assist developers, which can be removed at later stages.


frontend/src/components/BCGrid/BCGridServer.jsx
```diff
@@ -0,0 +1,202 @@
...
+  const [rowData, setRowData] = useState([])
+  const [isLoading, setIsLoading] = useState(true)
+  const [isError, setIsError] = useState(false)
+  const [error, setError] = useState('')
+
+  const apiService = useApiService()
+  const fetData = useCallback(
+    () =>
+      apiService({
+        method: 'post',
```
AlexZorkin, line 59: I think this will have to be an option. Currently the view is trying to post to the organizations/list get method.  Are you thinking this will always be post? The organization list endpoint will have to be altered.

backend/lcfs/web/api/user/session.py
```diff
@@ -67,78 +48,87 @@ async def query_users(
...
-        user_results = await self.session.execute(query)
-        results = user_results.scalars().all()
+        user_results = await self.session.execute(query.offset(offset).limit(limit))
+        # total_count = query.count()
+        results = user_results.scalars().unique().all()
 
         # Convert the results to UserBase schemas
-        return [UserBase.model_validate(user) for user in results]
+        return [UserBase.model_validate(user) for user in results], total_count
+# TODO: Need to redine search, create and modify endpoints
```
AlexZorkin, line 89: Does this need to be done now?

backend/lcfs/web/api/user/session.py
```diff
@@ -67,78 +48,87 @@ async def query_users(
...
-        Returns:
-            UserBase: The found user's data converted to a UserBase Pydantic model. If no user is found, returns None.
-        """
-        stmt = f'{USER_VIEW_STMT} and u.is_active = true and u.id = :user_id'
-        
-        user_results = await self.session.execute(text(stmt), {'user_id': user_id})
-        
-        user = user_results.fetchone()
-        return row_to_dict(user, UserBase) if user else None
+    #     user_results = await self.session.execute(text(stmt), {"user_id": user_id})
```
AlexZorkin, line 105: cleanup?

backend/lcfs/web/api/user/views.py
```diff
@@ -44,100 +45,164 @@ async def get_current_user(request: Request, response: Response = None) -> UserB
...
+            return EntityResponse(
+                status=status.HTTP_404_NOT_FOUND,
+                message="Failed",
+                success=False,
+                data=[],
+                error={"message": "No roles found"},
+            )
+
+        # Return the list of UserBase objects
+        return EntityResponse(
```
AlexZorkin, line 89: will we always use EntityResponse for these pagination responses?

backend/lcfs/web/api/user/views.py
```diff
@@ -44,100 +45,164 @@ async def get_current_user(request: Request, response: Response = None) -> UserB
...
-async def create_user(user_data: UserCreate, db: AsyncSession = Depends(get_async_db)) -> EntityResponse:
-    try:
-        user_repo = UserRepository(db)
-        created_user = await user_repo.create_user(user_data)
-        return EntityResponse(status=status.HTTP_201_CREATED, data=created_user, error={}, success=True, message="User created successfully")
-    except Exception as e:
-        raise HTTPException(status_code=500, detail=f"Failed to create user: {str(e)}")
+# TODO: yet to redefine and test endpoints for below.
+# @router.get("/{user_id}", response_model=EntityResponse, status_code=status.HTTP_200_OK)
+# async def get_user_by_id(user_id: int, response: Response = None) -> EntityResponse:
```
AlexZorkin, line 155: Are these just not working currently?

frontend/src/views/AdminUsers/components/DemoButtons.jsx
AlexZorkin, line 1: We should rename this component to something related to the grid or users table.

AlexZorkin (changes requested):
Looking great, just a few comments to address.

frontend/src/components/BCGrid/BCGridServer.jsx
prv-proton: yes, since we need to pass the filter and sort models this needs to be changed to post.

backend/lcfs/web/api/user/session.py
prv-proton: yes, create and edit endpoints will add as we refine the UI, as this ticket's main focus was ag-grid.

backend/lcfs/web/api/user/views.py
prv-proton: If it's only pagination then we wouldn't need it, as it involves sort and filters this is more efficient I felt. If not we could go with any 3rd party libraries.

backend/lcfs/web/api/user/views.py
prv-proton: they were failing so just commented. Will make further changes as Admin page is refined along with the organization ticket that I've picked up currently.

frontend/src/views/AdminUsers/components/DemoButtons.jsx
prv-proton: This model is only for demo purpose only, I plan to remove it at later stage. As this model can be a guide to someone who is going to work on Compliance Report table where we've pre-defined filters.

Issue #162, opened 2023-10-27
LCFS - Organization information is visible on an Organization Details page 


justin-lepitzki:
**Description:**
Add organization information to Organization Details page 

Display the following elements for both BCeID and IDIR users:
- Legal name of organization
- Operating name of organization (if different than legal name)
- Telephone number of organization
- Address for service, including postal address in B.C. and email address 
- Address in B.C. at which records are maintained (if different than address for service)

Display the following elements for BCeID users only: 
- A disclaimer below the address fields that says, "Email lcfs@gov.bc.ca to update address information."

**Wireframe:** 

- https://preview.uxpin.com/59145260c7d8cc48969ac79401cc03d84843c3d9#/pages/164329808
- **NOTE:** This card does not include the Users functionality that is displayed in the above wireframe, including the New user button, show inactive users, and the user table. These features will be handled in separate cards.

**Purpose and benefit to user:**
Provides information about the organization. This information will then be displayed within a compliance report (future work) as this info is required by the legislation

**Acceptance Criteria:**

- [x] Given I am a BCeID user, When I enter Organization Details page, then I can view the organization information and a disclaimer below the address fields that instructs on how to update address information.
- [x] Given I am a IDIR user, When I enter Organization Details page, then I want to see organization information without the disclaimer regarding address updates.

**Development Checklist:**
- [ ] Design and create a React component to encapsulate the organization information section.
- [ ] Implement feature flag to show or hide the disclaimer based on the user's type (BCeID or IDIR).
- [ ] Integrate the organization information component into the Organization Details page.

jig-patel:
Hey team! Please [add your planning poker estimate](https://app.zenhub.com/workspaces/Low-Carbon-Fuel-Standard-644c2db3be098a001977ad0b/issues/bcgov/lcfs/162?planning-poker) with Zenhub @AlexZorkin @hamed-valiollahi @jig-patel @kevin-hashimoto @prv-proton

PR #285, opened 2023-12-21
Feat: display org data on org page #162

kevin-hashimoto:
- feat: updates the api route to fetch org data from the db
  - new schema made specifically for the route with a more scalable and modern approach. other org schemas left untouched in case it messes with what is already working.
- feat: replaces org dummy data with real data
- feat: `constructAddress` util function that takes in an address object ( same format as what is returned from backend ) and spits out a formatted full address


backend/lcfs/web/api/organization/views.py
```diff
@@ -5,28 +5,39 @@
...
+    GetOrganizationResponse,
+)
 from lcfs.web.core.decorators import roles_required
 
 logger = getLogger("organization")
 router = APIRouter()
 get_async_db = dependencies.get_async_db_session
 
+
 # TODO: Implement permission check for this route to ensure that
```
AlexZorkin, line 31: plz remove this comment, no longer needed.

backend/lcfs/web/api/organization/views.py
```diff
@@ -5,28 +5,39 @@
...
 # only authorized users can access it.
-@router.post("/", response_model=OrganizationSchema, status_code=status.HTTP_201_CREATED)
+@router.post(
+    "/", response_model=OrganizationSchema, status_code=status.HTTP_201_CREATED
+)
 @roles_required("Government", "Adminstrator")
 async def create_organization(
```
AlexZorkin, line 37: I'm getting a duplicate key error when trying to create a new organization. Are you able to create in your local environment?

backend/lcfs/web/api/organization/views.py
```diff
@@ -62,17 +72,57 @@ async def create_organization(
...
+
+        return org
+
+    except Exception as e:
+        raise HTTPException(status_code=500, detail=f"Internal Server Error: {str(e)}")
+
+
 @router.put("/{organization_id}", response_model=OrganizationSchema)
-async def update_organization(organization_id: int, organization_data: OrganizationUpdateSchema, db: AsyncSession = Depends(get_async_db)):
+async def update_organization(
```
AlexZorkin, line 109: We don't have an edit view for org yet, so just a note that we'll have to test this again once the view is created.

AlexZorkin (approved):
Looks good, just a few comments to address.

backend/lcfs/web/api/organization/views.py
kevin-hashimoto: I think prashanth is working on the edit org side of things in a separate ticket.

backend/lcfs/web/api/organization/views.py
kevin-hashimoto: I've been using the seeded data to test so I havnt run in to this issue. Ill test it out on my local

kevin-hashimoto: this should be fixed now. will need to delete and seed the db again. the sequence was off when seeding so it caused dupe keys

Issue #208, opened 2023-11-18
LCFS - Show Organization Details on BCeID dashboard


justin-lepitzki:
**Description:**
Show the organization details on BCeID dashboard:

- Show operating name of organization
- Show address for service
- Show organization telephone and email
- "Users" link to Organization page 
- "Create new BCeID account" links to [https://www.bceid.ca/](https://www.bceid.ca/)

**Wireframe:** https://preview.uxpin.com/59145260c7d8cc48969ac79401cc03d84843c3d9#/pages/164908212/simulate/no-panels

**Purpose and benefit to user:**
User is displayed their organization information

**Acceptance Criteria:**
- [ ] Given I am a BCeID user, when I access my dashboard, then I should see the organization's details listed above as well as Users and Create new BCeID account links.
- [ ] Given I am a BCeID user, when I click on the "Users" link, then it should redirect me to the Organization page.
- [ ] Given I am a BCeID user, when I click on the "Create new BCeID account" link, then it should direct me to [https://www.bceid.ca/](https://www.bceid.ca/)

**Development Checklist:**
- [ ] Implement a UI component to display the organization's details.
- [ ] Include the component to the dashboard page.

**Notes:**

justin-lepitzki:
@Kailee-Douglas @airinggov the wireframe I included is based on this mockup from UX pin (https://preview.uxpin.com/59145260c7d8cc48969ac79401cc03d84843c3d9#/pages/164349076/simulate/sitemap). Can we revise the mockup of the Organization Details portion of the BCeID dashboard to include the organization telephone number and email (these are new info requirements from the new Act). See org details page for example info: https://preview.uxpin.com/59145260c7d8cc48969ac79401cc03d84843c3d9#/pages/164329808/simulate/sitemap

Kailee-Douglas:
@justin-lepitzki  here is an updated mock-up for your review: https://preview.uxpin.com/59145260c7d8cc48969ac79401cc03d84843c3d9#/pages/164908212

justin-lepitzki:
Hey team! Please [add your planning poker estimate](https://app.zenhub.com/workspaces/Low-Carbon-Fuel-Standard-644c2db3be098a001977ad0b/issues/bcgov/lcfs/208?planning-poker) with Zenhub @hamed-valiollahi @kevin-hashimoto @jig-patel @prv-proton @AlexZorkin

justin-lepitzki:
[...]

PR #290, opened 2023-12-28
Feat: show org details on dashboard #208

kevin-hashimoto:
- chore: merged in #162 as data fetching is the same
- chore: updated packages
- refactor: swap v3 react-query with v5 react-query
- refactor: get queries moved to custom hooks inside new hooks directory
- feat: new `Role` component as an alternative to `withRole` that can wrap components for easier composability
- refactor: test wrapper moved to separate file so it can be used in multiple test components
- feat: dashboard now displays organization detail of the current user. This component should later be wrapped in the new `Role` component to only how for bceid users


AlexZorkin (approved):
A nice cleanup of some key components.

Issue #227, opened 2023-11-24
LCFS - View transfer screen


justin-lepitzki:
**Description:**
This ticket involves the development of the 'View Transfer' screen for the LCFS system. Once a draft transfer has been sent to the second organization (the transfer partner), the transfer's status changes to 'Sent' and becomes visible to both organizations involved in the transfer. This screen will display various details about the sent transfer, including the transfer ID, comments, transaction history, and the date of the written agreement. The design of this screen must reflect these changes and provide the necessary information in a clear and accessible manner.

Key Features:
- Transfer details are displayed but are no longer editable.
- The breadcrumb indicates the 'Sent' stage.
- A description of the pending transfer is prominently displayed.
- The transfer ID is visible at the top of the page.
- A comment section for users to view and add comments.
- A new transaction history section showing all relevant transfer activities.
- Date of the written agreement displayed in the transaction history.
- A record indicating "signed and sent" is visible in the transaction history.
- Wireframe reference: [UXPin Wireframe](https://preview.uxpin.com/59145260c7d8cc48969ac79401cc03d84843c3d9#/pages/165000525/simulate/sitemap).

**Purpose and Benefit to User:**
This view screen is essential for users to monitor and manage pending transfers. It serves as the central interface for users to understand the current status of their transfers and to perform subsequent actions based on their roles and the transfer's progress.

**Acceptance Criteria:**
- Given I am a user with viewing privileges, when I access the 'View Transfer' screen, then I should see the transfer's status as 'Sent' along with all the detailed information as described.
- Given I am a user with specific role privileges, when I view the 'View Transfer' screen, then the appropriate action buttons and functionalities should be available to me based on my role.

**Development Checklist:**
- A. Implement the 'View Transfer' screen as per the provided wireframe.
- B. Ensure transfer details are visible but not editable.
- C. Integrate the transaction history section with real-time data updates.

**Notes:**
- The development team should work closely with the UI/UX designers to ensure the screen layout matches the provided wireframe.
- Special attention should be given to user role management to ensure the correct display of buttons and functionalities.
- Testing should include verifying that all information displayed is accurate and up-to-date, especially in the transaction history section.

prv-proton:
[...]

PR #334, opened 2024-01-23
Feat: LCFS - View transfer screen #227

prv-proton:
- Transfer Details in view mode.
- Remove unwanted stylings applied by default.


frontend/src/views/Transactions/TransferDetailsView.jsx
```diff
@@ -0,0 +1,193 @@
...
+// Sub components
+import { OrganizationBadge } from './components/OrganizationBadge'
+import { demoData } from './components/demo'
+import { AttachmentList } from './components/AttachmentList'
+import { Comments } from './components/Comments'
+
+const idirSteps = ['Draft', 'Sent', 'Submitted', 'Recorded']
+
+export const TransferDetailsView = () => {
+  const navigate = useNavigate()
```
AlexZorkin, line 33: please add some commented out code to get the transfer details from the api. Use the orgView as an example

```js
export const ViewOrg = () => {
  const [showActive, setShowActive] = useState(true)
  const navigate = useNavigate()
  const { orgID } = useParams()
  const { data: currentUser, isLoading: isCurrentUserLoading } =
    useCurrentUser()
  const { data: orgData, isLoading } = useOrganization(orgID)
  const handleEditClick = () => {
    navigate(ROUTES.ORGANIZATIONS_EDIT.replace(':orgID', orgID), {
      state: { orgID, isEditMode: true }
    })
  }
```
please also add a useTransfer hook that prepares us to be able to get transfer data from the api

```js
import { useApiService } from '@/services/useApiService'
import { useQuery } from '@tanstack/react-query'
import { useCurrentUser } from './useCurrentUser'

export const useOrganization = (orgID, options) => {
  const client = useApiService()
  const { data: currentUser } = useCurrentUser()
  const id = orgID ?? currentUser.organization.organization_id

  return useQuery({
    queryKey: ['organization', id],
    queryFn: async () => (await client.get(`/organizations/${id}`)).data,
    ...options
  })
}

AlexZorkin (changes requested):

frontend/src/views/Transactions/TransferDetailsView.jsx
prv-proton: updated

Issue #306, opened 2024-01-12
LCFS - Refactor Organization Route Handlers to Use Repository Pattern and Dependency Injection


AlexZorkin:
**Describe the task:**  
Refactor existing organization route handlers in the FastAPI application to use the repository pattern and dependency injection. This task involves identifying route handlers that directly interact with the database and modifying them to use repository classes for database interactions.

**Purpose:**  
The goal is to enhance the application's architecture by separating concerns, thereby making the codebase more maintainable, scalable, and testable. This approach also streamlines database interaction code, making it more consistent across the application.

**Acceptance Criteria:**
- [ ] Refactor the `export_organizations` endpoint to use an `OrganizationRepository` instead of directly interacting with the database.
- [ ] Update the `get_organization` endpoint to retrieve organization details via the repository pattern.
- [ ] Modify the `update_organization` endpoint to handle data updates through an `OrganizationRepository`.
- [ ] Ensure the `list_organizations` endpoint already using the repository pattern remains unchanged.
- [ ] Verify the `get_organization_statuses` and `get_organization_types` endpoints that use the repository pattern do not require changes.
- [ ] Refactor the `get_users_for_organization` endpoint to fetch users via a repository instead of direct database calls.
- [ ] After refactoring, ensure all endpoints function as expected and pass all existing tests.
- [ ] Update any related documentation to reflect the new repository usage in route handlers.

**Additional context:**
- Test extensively to ensure that the refactoring does not introduce regressions.
- Consider the impact on any frontend components that consume these endpoints, ensuring that API contracts remain consistent.

Grulin:
Nothing to test on the business area side.

PR #356, opened 2024-01-31
refactor: organization route handlers #306

kevin-hashimoto:
#306 
Refactor org route handlers to follow repository pattern

- feat: layer decorators that handle try exccept
- feat: exception handling
- feat: custom exceptions
- refactor: org route handlers
- refactor: split `session.py` into `repo.py` and `services.py`

notes:
- this pattern would be good to follow for other routes as well
- custom exceptions should be created as use cases arise. They should also be handled in the layer handler decorators as well
- because try except is handled in the decorators, we do not need to write try except blocks anymore
- view - routing, service - business logic, repo - database querying


AlexZorkin (approved):
Great work on this. We will follow this pattern for all our other entities.

PR #361, opened 2024-02-03
Feat: Signing declaration added to Transfer

AlexZorkin:
- added new input checkbox on AddTransfer view that lets the user set the signed declaration button
- removed username from user_profile with migration


backend/lcfs/db/models/UserProfile.py
```diff
@@ -25,9 +25,6 @@ class UserProfile(BaseModel, Auditable):
         String(150), unique=True, nullable=False, comment="keycloak Username"
     )
     email = Column(String(255), nullable=True, comment="Primary email address")
-    username = Column(
-        String(150), unique=True, nullable=False, comment="Login Username"
-    )
     display_name = Column(
```
prv-proton, line 28: Can we remove display_name field as well?

frontend/src/views/Transfers/AddTransfer/components/SigningAuthority.jsx
```diff
@@ -0,0 +1,49 @@
+import React from 'react'
+import { Box, FormControlLabel, Checkbox, Typography } from '@mui/material'
+import LabelBox from './LabelBox' // Assuming LabelBox is already defined in your project
```
prv-proton, line 3: LabelBox is not used inside this component.

frontend/src/views/Transfers/AddTransfer/components/SigningAuthority.jsx
```diff
@@ -0,0 +1,49 @@
...
+import BCTypography from '@/components/BCTypography'
+import withRole from '@/utils/withRole'
+
+const SigningAuthority = ({ register, errors }) => {
+  return (
+    <>
+      <BCTypography variant={'h6'} mt={2} mb={2} color={'primary'}>
```
prv-proton, line 10: To be consistent with other components we can define the header inside LabelBox.
    <LabelBox
      label="Signing Authority Declaration"
    >

prv-proton (approved):

prv-proton (changes requested):
comments provided

frontend/src/views/Transfers/AddTransfer/components/SigningAuthority.jsx
AlexZorkin: the wireframe has a different styling for Signing Authority so that's why we left it out.

AlexZorkin:
This pr is included is the full transfer branch so closing this one.

Issue #160, opened 2023-10-27
LCFS - Create Organizations page for IDIR users 


justin-lepitzki:
**Description:**
Create Organizations page for IDIR users. 

- Nav bar and main heading: Organizations
- Can use Material UI table, should be same as other tables (e.g., user table) for consistency
- four columns: Organization name, Balance, In reserve, Registered
- Created organizations will be populated in this table, similar to TFRS
- Default sort is alphabetical
- Material UI table should be accessible in that you can tab through the rows & it shows up in the screen reader
- Implement breadcrumb navigation on this page

**Wireframe:** https://preview.uxpin.com/59145260c7d8cc48969ac79401cc03d84843c3d9#/pages/164769377

**Purpose and benefit to user:**
Will facilitate the management of organizations by government

**Acceptance Criteria:**

- [ ] Given I am a IDIR User, When I am clicking and Organizations page, then I should be able to see all the organizations details like name, balance of compliance units, compliance units that are in reserve and status of registration.
- [ ] Sorted based on Organization name in alphabetical order in ascending order.
- [ ] Newly created organization details should be updated as and when they're added.

**Development Checklist:**
- [ ] Create a view in UI to show the details of Organizations for IDIR user.
- [ ] Follow common table design after finalizing.

**Notes:**

prv-proton:
Hey team! Please [add your planning poker estimate](https://app.zenhub.com/workspaces/Low-Carbon-Fuel-Standard-644c2db3be098a001977ad0b/issues/bcgov/lcfs/160?planning-poker) with Zenhub @AlexZorkin @hamed-valiollahi @jig-patel @kevin-hashimoto @prv-proton

justin-lepitzki:
@Kailee-Douglas  @airinggov can we get a wireframe for this one?

airinggov:
isn't that this one?
https://preview.uxpin.com/59145260c7d8cc48969ac79401cc03d84843c3d9#/pages/164769377

justin-lepitzki:
Yes! Sorry I should have looked through the wireframes before asking for one. Thanks Al!

justin-lepitzki:
3 points left

Grulin:
- Four columns: Organization name, Balance, In reserve, Registered. Formatting does not cover the entire box. user can expand the columns to fit the table, but it does not stay this way.
- Default sort is alphabetical, but sorting "status" function and download buttons scramble the numbers in "compliance units' and 'in reserve' columns every time you use them. 
[...]

PR #364, opened 2024-02-05
Feat: org transfer flag 222

kevin-hashimoto:
- feat: added operating_name to org model
- feat: added operating_name value to seeders
- feat: updated update org service/repo
- refactor: renamed `AddOrganization` to `AddEditOrg`
- feat: added component flag `mode` to `AddEditOrg` so it can use the right endpoints depending on adding or updating org

notes:
- `AddEditOrg` should probably be cleaned up down the road, it is getting quite big and there are some pieces that can be simplified
  - mutations can probably be placed in a custom hook
  - extract default form values into a constant 
  - Address and Attorney address sections can probably be extracted into its own component
    - use form context in this case
  - Use the callback version of watch to perform side effects in a useeffect callback so we are not needlessly watching multiple values when we dont need to. this will be the 'syncing logic'.
    - watch is a subscription in this case, make sure to unsubscribe.
  - rename registered fields to match payload schema structure
    - can use dot notation for nested objects ( address, attorney address )


frontend/src/views/Organizations/AddEditOrg/AddEditOrg.jsx
AlexZorkin, line 1: What are your thoughts on just checking whether orgID exists to determine "mode" rather than explicitly settings it in the route?

AlexZorkin, line 1: We definitely need to break this component up. Please write up a few ways we could separate out its parts. I'll use this in a followup ticket.

frontend/src/views/Transactions/Transactions.jsx
```diff
@@ -58,15 +62,19 @@ export const Transactions = () => {
...
-        <BCButton
-          variant="contained"
-          size="small"
-          color="primary"
-          startIcon={<FontAwesomeIcon icon={faCirclePlus} size="2x" />}
-          onClick={() => navigate(ROUTES.TRANSFERS_ADD)}
-        >
-          <BCTypography variant="subtitle2">New transaction</BCTypography>
-        </BCButton>
+        {currentUserOrgData?.org_status.status === 'Registered' && (
```
AlexZorkin, line 65: Let's add 'Registered' as a constant for the app

AlexZorkin (changes requested):

frontend/src/views/Organizations/AddEditOrg/AddEditOrg.jsx
kevin-hashimoto: I was thinking of that originally. that way we can combine create and update db functions too ( it would essentially become an upsert function ). One thing I like about having a flag is if there are any client side differences from updating vs create views ( so far there isnt, may be a future consideration? ).

Ill change it to check on the backend and adjust the update/create functions to become a upsert function. #160 also touches the create function so It will be fixed with this update as well.

kevin-hashimoto: Instead of mode, we can check if the path param `orgID` is present to determine to create or update.
With this we can still separate the http methods POST for creating and PUT for updating _with_ path params

kevin-hashimoto: added to original comment

AlexZorkin (approved):

Issue #304, opened 2024-01-12
LCFS - Develop Pytest Tests for User Management Endpoints in FastAPI Backend


AlexZorkin:
**Describe the task:**
Create pytest tests for the user management endpoints in the FastAPI backend. This includes writing tests for each endpoint (GET, POST, PUT, DELETE) to verify their functionality, including successful operations, error handling, and edge case scenarios.

**Purpose:**
The goal is to ensure the robustness, reliability, and correctness of the user management endpoints. By implementing comprehensive tests, we aim to detect bugs early, uphold code quality, and guarantee that the API behaves as expected under various scenarios.

**Acceptance Criteria:**
- [ ] Write a test to verify the successful retrieval of the current user's information.
- [ ] Write a test for fetching a user by ID, including handling a non-existent user (expecting a 404 response).
- [ ] Write tests for the list users endpoint, including checking the response structure and verifying sort, filter, and pagination functionalities.
- [ ] Write tests for creating a new user, covering both successful creation and failure due to invalid data.
- [ ] Write tests for updating a user, including successful update and failure scenarios (e.g., updating a non-existent user).
- [ ] Write tests for the delete user endpoint, ensuring correct functionality and handling of conditions (like deleting a user who has never logged in).
- [ ] Write tests for fetching a user's roles and user history, including cases where no roles or history entries are found.
- [ ] Ensure all tests validate response status codes, returned data, and error messages where appropriate.

**Additional context:**
- Familiarize with the user management endpoint logic to understand necessary mock setups and dependencies.
- Follow the async conventions used in the existing test suite for consistency.
- Collaborate with the development team for any specific scenarios or edge cases that need to be tested.

```python
import pytest
from fastapi import FastAPI
from httpx import AsyncClient
from starlette import status

@pytest.mark.anyio
async def test_get_current_user_success(client: AsyncClient, fastapi_app: FastAPI) -> None:
    # TODO: Implement test for successful retrieval of current user
    pass

@pytest.mark.anyio
async def test_get_user_by_id_success(client: AsyncClient, fastapi_app: FastAPI) -> None:
    # TODO: Implement test for successful retrieval of a user by ID
    pass

@pytest.mark.anyio
async def test_get_user_by_id_not_found(client: AsyncClient, fastapi_app: FastAPI) -> None:
    # TODO: Implement test for handling non-existent user ID
    pass

@pytest.mark.anyio
[...]

Issue #187, opened 2023-11-09
LCFS - Download User Information


justin-lepitzki:
**Description:**
- Enable the ability to download BCeID user information from the Organizations page into a spreadsheet
- IDIR users only 
- User information to download includes: Last name, First name, Email, BCeID User ID, Title, Phone, Mobile, Status, Role(s), Organization name
- File type = .xls
- Workbook name format: BC-LCFS-bceid-YYYY-MM-DD.xls, where YYYY-MM-DD is the day the file was downloaded
- Worksheet name: BCeID Users
- Spreadsheet columns should be formatted to give adequate spacing for the information displayed
- Add excel icon within the button (see screenshot below); note that this is not shown in the wireframe below but we would like to display the excel icon. 
- The button label switches to "Downloading..." after the user clicks on the button and the sheet is being downloaded (loading). The button then switches back to the regular label after it has downloaded. 

**Note:** This feature is near identical to the one that exists in TFRS

**Wireframe:** 
- UXPin: https://preview.uxpin.com/59145260c7d8cc48969ac79401cc03d84843c3d9#/pages/164769377/simulate/no-panels
- Screenshot of spreadsheet: [Private Zenhub Image](https://api.zenhub.com/attachedFiles/eyJfcmFpbHMiOnsibWVzc2FnZSI6IkJBaHBBOC9BQVE9PSIsImV4cCI6bnVsbCwicHVyIjoiYmxvYl9pZCJ9fQ==--0578d2d48ca3cabc08bcd7dde94c5963ca1a5678/Screenshot%202023-11-09%20bceid%20users%20download%20as%20xls.png)

- Excel icon within the download button (from TFRS): [Private Zenhub Image](https://api.zenhub.com/attachedFiles/eyJfcmFpbHMiOnsibWVzc2FnZSI6IkJBaHBBN3JBQVE9PSIsImV4cCI6bnVsbCwicHVyIjoiYmxvYl9pZCJ9fQ==--b3262732bd9c2692d8107a83b2c11904867287ff/Screenshot%202023-11-09%20Download%20buttons%20in%20TFRS.png)
[...]

Issue #175, opened 2023-11-03
LCFS - Create new IDIR users 


justin-lepitzki:
**Description:**

Create new IDIR users:

- New user button on the Users sub-tab within the Administration page (see dependencies below for the cards to create these pages)
- Button only visible to IDIR users with the Administrator role
- Clicking on the button takes the user to the 'New User' page (see wireframe)
- Certain fields are required while other fields are optional; see wireframe for which fields are required vs. optional
- Telephone number is free form text to allow for users to input extensions
- User can be assigned role(s), they are stackable
- Clicking on the Save button adds the user to the database and they are then visible in the associated User react table in the UI; it also takes them back to the Users sub-page
- Clicking on the Back button takes the user back to the Users sub-page; no records are saved in the database
- Implement breadcrumb navigation on this page

Note: The status radio buttons and access control for logging in (active / inactive) will be handled in a different card

**Wireframe:**
- New user button: https://github-production-user-asset-6210df.s3.amazonaws.com/132306457/280809091-4b14936b-a85d-4d43-a7f1-52b583facedf.png
- New user page: https://preview.uxpin.com/59145260c7d8cc48969ac79401cc03d84843c3d9#/pages/164316070/simulate/no-panels

**Purpose and benefit to user:**
IDIR users can add new users to government to help complete required tasks

**Acceptance Criteria:**

- [x] Given I am an IDIR user with an Administrator role, when I navigate to the Users sub-tab within the Administration page, then I should see a 'New User' button.
- [x] Given I click on the 'New User' button, when I am taken to the 'New User' page, then I should be able to fill out required and optional fields as indicated on the wireframe.
- [ ] Given I have entered all the necessary information on the 'New User' page, when I click the 'Save' button, then the new user should be added to the database, shown in the User react table, and I should be redirected back to the Users sub-page.
- [ ] Given I am on the 'New User' page, when I click the 'Back' button, then I should return to the Users sub-page without any new records being saved in the database.
- [ ] Given I am on the 'New User' page, when I interact with the page, then the breadcrumb navigation should reflect the path taken to arrive at this page.

**Development Checklist:**
- [ ] Create a New user button on the User sub-tab in Administration page.
- [ ] Button should only be visible to Idir use with admin role.
- [ ] Certain fields are option and some are required. Please check wireframe . 
- [ ]  User would have role and can be assigned. Role is not part of this ticket but this ticket is depended on role creation ticket.
[...]

Issue #174, opened 2023-11-03
LCFS - Create new BCeID users associated with their organization 


justin-lepitzki:
**Description:**

BCeID & IDIR users can create new BCeID users associated with an (their) organization

- New user button on the Organization Details page
- New user button only visible to BCeID users with the Managing Users role and IDIR users with the Administrator role
- Clicking on the button takes the user to the 'New User' page (see wireframe)
- Certain fields are required while other fields are optional; see wireframe for which fields are required vs. optional
- Telephone number is free form text to allow for users to input extensions
- User can be assigned role(s), they are stackable
- Clicking on the Save button adds the user to the database and they are then visible in the associated User react table in the UI; it also takes the user back to the Organization Details page
- Clicking on the Back button takes the user back to the Organization Details page; no records are saved in the database
- Implement breadcrumb navigation on this page

Note: The status radio buttons and access control for logging in (active / inactive) will be handled in a different card

**Wireframe:**
- New user button: https://preview.uxpin.com/59145260c7d8cc48969ac79401cc03d84843c3d9#/pages/164774350/simulate/no-panels
- New user page: https://preview.uxpin.com/59145260c7d8cc48969ac79401cc03d84843c3d9#/pages/164314016/simulate/no-panels

**Purpose and benefit to user:**
Ability to create new users

**Acceptance Criteria:**

- [x] Given I am a BCeID user, When I click on the Organization tab from the navigation bar I should be able to see the button to create new user for the organization.
- [x] Given I am a BCeID user, When I click on the 'New User' button I should be taken to the user form page where I can add the details of the user.

**Development Checklist:**
- [ ] Follow the same steps and design that were used to create the user form for IDIR user page.

**Notes:**

prv-proton:
Hey team! Please [add your planning poker estimate](https://app.zenhub.com/workspaces/Low-Carbon-Fuel-Standard-644c2db3be098a001977ad0b/issues/bcgov/lcfs/174?planning-poker) with Zenhub @AlexZorkin @jig-patel @kevin-hashimoto @prv-proton

Grulin:
- "Last Name" validation is persisting.
- Unable to add 'New User' to organization account because clicking on the 'Save' button adds the user to the database and they are then visible in the associated User react table in the UI; it also takes them back to the Users sub-page
- Certain fields are required while other fields are optional; see wireframe for which fields are required vs. optional
- Implement breadcrumb navigation on this page



[...]

PR #375, opened 2024-02-09
fix: user management fixes and test cases.

prv-proton:
- #304 
- #175 
- #187 
- ag-grid ability to store filter and sort state in browser cache.
- ag-grid styling fix for User and Organization table.
- backend user modules re-structuring to new format
- redis caching for retrievals, and flushing them in case of updates, adds or deletes.


prv-proton:
- Mutation and edit functionality will be added as part of #174

PR #405, opened 2024-02-20
Fix: dev log out error 

kevin-hashimoto:
- fix: issue with logging out on dev
- fix: config values not being available to use


AlexZorkin (approved):
We just need to confirm with Kuan that the mounted config file is added to the dev environment.

Issue #409, opened 2024-02-21
LCFS - Implement Transactions List View with Real Data


AlexZorkin:
**Description:** This card involves the completion of the Transactions List View functionality in our application, ensuring that it is integrated with the backend to load and display real transaction records. The view should dynamically display transactions of various types (e.g., transfers, administrative adjustment, initiative agreements) and route the user to the appropriate detailed view when a transaction is clicked.

**Purpose and benefit to user:** This feature allows users to quickly view and access their transaction history within our application, providing a comprehensive overview of their compliance unit history. By supporting different transaction types and enabling easy navigation to detailed views, we enhance user experience and utility, making compliance management more intuitive and efficient.

**Acceptance Criteria:**

- [x] Given I am a user, When I navigate to the Transactions List View, then I see my real transaction records loaded.
- [x] Given I am a user, When I click on a transaction of type 'transfer', then I am routed to the transfer view with that transaction's details loaded.

**Development Checklist:**
- [ ] Finish backend repository methods to retrieve all transactions and format data including types.
- [ ] Implement frontend logic to display transactions and determine their types.
- [ ] Ensure the frontend correctly routes to the detailed view for a transaction when clicked.
- [ ] Test integration between frontend and backend to ensure accurate data display and routing.
- [ ] Implement lazy loading or pagination if the transaction list is very long to improve performance.

**Notes:**
- Ensure that the backend accurately categorizes each transaction type to facilitate correct frontend routing.
- Include error handling for cases where transaction data cannot be loaded.

Grulin:
This card #409 functions great, except that for the acceptance criteria 'when I navigate to the Transactions List View, then I see my real transaction records loaded' I can see all transactions by all organizations. 

On the BCeID side, view of transactions should be limited only to those transactions involving the organization viewing this transactions table. 

Please see screenshot for example of transactions visible to LCFS Org 1. As a user of Org 1, I am also able to click on these transactions and view them:


![Image](https://github.com/bcgov/lcfs/assets/156380046/3167bf2f-086c-4575-ba02-ceef1146e695)

Issue #228, opened 2023-11-24
LCFS - Delete a draft transfer


justin-lepitzki:
**Description:**
The ability to delete a draft transfer:

- When viewing a draft transfer
- There is a Delete Draft button
- Clicking on the button brings up a confirmation modal window
- Clicking on the Cancel button within the modal window takes the user back to the draft transfer
- Clicking on the Delete Draft button deletes the draft (i.e., hides it in the UI) and takes the user back to the Transactions page; the transfer is no longer visible in the Transactions table
- A success alert should appear (alert from Material UI: https://mui.com/material-ui/react-alert/
- Users must have the Transfers role to delete draft transfers
- Once draft is deleted it should change it status in delete in DB and shouldn't display on UI.

**Wireframe:** 
https://preview.uxpin.com/59145260c7d8cc48969ac79401cc03d84843c3d9#/pages/164927056/simulate/no-panels

**Screenshot of confirmation modal window:** [Image](https://github.com/bcgov/lcfs/assets/131904954/efd1edc5-3de9-4625-ae77-be4cda388601)

**Purpose and benefit to user:**
Users can delete draft transfers if they no longer wish to execute the transfer

**Acceptance Criteria:**

- [ ] Given I am a <role>, When I am <doing something>, then <this happens>
- [ ] Given I am a <role>, When I am <doing something>, then <this happens>

**Development Checklist:**
- [x] A Create delete draft button when on edit transfer page
- [x] B Make sure user who has transfer role are able to delete the transfer
- [x] C Make sure to change the status to delete in DB and not completely delete the entry in db. On UI, we should display the transfer whose status is not deleted.

**Notes:**

prv-proton:
Hey team! Please [add your planning poker estimate](https://app.zenhub.com/workspaces/Low-Carbon-Fuel-Standard-644c2db3be098a001977ad0b/issues/bcgov/lcfs/228?planning-poker) with Zenhub @jig-patel @prv-proton

PR #420, opened 2024-02-24
Feat: delete draft transfer #228

kevin-hashimoto:
#228 

- refactor: overhaul of the `AddEditTransfer` component and its child components
- feat: confirmation modal
- feat: button cluster configuration and display
- feat: new endpoints

TODO:
- ~~Transaction page grid display ( #409 related? )~~
  - ~~Create the endpoint for paginated transactions~~
  - ~~Alter the grid component on the Transactions page to display the paginated transaction data~~
- ~~Transaction page alert display~~
  - ~~detect alerts when a user lands on the Transactions page.~~
  
  
  Notes:
- The new patterns in the `AddEditTransfer` component should set us up to easily integrate the rest of the transfers features. This should be merged in before starting other components of the transfers.
- Different states of the child component UIs should be determined by checking the status of the transfer. `transferData?.transfer_status.status`

Usage:
- `buttonClusterConfig` determines the display and functionality of the buttons that should be displayed at the bottom of the transfer pages.
  - each key of the config object represents the state that the transfer is in.
  - check `buttonConfigs.js` to see examples of how the ui configuration is set for the buttons
  - attach a `handler` to the button that calls a function to do something
    - You'll have access to the form data in the handler. 
    - For buttons that open a modal, set the modal data with `setModalData`. It takes `onConfirm`, `buttonText`, `text` properties.
  - You can set the button as `disabled` as well
- There are various mutation functions:
  - `createDraft`: creates a new draft transfer.
  - `updateDraft`: updates an already existing draft transfer but does not update its status.
  - `updateTransfer`: updates an already existing transfer, updates status and comments if there are any
- Depending on what happens after each step, the `updateTransfer` may have to be slightly modify to accommodate.
- Within each of the child form components ( eg. `TransferDetails`, `AgreementDate` etc. ), we can determine Transfer status with  the `useTransfer` hook, then modify ui based on that ( changing input field to normal text etc ).


backend/lcfs/web/api/transfer/services.py
```diff
@@ -106,4 +112,20 @@ async def update_transfer(self, transfer_data: TransferUpdate) -> TransferSchema
...
\ No newline at end of file
+        return TransferSchema.model_validate(updated_transfer)
+
+    @service_handler
+    async def update_transfer(self, transfer_id: int, transfer_data: TransferUpdate) -> TransferSchema:
+        transfer = await self.repo.get_transfer_by_id(transfer_id)
+        if not transfer:
+            raise DataNotFoundException(
+                f"Transfer with ID {transfer_id} not found")
+        transfer.transfer_status_id = transfer_data.transfer_status_id
```
AlexZorkin, line 123: this value has recently been updated to current_status_id so we'll need to swap that out here. Reason being is the TransferHistory table will keep the change history for each transfer, ie when their status changes. Please add the creation of a new TransferHistory record here with the new status, and set current_status on the Transfer record.

backend/lcfs/web/api/transfer/services.py
AlexZorkin, line 1: Please make sure the backend unit tests are working with these changes, and add any additional coverage needed for the update_transfer method.

backend/lcfs/web/api/transfer/views.py
AlexZorkin, line 1: add/update the backend unit tests for update_transfer_draft

AlexZorkin (changes requested):
Looks great. You'll have to merge release into this branch first because there are a few breaking changes that need to be fixed around transfer_status getting renamed to current_status. Mentioned in comments.

kevin-hashimoto:
@AlexZorkin fixes addressed

Issue #251, opened 2023-12-06
LCFS - Organization can rescind a transfer


justin-lepitzki:
**Description:**

Develop the ability for organizations to rescind a transfer.

Note card #220 is likely needs to be completed prior to this card. 

- Rescind button 
- Organization 1 visibility in the following statuses: Sent, Submitted, Recommended
- Organization 2 visibility in the following statuses: Submitted, Recommended
- Must have the Transfer or Signing Authority roles, otherwise button is greyed out
- Clicking on the button brings up a confirmation modal window
- Clicking the cancel button in the modal window brings the user back to the view transfer screen
- Clicking on the Rescind button in the modal window (1) brings up a success message/alert/modal, (2) The user is returned to the transaction page, (3) The transaction row for the transfer is highlighted, and (4) the transfer status is changed to Rescinded.
- A transaction history record is saved to the database and visible in the UI when viewing the transfer: Rescinded on [Month DD, YYYY] by [User] of [Organization]. This record is based on the user with the Transfer and/or Signing Authority role that clicked the Rescind button. 
- Success alert from Material UI: https://mui.com/material-ui/react-alert/
- The transfer breadcrumb within the view page is updated with the Rescind state; this icon is placed in the breadcrumb at the point in which the transfer was rescinded (i.e., replaces the icon for the current status of the transfer at the time it was rescinded; see wireframe)
- With a Rescinded status, the transfer workflow has ended. Users cannot take further action on this transfer. 

**Data Validation & restrictions:**
Being able to rescind a transfer needs to have the following restrictions: 
- Requires the Transfer or Signing Authority role to click on the Rescind Transfer button, otherwise greyed out 

**Validation Messaging:**
May be dependent on front-end vs. back-end validation:

- Does not have the Transfer or Signing Authority role (Rescind Transfer button): "The Transfer or Signing Authority roles are required to rescind a transfer."

**Wireframe:** 
- Rescind button (organization 1): https://preview.uxpin.com/59145260c7d8cc48969ac79401cc03d84843c3d9#/pages/165000525/simulate/sitemap
- Rescinded transfer (view page): https://preview.uxpin.com/59145260c7d8cc48969ac79401cc03d84843c3d9#/pages/165000529/simulate/sitemap
- Confirmation alert: [Image](https://github.com/bcgov/lcfs/assets/131904954/fa9e5967-d6fc-4e63-a179-26d7dfc45aa1)

**Purpose and benefit to user:**
[...]

PR #444, opened 2024-03-04
feat: rescind transfer 251

kevin-hashimoto:
#251

- feat: rescind transfer
- refactor: view transfer page to use button cluster configuration
- feat: new useHasRoles hook to check if current user has necessary roles
- refactor: move updateTransfer mutation to a custom hook


frontend/src/views/Transfers/ViewTransfer/ViewTransfer.jsx
```diff
@@ -1,46 +1,48 @@
...
 
 export const ViewTransfer = () => {
+  const [modalData, setModalData] = useState(null)
   const { t } = useTranslation(['common', 'transfer'])
   const iconSizeStyle = {
     fontSize: (theme) => `${theme.spacing(12)} !important`
   }
   const navigate = useNavigate()
   const { transferId } = useParams()
+  const hasRoles = useHasRoles([roles.transfers, roles.signing_authority])
```
prv-proton, line 44: wouldn't this restrict us from using any other roles? Say for certain content to display we want admin roles and for others, supplier roles. But using this would make us to declare multiple hasRoles variables like hasAdminRole, hasTransferRole?

kevin-hashimoto: yea thats true, if we are creating components or functions that need to cater to multiple users with different roles this wont work.

We would need to specify roles within the component/function with conditionals to account for different users.

Any ideas on how we might be able to improve this?

kevin-hashimoto: actually, the `useCurrentUser` hook has a `hasRoles` function that can replace this. that way we can use that function where ever we need

AlexZorkin (approved):
The buttonClusterConfig is a nice solution to our many status dependant buttons. I wonder if we can extend it further to add user role specific conditional logic. Something to think about for the next sprint.

PR #462, opened 2024-03-06
fix: styling & navigation fixes

prv-proton:
- To wrap the roles in ag-grid, to adjust the row height.
- Add new organization user navigation issues for IDIR users.
- Links to enable to open in new tab.
- missing AddPlainComment component.


AlexZorkin (approved):
Nice work on the row opening in new tabs.

Issue #316, opened 2024-01-17
LCFS - Director can return transfer to analyst


justin-lepitzki:
**Description:**
Develop the ability for the Director to return the transfer to the analyst (i.e., status change): 

- 'Return to Analyst' button visible to Government users with the Director role when the transfer has a status of 'Recommended'
- 'Return to Analyst' button in the bottom right hand corner of the page
- Clicking on the 'Return to Analyst' button brings up a confirmation modal window
- Clicking the cancel button in the modal window brings the user back to the view transfer screen
- Clicking on the 'Return to Analyst' button in the modal window (1) brings up a success message/alert/modal, (2) The user is returned to the transaction page, (3) The transaction row for the transfer is highlighted, and (4) the transfer status is changed back to Submitted.
- No new transaction history record is created. 
- Success alert from Material UI: https://mui.com/material-ui/react-alert/
- The transfer breadcrumb within the view page is updated back to the Submitted state 

**Confirmation Model Messaging:**
Similar to when the Director refuses a transfer, we want to provide a warning message to the Director when they click on the Return to Analyst button. See attached screenshot for formatting and UI styling (e.g., yellow warning background, warning icon)
- You are strongly encouraged to provide an explanation to the analyst as to why the transfer is being returned.
- Are you sure you want to return the transfer to the Analyst? This will change the status of the transfer back to Submitted.

**Wireframe:**

- Director view of recommended transfer: https://preview.uxpin.com/59145260c7d8cc48969ac79401cc03d84843c3d9#/pages/165001091/simulate/no-panels
- Warning message for Director within the confirmation modal window: 
![Image](https://github.com/bcgov/lcfs/assets/131904954/afb7e605-9019-4ebb-8a8c-2e7cac0d4ed3)

**Purpose and benefit to user:**
Ability for the director to return the transfer to the analyst to revise their recommendation and/or provide additional information to the director support the decision to record or refuse the transfer. 

**Acceptance Criteria:**

- [ ] Given I am an IDIR user, When the status of the transfer is 'Recommended', the 'Return to analyst' button should show
- [ ]  When the 'Return to analyst' button is clicked, a modal should appear
- [ ] The modal should display a confirmation of if the transfer should 'return to analyst'
- [ ] Once confirmed, the transfer status should return to 'submitted' and the breadcrumb should highlight the 'submitted' step

**Development Checklist:**
- [ ] No new history records should be logged
- [ ] Only the status should change. 'Recommended' to 'Submitted'
[...]

PR #469, opened 2024-03-07
feat: director can return transfer to analyst

kevin-hashimoto:
#316 

- refactor: consolidated button styling config
- feat: Director role IDIR users can return transfer to analyst

### DIFF
--- a/frontend/src/views/Transfers/ViewTransfer/ViewTransfer.jsx
+++ b/frontend/src/views/Transfers/ViewTransfer/ViewTransfer.jsx
@@ -262,7 +262,8 @@ export const ViewTransfer = () => {
             title: t('confirmation'),
             content: t('transfer:returnConfirmText'),
             warningText: t('transfer:returnWarningText')
-          })
+          }),
+        disabled: !hasRoles(roles.director)
       },
       {
         ...containedButton(t('transfer:recordTransferBtn')),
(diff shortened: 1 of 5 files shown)


AlexZorkin (commented):
Need to fix a few conflicts, but otherwise looks good.

frontend/src/views/Transfers/ViewTransfer/ViewTransfer.jsx
```diff
@@ -224,7 +236,31 @@ export const ViewTransfer = () => {
...
-        ...saveDraftButton(t('saveBtn')),
+        ...redOutlinedButton(t('transfer:refuseTransferBtn'))
+      },
+      {
+        ...outlinedButton(t('transfer:returnToAnalystBtn')),
+        handler: () =>
+          setModalData({
+            primaryButtonAction: () =>
+              updateTransfer({
+                newStatus: 4,
```
AlexZorkin, line 247: Can you think of a way so we can use status names rather than ids here?

AlexZorkin (approved):
Looks good, just one comment on a possible improvement. We can move the result to a new ticket if its more than a 1.

AlexZorkin:
Just a bump to review the conflicts and comment.

frontend/src/views/Transfers/ViewTransfer/ViewTransfer.jsx
kevin-hashimoto: would the goal be to make it more readable or to pass the actual string value?

few options come to mind:

1.
```
export const statuses = {
  ...
  sent: {
    id: 4,
    label: 'Sent'
  }
  ...
}
```
and then pass in `statuses.sent.id`.

2. create a util function that maps a string value to a status id
3. using the actual string value of a status and handling the mapping on the backend

AlexZorkin: I think we should use the string value, and then map that to the Enum value on the backend. Updating a transfer on the backend has the same issue, checking against the ids without knowing their value. I will create a new ticket for this refactor. Please just fix the existing conflicts and we'll close this pr.

PR #472, opened 2024-03-09
Feat: Audit Field Updates - 457

AlexZorkin:
- added base repository to repos for automatic audit field updates
- fixed transaction seeder to work with materialized view
- updated base models and dependancies


backend/lcfs/web/api/organizations/repo.py
```diff
@@ -20,7 +21,7 @@
 logger = getLogger("organizations_repo")
 
 
-class OrganizationsRepository:
+class OrganizationsRepository(BaseRepository):
     def __init__(self, db: AsyncSession = Depends(get_async_db_session)):
         self.db = db
```
prv-proton, line 26: super().__init__(db)

backend/lcfs/web/api/repo.py
```diff
@@ -0,0 +1,23 @@
+from sqlalchemy.ext.asyncio import AsyncSession
+from lcfs.db.base import Auditable
+
+class BaseRepository:
+    def __init__(self, db: AsyncSession):
+        self.db = db
+
+    async def commit_to_db(self):
+        user = self.db.info.get('user')
+        username = user['keycloak_username'] if user and 'keycloak_username' in user else 'no_user'
```
prv-proton, line 10: since this is not a dictionary object it gave me error.
so changed to below code in my local.
`username = user.keycloak_username if user and user.keycloak_username else 'no_user'`

but still faced below error on my local while creating a new user. It might possible due to the @transactional decorator used in users/services.py. Please have a look at it. Thanks.

`[ERROR] [2024-03-12 16:43:27,927]  lcfs.web.api.user.repo.wrapper - Repo error in create_user (file: lcfs/backend/lcfs/web/api/user/repo.py) - Can't operate on closed transaction inside context manager.  Please complete the context manager before emitting further commands.
[ERROR] [2024-03-12 16:43:27,928]  lcfs.web.api.user.views.wrapper - 
INFO:     127.0.0.1:39888 - "POST /api/users HTTP/1.1" 500 Internal Server Error`

prv-proton (changes requested):

backend/lcfs/web/api/repo.py
AlexZorkin: Thank you for catching this. I fixed the test auth user to be a UserProfile object which is where the discrepancy was. I also fixed the session closing early on user create.

Issue #256, opened 2023-12-08
LCFS - Analyst transfer recommendation to the director


justin-lepitzki:
**Description:**
Develop the analyst recommendation section on the view transfer page and the ability for the analyst to make a recommendation on a transfer. 

- 'Analyst Recommendation to the Director' heading
- 'Recommend Record' and 'Recommend Refuse' radio buttons
- 'Recommend' button in the bottom right hand corner of the page
- Clicking on the 'Recommend' button brings up a confirmation modal window
- Clicking the cancel button in the modal window brings the user back to the view transfer screen
- Clicking on the 'Recommend' button in the modal window (1) brings up a success message/alert/modal, (2) The user is returned to the transaction page, (3) The transaction row for the transfer is highlighted, and (4) the transfer status is changed to Recommended.
- A transaction history record is saved to the database and visible in the UI when viewing the transfer: Recommended on [Month DD, YYYY] by [User] of [Organization]. This record is based on the IDIR user with the Government Analyst role that clicked the Recommend button. 
- This recommendation transaction history record is only visible to IDIR users
- The comment free text field (government comments to organizations) is saved to the database, if there is a comment added (see card #261 for development of this feature).
- The two radio buttons are no longer selectable; they are greyed out, with the single selection made
- Success alert from Material UI: https://mui.com/material-ui/react-alert/
- The transfer breadcrumb within the view page is updated to the Recommended state for IDIR users only; the transfer remains in the Submitted state for BCeID users

**Data Validation & restrictions:**
- Only IDIR users with the Government Analyst role can make recommendations (select the radio buttons, click on the Recommend button)
- Only one radio button can be selected (either Recommend Record or Recommend Refuse, but not both)
- One of the radio buttons must be selected before the analyst can click on the Recommend button (greyed out until a recommendation is made)

**Validation Messaging:**
May be dependent on front-end vs. back-end validation:
- Does not have the Government Analyst role (Recommend button): "The Government Analyst role is required to make a recommendation to the director."
- Has not selected one of the radio buttons (Recommend button): "A recommendation is required before proceeding."

**Wireframe:** https://preview.uxpin.com/59145260c7d8cc48969ac79401cc03d84843c3d9#/pages/165001075/simulate/no-panels

**Purpose and benefit to user:**
Provides the necessary functionality for government users to make recommendations to the director 

**Acceptance Criteria:**

[...]

PR #475, opened 2024-03-12
Feat: Analyst recommend to director 256

kevin-hashimoto:
#256 

- feat:  recommendation component that houses the recommend options for an analyst
- feat:  transfer_recommendation_status table for the analyst recommendation status
- feat: seeder for transfer recommendation status
- feat: button cluster for the submitted status transfer


AlexZorkin (commented):
What's the benefit of user the transfer_recommendation_status over just setting the transfer status to either Recorded or Refused?

kevin-hashimoto:
> What's the benefit of user the transfer_recommendation_status over just setting the transfer status to either Recorded or Refused?

My thinking was that since the recommendation of 'refuse' or 'record' was not really a transfer status, but more of a sub status of 'recommended' I opted to keep that separate. It does however create a bit of overhead managing another table but I think that overhead is quite small. Thinking of it now, we could just create another column that takes a recommendation enum ( 'refuse', 'record' ). thoughts?

AlexZorkin:
I like your idea of creating a new enum column on Transfer that holds the recommendation value. Eliminates the need for a join every time we call Transfer.

AlexZorkin (changes requested):
I like your idea of creating a new enum column on Transfer that holds the recommendation value. Eliminates the need for a join every time we call Transfer. Please go ahead with that change and fix any conflicts. Thanks :)

AlexZorkin (approved):
This is great. Thank you for changing up the recommendation.

PR #484, opened 2024-03-15
Fix: Organization User create bug fix

prv-proton:
Fix for minor bug found while trying to create a user via BCeID user.


AlexZorkin (approved):
I missed this one, thank you for fixing it. We should add a test for this.

PR #496, opened 2024-03-18
Fix: Updated Org Transactions

AlexZorkin:
- fixed backend unit tests for new user config
- updated query logic for org transactions to not return other org txns


prv-proton (approved):
Looks good.

PR #512, opened 2024-03-22
Feat: Transfer History List - 478

AlexZorkin:
- migration adding user_profile to transfer_history table
- hooked up real data for transfer history list
- created a separate transfer history component
- added dynamic labelling based on transfer status for history list

### DIFF
--- a/backend/lcfs/web/api/transfer/repo.py
+++ b/backend/lcfs/web/api/transfer/repo.py
@@ -15,6 +15,7 @@ from lcfs.db.models.Transfer import Transfer
 from lcfs.db.models.TransferStatus import TransferStatus, TransferStatusEnum
 from lcfs.db.models.TransferCategory import TransferCategory
 from lcfs.db.models.TransferHistory import TransferHistory
+from lcfs.db.models.UserProfile import UserProfile
 
 logger = getLogger("transfer_repo")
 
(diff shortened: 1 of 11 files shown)


backend/lcfs/web/api/transfer/repo.py
```diff
@@ -143,7 +147,8 @@ async def add_transfer_history(self, transfer_id: int, transfer_status_id: int)
         """
         new_history_record = TransferHistory(
             transfer_id=transfer_id,
-            transfer_status_id=transfer_status_id
+            transfer_status_id=transfer_status_id,
+            user_profile=self.request.user
```
kevin-hashimoto, line 151: seems to be an error when saving a record.
`'TransferRepository' object has no attribute 'request'`

probably just adding the `request` object to the init up top

backend/lcfs/web/api/transfer/repo.py
kevin-hashimoto, line 1: `update_transfer` and `create_transfer` will need `transfer_history` in the `self.sb.refresh` list or else a `MissingGreenlet` error occurs and throws an error on the front end. Tranfer will save in the background regardless though.

kevin-hashimoto (changes requested):

backend/lcfs/web/api/transfer/repo.py
AlexZorkin: I updated the service to pass down the user_profile_id. We probably should avoid having the request injected into the repo layer, seems a bit anti-repository pattern.

backend/lcfs/web/api/transfer/repo.py
AlexZorkin: Thanks for this catch. I have updated the refresh lists.

kevin-hashimoto (approved):
LGTM!

PR #524, opened 2024-04-04
Feat: Transfer Balance Validation

AlexZorkin:
- Added backend validation on transfer create and update to ensure Suppliers can't create transfers with a quantity greater than their compliance units balance
- Added a ref to BCAlert so it can render multiple times on click
- Fixed render bug on org user display
- Conditional rendering of supplier balance added to header component
- Formatting in supplier balance fixed

### DIFF
--- a/backend/lcfs/web/api/transfer/validation.py
+++ b/backend/lcfs/web/api/transfer/validation.py
@@ -18,7 +18,7 @@ class TransferValidation:
         self.org_repo = org_repo
         self.request = request
 
-    async def update_transfer(self, request: Request, transfer_create: TransferCreateSchema):
+    async def governent_update_transfer(self, request: Request, transfer_create: TransferCreateSchema):
         # Ensure only the valid statuses are passed.
         if transfer_create.current_status not in LCFS_Constants.GOV_TRANSFER_STATUSES:
             raise HTTPException(
--- a/backend/lcfs/web/api/transfer/validation.py
+++ b/backend/lcfs/web/api/transfer/validation.py
@@ -18,7 +18,7 @@ class TransferValidation:
         self.org_repo = org_repo
         self.request = request
 
-    async def governent_update_transfer(self, request: Request, transfer_create: TransferCreateSchema):
+    async def government_update_transfer(self, request: Request, transfer_create: TransferCreateSchema):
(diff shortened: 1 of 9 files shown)


frontend/src/views/Transfers/AddEditViewTransfer.jsx
```diff
@@ -210,14 +211,21 @@ export const AddEditViewTransfer = () => {
...
       }
+      alertRef.current.triggerAlert()
     },
     onError: (_error, _variables) => {
-      setAlertMessage(
-        transferId
-          ? t('transfer:actionMsgs.errorUpdateText')
-          : t('transfer:actionMsgs.errorCreateText')
-      )
+      const errorMsg = _error.response.data?.detail
```
prv-proton, line 217: would also need to clear the confirmation modal. As it still persisting when we click on "Sign and send" button.

backend/lcfs/web/api/transfer/validation.py
```diff
@@ -18,7 +18,7 @@ def __init__(
         self.org_repo = org_repo
         self.request = request
 
-    async def update_transfer(self, request: Request, transfer_create: TransferCreateSchema):
+    async def governent_update_transfer(self, request: Request, transfer_create: TransferCreateSchema):
```
prv-proton, line 21: just a typo on "government"

backend/lcfs/web/api/transfer/views.py
```diff
@@ -33,14 +33,14 @@ async def get_transfer(
 @router.put("/{transfer_id}", response_model=TransferSchema, status_code=status.HTTP_200_OK)
 @view_handler
 @roles_required("Government")
-async def update_transfer(
+async def governent_update_transfer(
```
prv-proton, line 36: same as above, typo.

prv-proton (changes requested):
All functionality looks good. Thank you.

frontend/src/views/Transfers/AddEditViewTransfer.jsx
AlexZorkin: I added the modal close, as well as a scroll to top so the user can see the error.

PR #537, opened 2024-04-08
feat: vitest unit testing base

kevin-hashimoto:
- feat: vitest coverage
- feat: vitest ui
- feat: react testing library setup
- feat: msw setup
- feat: multiple test files with examples that should cover the majority of our tests
- chore: package update


frontend/src/components/Logout.jsx
```diff
@@ -3,13 +3,13 @@ import BCButton from '@/components/BCButton'
 import { logout } from '@/utils/keycloak'
 import { useCurrentUser } from '@/hooks/useCurrentUser'
 
-const Logout = () => {
+export const Logout = () => {
```
AlexZorkin, line 6: Is removing the default just a stylistic choice?

frontend/src/tests/utils/wrapper.jsx
```diff
@@ -4,17 +4,21 @@ import { CssBaseline, ThemeProvider } from '@mui/material'
...
+    mutations: {
       retry: false
     }
-  },
-  logger: {
-    log: console.log,
-    warn: console.warn,
-    error: () => {}
   }
+  // logger: {
```
AlexZorkin, line 17: was this too loud? should we just remove it?

frontend/src/views/Organizations/AddEditOrg/AddEditOrg.test.jsx
AlexZorkin, line 1: are these all left here as reference?

frontend/src/views/Organizations/Organizations.test.jsx
AlexZorkin, line 1: same, left for reference?

AlexZorkin (changes requested):
It's nice to get back to baseline! We can leave the commented out tests for later reference. One thing that would be helpful is if you could create a shortlist of the components you think we should add tests for. I will create followup tickets to add coverage on our gaps.

frontend/src/components/Logout.jsx
kevin-hashimoto: moves it inline with our other components that follow explicit export/imports

frontend/src/tests/utils/wrapper.jsx
kevin-hashimoto: new react-query version deprecates custom loggers. I've removed the comment

kevin-hashimoto:
> It's nice to get back to baseline! We can leave the commented out tests for later reference. One thing that would be helpful is if you could create a shortlist of the components you think we should add tests for. I will create followup tickets to add coverage on our gaps.

@AlexZorkin added todo tests which should now show up in the test ui as 'skipped'

AlexZorkin (approved):
Lots todo haha. Thanks for getting us back to baseline.

PR #582, opened 2024-04-23
Feat: Refactor of Session System - 545

AlexZorkin:
- Removed BaseRepository class in favor of global session management
- Moved create and update user method to dependancies
- refactored services and repos using flush and refresh


kevin-hashimoto (approved):
LGTM! Im liking how its a bit more simpler now

PR #591, opened 2024-05-03
Feat: Add Transaction Filtering by Organization for IDIR Users - 543

### LINKED ISSUE #543: LCFS - 'All organization compliance units' feature for IDIR users on the transactions page
**Describe the task**

For IDIR users on the the Transactions page, they should have a display of "compliance unit balance" that reflects what a filter is set to (either as an all organizations default or a specific organization once the drop-down filter is chosen) and the ability to filter out transactions involving specific registered organizations (both sending or receiving compliance units). 

This includes a 2 part feature:

1) Develop a feature to allow IDIR users to see the balance of the the filtered choice (either all organizations or one chosen organization)
2) Develop a drop-down menu of all registered organizations for IDIR users to filter out transactions that include that specific organization. 

Note: Currently, BCeID users are able to see "All organizations compliance units:" and "Show transactions involving:" and a drop-down list of all organizations. This feature should **only** be available to users on the IDIR side. Please make this feature not visible or accessible to BCeID users.

Please see screenshot of what the feature should look like on the IDIR side:

Image

Please see screenshot of section that should not be visible to BCeID users:

Image

**Purpose**
The purpose of this function is to allow government users to easily find transactions that include a specific organization and their compliance unit balance. The reason this should not be visible to BCeID users is that it shares organization's names and all organizations compliance units; two pieces of information that should be restricted to IDIR use only to ensure privacy for all BCeID users.

**Acceptance Criteria**
- [ ] first
- [ ] second
- [ ] third

**Additional context**
- Add any other context about the task here.
- Or here

hamed-valiollahi:
This PR introduces the ability for IDIR users to filter transactions by organization on the Transactions page. This enhancement allows viewing and filtering of compliance units based on specific organizations, ensuring that these capabilities are restricted to IDIR users for privacy reasons.

@AlexZorkin, I have a suggestion regarding the backend unit tests. Since we agreed to add models to the database for each test, this means we need to add models for each layer of the API (repo, service, and view). Given that our application is not very complex to debug, and considering the effort required for developers to write backend tests three times for each scenario, also sometimes the responses from the repo and service layers depend on the user and their roles, do you think it might be a good idea to test just the endpoints instead of each individual layer?

Closes #543


AlexZorkin:
> This PR introduces the ability for IDIR users to filter transactions by organization on the Transactions page. This enhancement allows viewing and filtering of compliance units based on specific organizations, ensuring that these capabilities are restricted to IDIR users for privacy reasons.
> 
> @AlexZorkin, I have a suggestion regarding the backend unit tests. Since we agreed to add models to the database for each test, this means we need to add models for each layer of the API (repo, service, and view). Given that our application is not very complex to debug, and considering the effort required for developers to write backend tests three times for each scenario, also sometimes the responses from the repo and service layers depend on the user and their roles, do you think it might be a good idea to test just the endpoints instead of each individual layer?
> 
> Closes #543

Yes we could approach the backend tests in this way but there are cases especially in the service layer where you may want to test more atomically. Overall though I agree we should move away from the test file for each layer, and make it more contextual.

hamed-valiollahi:
> > This PR introduces the ability for IDIR users to filter transactions by organization on the Transactions page. This enhancement allows viewing and filtering of compliance units based on specific organizations, ensuring that these capabilities are restricted to IDIR users for privacy reasons.
> > @AlexZorkin, I have a suggestion regarding the backend unit tests. Since we agreed to add models to the database for each test, this means we need to add models for each layer of the API (repo, service, and view). Given that our application is not very complex to debug, and considering the effort required for developers to write backend tests three times for each scenario, also sometimes the responses from the repo and service layers depend on the user and their roles, do you think it might be a good idea to test just the endpoints instead of each individual layer?
> > Closes #543
> 
> Yes we could approach the backend tests in this way but there are cases especially in the service layer where you may want to test more atomically. Overall though I agree we should move away from the test file for each layer, and make it more contextual.

Thank you, Alex, for considering my suggestion! I appreciate you sharing your thoughts with me. :pray:

frontend/src/views/Transactions/Transactions.jsx
```diff
@@ -105,6 +106,13 @@ export const Transactions = () => {
     }
   }
 
+  const apiEndpoint = useMemo(() => {
```
prv-proton, line 109: Hi @hamed-valiollahi , just a suggestion we could use the filtering model of ag-grid to filter out the organizations like below. You can refer this in ViewOrganization component.
`  useEffect(() => {
    if (gridRef.current) {
      // clear any previous filters
      localStorage.removeItem(`${gridKey}-filter`)
      const statusFilter = gridRef?.current?.api?.getFilterInstance('isActive')
      if (statusFilter) {
        statusFilter.setModel({
          type: 'equals',
          filter: showActive ? 'Active' : 'Inactive'
        })
      }
      gridRef?.current?.api?.onFilterChanged()
    }
  }, [showActive])`

hamed-valiollahi: @prv-proton, Hi Prashanth, Thanks for reviewing my PR and for your valuable feedback. The approach you suggested is definitely better and follows best practices. Initially, I wanted to use AGGrid filtering to avoid changes to the backend. However, I found out that AGGrid can't use an OR condition between two columns. I need to filter transactions where either `toOrganizationId` or `fromOrganizationId` matches the selected organization. Since you have more experience with AGGrid, please correct me if I'm wrong. Thank you

prv-proton: Yes, you're right in the backend I've implemented and_ conditions only. But you can create your own custom fitlers as well. Probably in future you can refer to 
/lcfs/backend/lcfs/web/api/user/repo.py (apply_filters method)

hamed-valiollahi: @prv-proton, Thanks Prashanth for your comment. I mean that when I want to pass the `toOrganizationId` and `fromOrganizationId` from JavaScript to follow AGGrid filtering, I need to use [Advanced Filtering](https://www.ag-grid.com/javascript-data-grid/filter-advanced/), which is part of a [paid plan](https://www.ag-grid.com/license-pricing/). Also, another issue I've faced is that we don't have any `toOrganizationId` or `fromOrganizationId` columns available for filtering.

Example:
```js
const initialAdvancedFilterModel: AdvancedFilterModel = {
  filterType: "join",
  type: "AND",
  conditions: [
    {
      filterType: "join",
      type: "OR",
      conditions: [
        {
          filterType: "number",
          colId: "age",
          type: "greaterThan",
          filter: 23,
        },
        {
          filterType: "text",
          colId: "sport",
          type: "endsWith",
          filter: "ing",
        },
      ],
    },
    {
      filterType: "text",
      colId: "country",
      type: "contains",
      filter: "united",
    },
  ],
};
```

prv-proton: @hamed-valiollahi , this is just a filter schema which you'll pass from the frontend, and backend will decide how to handle the filtering as we're doing server-side rendering. If the columns aren't present then you can include them and hide by setting a flag in column definition.

hamed-valiollahi: @prv-proton , Awesome! Thanks a lot, Prashanth, for the clarification. I'll apply the changes. I appreciate you taking the time to leave comments and review my PR. 🙏

AlexZorkin (approved):

Issue #505, opened 2024-03-20
LCFS - Fix formatting issues with roles display in AG Grid user tables when compressing


justin-lepitzki:
**Describe the task**
The display formatting for roles pills in the AG Grid user tables is getting cut off when the window is compressed. We have done some recent work in this area by having the row expand vertically to properly display all of the roles pills. The roles pills are still getting cut off in some cases, or pushing in the next user role. We would like the row to expand adaqutely to allow for the display of all pills while not encroaching on the next row entry

**Screenshot of display issues:**
- ![Image](https://github.com/bcgov/lcfs/assets/131904954/6f68497f-ad5c-4cc8-90a8-079e7e5989b0)
- ![Image](https://github.com/bcgov/lcfs/assets/131904954/3e382caf-53f0-454d-b16b-8aa25a3c52d2)

**Purpose**
Fix formatting display issues 

**Acceptance Criteria**
- [ ] Ensure that roles pills in the AG Grid user tables are displayed adequately without being cut off, even when the window is compressed.
- [ ] Verify that the row expands vertically as necessary to properly display all roles pills without encroaching on the next row entry.
- [ ] Confirm that the display of roles pills remains consistent and does not push other elements or pills into adjacent rows.
- [ ] Test the behavior across various screen sizes and resolutions to ensure responsiveness and proper display of roles pills.
- [ ] Ensure that the solution does not negatively impact the overall layout or usability of the user tables.

**Additional context**
- Add any other context about the task here.
- Or here

Grulin:
@justin-lepitzki I think this issue has been fixed. Not sure we need this card:

BCeID user role view:


![Image](https://github.com/bcgov/lcfs/assets/156380046/1cc3af5e-bb39-491a-9d32-bd27fbbed1db)


![Image](https://github.com/bcgov/lcfs/assets/156380046/c6076500-3c08-4b26-971b-39c2d0501e29)



IDIR user role view:



![Image](https://github.com/bcgov/lcfs/assets/156380046/72a690db-fa43-4fa8-9759-21ffb590ef1d)

Grulin:
When signed in as an IDIR user, the formatting is working for the BCeID roles, but still cuts off additional roles for IDIR users. Please see screenshot:

![Image](https://github.com/bcgov/lcfs/assets/156380046/99560c8d-f561-479e-994e-3d4c3f558205)

PR #595, opened 2024-05-03
LCFS - Fix formatting issues with roles display in AG Grid user table…

prv-proton:
#505 -
 formatting issue fix for both Admin and Organization Users table.


AlexZorkin (approved):
More optimized!

Issue #574, opened 2024-04-17
LCFS - Backend data table to include Additional CI Attributable to Use (UCI) - gCO2e/MJ needed for calculating compliance units


Grulin:
**Context**
Backend data table(s) are needed to store the values that go in to formulas used for calculations in the compliance reporting feature of the LCFS portal. The two formulas used are:

Compliance units

![Image](https://github.com/bcgov/lcfs/assets/156380046/7b713f3e-b405-4b88-9133-6ea5d5f24f4b)

Target carbon intensity

![Image](https://github.com/bcgov/lcfs/assets/156380046/34de47f7-c15e-4eef-8ff4-b60a0ff941ea)

**Reference:**
[Calculating compliance units PDF](https://github.com/bcgov/lcfs/files/15068150/Calculating.compliance.units.2024-02-09.1.pdf)

**Description**
This card is for adding the values for the 'Additional CI Attributable to Use (UCI) - gCO2e/MJ' to the data table(s). In the compliance unit formula, UCIs are added to the recorded carbon intensity of the fuel to account for additional carbon intensity attributed to the use of the fuel. 

UCIs are organized by fuel type and end use in the following format: Fuel type | End use | EER

The following inputs should be included in the rows:

- LNG | Marine - General | 27.3
- LNG | Marine - Operated within 51 to 75% of load range | 17.8
- LNG | Marine - Operated within 76 to 100% of load range | 12.2
- LNG | Marine, w/ methane slip reduction kit - General | 10.6
- LNG | Marine, w/ methane slip reduction kit - Operated within 51 to 75% of load range  | 8.4
- LNG | Marine, w/ methane slip reduction kit - Operated within 76 to 100% of load range | 8.0
- LNG | Unknown | 27.3
- LNG | Other | 0
- Other | All uses | 0

**Purpose and Benefit to User:**

[...]

Issue #532, opened 2024-04-05
LCFS - Backend data table to include energy densities needed for calculating compliance units


Grulin:
**Context**
Backend data table(s) are needed to store the values that go in to formulas used for calculations in the compliance reporting feature of the LCFS portal. The two formulas used are:

Compliance units:

![Image](https://github.com/bcgov/lcfs/assets/156380046/729c9b96-fdd0-424b-abb3-255ae62dff11)

Target carbon intensity:

![Image](https://github.com/bcgov/lcfs/assets/156380046/9eeeab8b-2ee6-402a-9327-f7c5a0fc9210)

**Reference:**
[Calculating compliance units PDF](https://github.com/bcgov/lcfs/files/15068138/Calculating.compliance.units.2024-02-09.1.pdf)

**Description**
This card is for adding the values for 'Energy Densities' to the data table(s). Energy densities are determined in accordance with the regulations by fuel type and are included in the energy content (EC) portion of the compliance units formula (quantity x energy density = EC). 

There are three fuel types included that allow for custom input by the user to account for the energy density of fuel types we may not be aware of yet (other gasoline, other diesel and other jet). 

The energy densities are formatted by: Fuel type | Units | Quantity Units
*Units and quantity of units input by the user will have to handled in a way that allows for consistency and minimum administrative burden. 

The following inputs should be included in the rows for energy density:

- Fossil-derived gasoline | 34.69 | MJ/L
- Ethanol | 23.58 | MJ/L
- Renewable gasoline | 34.69 | MJ/L
- Renewable naphtha | 34.51 | MJ/L
- Other - gasoline category | user input | user input
- Fossil-derived diesel | 38.65 | MJ/L
- Biodiesel | 35.40 | MJ/L
- HDRD | 37.89 | MJ/L
- Other diesel fuel | 36.51 | MJ/L
- Other - diesel category | user input | user input
- Fossil-derived jet fuel | 37.40 | MJ/L
- Alternative jet fuel | 36.00 | MJ/L
- Other - jet fuel category | user input | user input
- Electricity | 3.60 | MJ/kWh
- Hydrogen | 141.76 | MJ/kWh
- Propane | 25.62 | MJ/L
- CNG | 38.27 | MJ/m3
- LNG | 53.54 | MJ/kg

**Purpose and Benefit to User:**

[...]

Issue #530, opened 2024-04-05
LCFS - Backend data table to include Energy Effectiveness Ratios (EERs) needed for calculating compliance units


Grulin:
**Context**
Backend data tables are needed to store the values that go in to formulas used for calculations in the compliance reporting feature of the LCFS portal. The two formulas used are:

1) Compliance units

![Image](https://github.com/bcgov/lcfs/assets/156380046/dfd1ee86-db61-4254-af65-e504d75a3631)

2) Target carbon intensity 

![Image](https://github.com/bcgov/lcfs/assets/156380046/096fbf0a-a4ab-4696-bf37-cd82b716b294)

**Reference:** 
[Calculating compliance units PDF](https://github.com/bcgov/lcfs/files/15068119/Calculating.compliance.units.2024-02-09.1.pdf)

**Description:**
This card is for adding the values for the 'Energy Effectiveness Ratios (EERs)' to the data table(s). EERs are dependent on fuel type, fuel category, and end use. Meaning, fuel types fall under one or more fuel category; each fuel type has one or more prescribed end uses which then have a corresponding EER value. This EER value is used in the compliance unit calculation formula. 

The following EERs are organized by fuel category in the format: Fuel type | End use | EER. For more information about the dependencies and relationships with respect to EERs (e.g., fuel type, fuel category, end use).

For the gasoline fuel category. the following inputs should be included in the rows:

- CNG | Any | 0.9
- Electricity | Light duty motor vehicle | 3.5
- Electricity | Other or unknown | 1.0
- Hydrogen | Fuel cell vehicle | 2.4
- Hydrogen | Other or unknown | 0.9
- Propane | Any | 0.9
- All other fuels | Any | 1.0

For the diesel fuel category, the following inputs should be included in the rows:

- CNG | Any | 0.9
- Electricity | Battery bus | 3.8
- Electricity | Battery truck | 3.2
- Electricity | Cargo handling equipment | 2.5
- Electricity | Fixed guiderail | 2.9
- Electricity | Ground Support equipment | 2.5
- Electricity | Heavy Forklift | 3.9
- Electricity | Marine | 2.5
- Electricity | Shore power | 2.8
- Electricity | Trolley bus | 2.4
- Electricity | Other or unknown | 1.0
- Hydrogen | Fuel cell vehicle | 1.8
- Hydrogen | Other or unknown | 0.9
- LNG | Compression-ignition engine | 1.0
- LNG | Other or unknown | 0.9
[...]

PR #600, opened 2024-05-08
Compliance units - fuel code data tables.

prv-proton:
- #530 
- #532 
- #574


AlexZorkin (approved):
Looks great. It will be good to get all these lookup tables completed.

PR #616, opened 2024-05-15
feat: fuel code auto suggest text

### LINKED ISSUE #434: LCFS - Auto-suggest next fuel code version and prepopulate data upon selection
**Description:**
When an IDIR user is adding a new fuel code and inputs the fuel code number into the 'Fuel code' field, the system suggests the next version of the code. This is achieved by scanning the database of existing codes to determine what the next version will be. See the "fuel code versioning" section below for more information on versioning. 

The user must select this next version, they are not able to indicate their own version of the code. We achieve this in TFRS by preventing decimals being used in the input field. 

Once the user selects the next version of the fuel code, the system should prepopulate some of the input fields with data from the previous version. See the "prepopulate" section below for more information. 

NOTE: This functionality exists in TFRS; we want to replicate it in the LCFS portal.

**Fuel code versioning:**
Fuel codes are assigned a prefix "BCLCF" and then given a unique 3+ digit number that follows sequential order. The first ever fuel code was BCLCF101.0. The '.0' end portion of the fuel code indicates its version. Fuel codes have effective periods when they are active, usually 1-3 years; after this period, the fuel code expires. The fuel producer is the one who is responsible for applying for fuel codes for their fuel production facilities; when a code is nearing expiry, the producer can apply for renewal of the fuel code. Once the application is processed, the renewed code will be given a new version number. For example, when BCLCF101.0 expires, the next version will be BCLCF101.1. 

**Prepopulate data when selecting the next version:**

The following fields should be prepopulated when the user selects the next version: 
- Prefix
- Company
- Fuel
- Feedstock
- Feedstock location
- Misc. (e.g. CCS, EOR, waste stream, non rendered, etc.)
- Fuel production facility location
- Feedstock transport mode
- Finished fuel transport mode
- Former company name

**Wireframe:**

https://preview.uxpin.com/59145260c7d8cc48969ac79401cc03d84843c3d9#/pages/165712417/simulate/sitemap

This auto-suggestion ensures data quality by mitigating the risk that IDIR users enter a version already in use or skips a version. Some expired codes are still used for late or supplemental reporting, so it is vital that each fuel code is distinct and doesn't repeat any previous versions. 

**Acceptance Criteria:**

- [ ] Given I am a BCeID user, When I am inputing a fuel code, I should be suggested the current available version of the fuel code
- [ ] Given I am a BCeID, When I am inputing a fuel code, I should not be able to input a decimal so i cannot type an arbitrary version.

**Development Checklist:**
- [ ] Create an endpoint that pulls fuel codes
- [ ] Possibly have filtering logic for the above endpoint
- [ ] Possibly use a fuzzy search library like fuse.js on the front end
- [ ] utility function to prevent decimal in input

**Notes:**
 TFRS already has this feature. reference TFRS to see how it was done there

kevin-hashimoto:
- feat: auto suggest text for fuel codes 'fuel code' field, closes #434

### DIFF
--- a/backend/lcfs/db/seeders/dev/feedstock_fuel_transfer_mode_seeder.py
+++ b/backend/lcfs/db/seeders/dev/feedstock_fuel_transfer_mode_seeder.py
@@ -1,0 +1,80 @@
+import logging
+from sqlalchemy import select, func
+from lcfs.db.models.FeedstockFuelTransportMode import FeedstockFuelTransportMode
+
+logger = logging.getLogger(__name__)
+
+
+async def seed_feedstock_fuel_transfer_modes(session):
+    feedstock_fuel_transfer_modes_to_seed = [
+        {
+            "fuel_code_id": 1,
+            "transport_mode_id": 5,
+        },
+        {
+            "fuel_code_id": 1,
+            "transport_mode_id": 4,
+        },
+        {
+            "fuel_code_id": 2,
+            "transport_mode_id": 3,
+        },
+        {
+            "fuel_code_id": 2,
+            "transport_mode_id": 2,
+        },
+        {
+            "fuel_code_id": 3,
+            "transport_mode_id": 1,
+        },
+        {
+            "fuel_code_id": 3,
+            "transport_mode_id": 5,
+        },
+        {
+            "fuel_code_id": 4,
+            "transport_mode_id": 4,
+        },
+        {
+            "fuel_code_id": 4,
+            "transport_mode_id": 3,
+        },
+        {
+            "fuel_code_id": 5,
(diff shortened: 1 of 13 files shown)


AlexZorkin (changes requested):
This is working well overall. I have one minor note: when typing into the fuel code field, the auto-suggest function operates effectively. But if you delete your input completely, the field auto-deselects and displays a warning. Is it possible to keep the field input active even after all the input is deleted?

kevin-hashimoto:
@AlexZorkin the auto validation issue should be fixed now

AlexZorkin (approved):

Issue #534, opened 2024-04-05
LCFS - Backend data table for determining carbon intensity needed for for compliance reporting calculation


Grulin:
**Context**
Backend data table(s) are needed to store the values that go in to formulas used for calculations in the compliance reporting feature of the LCFS portal. The two formulas used are:

Compliance units:

![Image](https://github.com/bcgov/lcfs/assets/156380046/1b28c62b-2b78-4d5c-bd9e-4746a5129df4)

Target carbon intensity:

![Image](https://github.com/bcgov/lcfs/assets/156380046/57f9c8e9-c0d0-4360-a450-ff2e5bece288)

**Reference:**
[Calculating compliance units PDF](https://github.com/bcgov/lcfs/files/15068209/Calculating.compliance.units.2024-02-09.1.pdf)

**Description**
The Low Carbon Fuels Act requires organizations to specify the specific provision in the legislation that is being relied upon to determine the carbon intensity of a supplied fuel. Base fuels (e.g., fossil-derived gasoline, diesel, jet fuel) only have one method for determining carbon intensity (the value is prescribed by the regulation) while the rest of the fuel types have two methods: (1) default carbon intensity, which are values prescribed in regulation, or (2) fuel codes, which are approved carbon intensity pathways. 

The options for determining carbon intensity use the following format: Fuel type | Provision 1 of the Act | Provision 2 of the Act (if applicable)

The following inputs should be included for determining carbon intensity:

Fossil-derived gasoline | Prescribed carbon intensity - section 19 (a) 
Ethanol | Fuel code - section 19 (b) (i) | Default carbon intensity - section 19 (b) (ii)
Renewable gasoline | Fuel code - section 19 (b) (i) | Default carbon intensity - section 19 (b) (ii)
Renewable naphtha | Fuel code - section 19 (b) (i) | Default carbon intensity - section 19 (b) (ii)
Fossil-derived diesel | Prescribed carbon intensity - section 19 (a) 
Biodiesel | Fuel code - section 19 (b) (i) | Default carbon intensity - section 19 (b) (ii)
HDRD | Fuel code - section 19 (b) (i) | Default carbon intensity - section 19 (b) (ii)
Other diesel fuel | Fuel code - section 19 (b) (i) | Default carbon intensity - section 19 (b) (ii)
Fossil-derived jet fuel | Prescribed carbon intensity - section 19 (a) 
[...]

Issue #531, opened 2024-04-05
LCFS - Backend data table to include default and prescribed carbon intensities needed for calculating compliance units


Grulin:
**Context**
Backend data table(s) are needed to store the values that go in to formulas used for calculations in the compliance reporting feature of the LCFS portal. The two formulas used are:

Compliance units:

![Image](https://github.com/bcgov/lcfs/assets/156380046/c44606e3-78a2-43f1-b79b-fc356c8941ad)

Target carbon intensity:

![Image](https://github.com/bcgov/lcfs/assets/156380046/e5460a48-ab97-4f6f-a4b0-e985ba51328d)

**Reference:**
[Calculating compliance units PDF](https://github.com/bcgov/lcfs/files/15068187/Calculating.compliance.units.2024-02-09.1.pdf)

**Description**

This card is for adding the values for 'Carbon intensities: default & prescribed (gCO2e/MJ)' to the data table(s). These values are used in the compliance units formula when determining the recorded carbon intensity (RCI) of the fuel. If a published carbon intensity for a fuel is unavailable, the fuel supplier may use the default carbon intensity determined in accordance with the regulations by fuel type. 

The carbon intensity for base fuels and their fossil-derived components remains prescribed in the regulations (Fossil-derived gasoline, fossil-derived diesel and fossil-derived jet).

The default and prescribed carbon intensities values are formatted by: Fuel type | Unit in gCO2e/MJ

The following inputs should be included in the rows for default and prescribed carbon intensities:

- Fossil-derived gasoline | 93.67
- Ethanol | 93.67
- Renewable gasoline | 93.67
- Renewable naphtha | 93.67
- Other - gasoline category | 93.67
- Fossil-derived diesel | 94.38
- Biodiesel | 100.21
- HDRD | 100.21
- Other diesel fuel | 100.21
- Other - diesel category | 100.21
- Fossil-derived jet fuel | 88.83
- Alternative jet fuel | 88.83
- Other - jet fuel category | 88.83
- Electricity | 12.14
- Hydrogen | 123.96
- Propane | 79.87
- CNG | 63.91
- LNG | 90.11

**Purpose and Benefit to User:**

[...]

PR #648, opened 2024-05-24
Backend data tables for provision of the act and default carbon inten…

prv-proton:
- #531 
- #534


backend/lcfs/db/models/ProvisionOfTheAct.py
```diff
@@ -0,0 +1,18 @@
+from sqlalchemy import Column, Integer, Float
+from lcfs.db.base import BaseModel, Auditable, DisplayOrder
+from sqlalchemy.orm import relationship
+from sqlalchemy import ForeignKey
+from sqlalchemy import String
+
+
+class ProvisionAct(BaseModel, Auditable, DisplayOrder):
```
AlexZorkin, line 8: I think we want to keep the consistency in naming, so this model file name could just be ProvisionAct.py I agree that it makes sense to shorten it to avoid the awkward field names.

AlexZorkin (changes requested):

backend/lcfs/db/models/ProvisionOfTheAct.py
prv-proton: renamed the file. Thanks for the review.

Issue #601, opened 2024-05-08
LCFS - New compliance report main page


Grulin:
**Description:**
On the BCeID side, create the main page of an annual compliance report that will build out the different sections. Other sections will be determined in other cards. 

The page should have the following:

- Breadcrumb at the top of the page to indicate this location: Home> Compliance reporting>[Compliance year] Compliance report
- Title of the page: [Compliance year] compliance report
- Include a grey box that shows the company name, address for service and address in B.C.
- Text that reads: "Did [fuel supplier] engage in any of the following activities between January 1, [Compliance year], and December 31, [Compliance year]?
- Text that reads: "Click links to report activities:"
- Links that will lead to the following report components:

> - Supply of fuel (e.g., gasoline, diesel, jet fuel, electricity, hydrogen etc.)
> - Final supply equipment (FSE) identification for electricity supply
> - Allocation agreements (e.g., allocating responsibility for fuel)
> - Notional transfer of eligible renewable fuel
> - Fuels for use other than for an applicable purpose (i.e. transportation or prescribed purposes) *Changed wording
> - Exporting fuel that was previously marketed in British Columbia

- The ability to upload supporting documents for your reports.
- A collapsible text box (default is expanded when first accessing page, collapses once user returns to the main report page. 
- Box will have the compliance report introduction wording in it. This is being finalized by the C&E team and will be updated once they have completed the wording. 
- At the bottom of the page, "Questions?" with the text "Please contact lcfs@gov.bc.ca with any questions you may have about this form"
- Legal disclaimer at the very end of the page , "This information does not replace or constitute legal advice. Users are responsible for ensuring compliance with the _Low Carbon Fuels Act_ and Regulations."

**Wireframe:**
Please see wireframe for design and content:
https://preview.uxpin.com/da68049c0af15edd7ce4e47e4bee431fed40c30d#/pages/173341051/simulate/sitemap

**Purpose and benefit to user:**
- The purpose of this page is to provide the user a main page for their compliance report that they can return to when adding information. It is the initial page displayed and can be saved as draft upon opening, only one report can be created for a compliance year.
- This also serves as the legal introduction to the report and provides resources to the supplier to reach out to the LCFS team if they have any questions. 
- This will also be the area where a supplier can attach supporting documents to their report to expedite processing. 

**Acceptance Criteria:**

[...]

PR #651, opened 2024-05-27
LCFS- New compliance report main page - 601

prv-proton:
#601


AlexZorkin (approved):
A great start on our compliance reports feature.

PR #667, opened 2024-06-03
fix: fuel code date fields

### LINKED ISSUE #623: LCFS - BUG - Last updated date field, Fuel codes
**Describe the Bug**
The Last updated date field should not be user input, it is a system generated value when an action is taken, Save, Approve etc. There is a missing field for Approval date which is a user input field. See screen shots.

**Expected Behaviour**
- Approval date is a user entered field
- Last updated date is a system generated value that is only displayed in display mode, not edit mode.
- There are no default values for any dates, they should all be null by default

**Implications**
- Users are not able to input an Approval date which is required for the fuel code to become active
- Last updated date should not be set by the user as this would not be accurate

**Steps To Reproduce**
User/Role: idir user
1. Go to Administration /  Fuel code / New fuel code
2. See error

existing error
Image

proper format
Image

kevin-hashimoto:
- fix: approval date and last updated date field fix on fuel code, closes #623


frontend/src/views/FuelCodes/AddFuelCode/_schema.jsx
```diff
@@ -291,22 +291,22 @@ export const fuelCodeColDefs = (t, optionsData) => [
...
-        {params.value ? params.value : 'YYYY-MM-DD'}
-      </Typography>
-    ),
-    editable: false, // TODO: change as per #516
-    cellDataType: 'dateString',
-    valueGetter: (params) => {
-      return new Date().toLocaleDateString()
-    }
-  },
+  // {
```
AlexZorkin, line 294: just delete this

AlexZorkin (changes requested):

PR #670, opened 2024-06-04
Feat: AG Grid Duplicate Row and Cleanup

AlexZorkin:
- cleaned up some of the ag grid components with better error handling and props definitions
- added the duplicate row feature to the input grid component


kevin-hashimoto (approved):
Works great!

Issue #673, opened 2024-06-07
LCFS - FSE input form


airinggov:
**Description**: As a BCeID user and supplier of electricity, I am required to report my final supply equipment (FSE). 

**Wireframe**: 
https://preview.uxpin.com/da68049c0af15edd7ce4e47e4bee431fed40c30d#/pages/177161091?mode=m

**Purpose and benefit to user**: The form enables BCeID users to report final supply equipment data, ensuring compliance with regulations. The duplicate row functionality saves time by allowing users to quickly create new entries based on existing data, while ensuring unique identification for each piece of equipment.

**Fields for Equipment Information Entry:**

All fields are mandatory except Notes.

- Dates of supply from: Date picker, must fall within the year of the report
- Dates of supply to: Date picker, must fall within the year of the report and must be later than the 'From' date
- Serial number: Alphanumeric, free-form text
- Manufacturer: Free-form text with predictive text from existing values
- Level of equipment: Select with 4 options:
  - Level 3 - Direct Current Fast Charging
  - Level 2 - High voltage, operating above Level 1
  - Level 1 - Low voltage, operating at 120V AC or less
  - Other
- Fuel measurement type: Select with 4 options:
  - Separate utility meter
  - Equipment meter (remote access)
  - Equipment meter (physical access)
  - No meter or estimated 
- Intended use: Multi-select with 12 options:
  - Light duty motor vehicle
  - Battery truck
  - Battery bus
  - Trolley bus
  - Fixed guiderail
  - Heavy forklift
  - Cargo handling equipment
  - Ground support equipment
  - Aircraft
  - Marine
  - Shore power
  - Other
- Street address: Free-form text
- City: Free-form text
- Postal code: Free-form text, validate postal code format
- Latitude: Numeric
- Longitude: Numeric
- Notes: Free-form text, 255 char max. Optional field.

**Duplicate Row Functionality**:
- Serial Number: Not copied
- Latitude: Not copied
- Longitude: Not copied
- All other fields are copied

**Auto-generation of Registration number**:
- Supplier ID: Forced to 5 digits by adding leading zeroes
- Postal code: Last 3 characters
- Sequential number: 001-999 this number will reset to 001 for each different postal code change

**Validation Rules**:
- Dates of Supply: Must be within the year of the report. The 'From' date must be earlier than the 'To' date. Both dates cannot be the same.

**Acceptance Criteria**:
Given I am a BCeID user
[...]

PR #706, opened 2024-06-25
LCFS - FSE input form #673

prv-proton:
- #673


frontend/src/views/FinalSupplyEquipments/_schema.jsx
```diff
@@ -0,0 +1,359 @@
...
+  api
+) => [
+  {
+    colId: 'action',
+    cellRenderer: 'actionsRenderer',
+    cellRendererParams: {
+      enableDuplicate: true,
+      enableEdit: false,
+      enableDelete: true,
+      onDuplicate: (props) => duplicateRow({ ...props, api }),
```
AlexZorkin, line 54: you don't nee to pass these methods through, just the api as a props

frontend/src/views/FinalSupplyEquipments/_schema.jsx
```diff
@@ -0,0 +1,359 @@
...
+      add: [newRow],
+      addIndex: props.node?.rowIndex + 1
+    })
+    props.api.stopEditing()
+  } else {
+    console.error('API is undefined')
+  }
+}
+
+const deleteRow = (props) => {
```
AlexZorkin, line 30: we don't need this method here as its handled in the actionsRenderer

frontend/src/views/FinalSupplyEquipments/_schema.jsx
```diff
@@ -0,0 +1,359 @@
+import { CommonArrayRenderer } from '@/utils/cellRenderers'
+import { suppressKeyboardEvent } from '@/utils/eventHandlers'
+import { Typography } from '@mui/material'
+import moment from 'moment'
+import { v4 as uuid } from 'uuid'
+
+// Copy the desired columns to new row
+const duplicateRow = (props) => {
```
AlexZorkin, line 8: we don't need this method here as its handled in the actionsRenderer

frontend/src/views/FinalSupplyEquipments/_schema.jsx
```diff
@@ -0,0 +1,359 @@
...
+    valueGetter: (params) => {
+      return [params.data.supplyFromDate, params.data.supplyToDate]
+    },
+    valueSetter: (params) => {
+      params.data.supplyFromDate = params.newValue[0]
+      params.data.supplyToDate = params.newValue[1]
+      return true
+    }
+  },
+  // {
```
AlexZorkin, line 112: commented code needed?

AlexZorkin (approved):

frontend/src/views/FinalSupplyEquipments/_schema.jsx
prv-proton: Thanks for the comments, it's fixed now.

frontend/src/views/FinalSupplyEquipments/_schema.jsx
prv-proton: Yes, I used the date range selector instead of two separate dates as given in the UX design. So after confirming from Al about which one goes well here, will delete the redundant code.

Issue #700, opened 2024-06-20
LCFS - Fuel supply, backend data table


airinggov:
https://preview.uxpin.com/da68049c0af15edd7ce4e47e4bee431fed40c30d#/pages/176471457?mode=m

### Description
- **Objective**: Ensure database schema alignment and extend the backend capabilities to include API endpoints for managing entries in the 'Fuel supply' data table.
- **Frontend Interaction**: Users will be able to add, update, and delete records through the frontend interface.

Here's the formatted description of the input form fields for the "Fuel supply" feature:

### Input Form Fields:
1. **Fuel Type**
   - **Input Type**: Drop-down menu
   - **Options**: from lookup table
   - **User Interaction**: No input allowed

2. ** Fuel Type Other**
   - **Input Type**: Free form text. 
   - **Condition**: Hidden by default. Appears only if 'Other' is chosen in the 'Fuel Type' drop-down menu.
   - **User Interaction**: Predictive text from existing values.
   - **Notes**: We might consider saving values input here in the Fuel type table with a flag to exclude them from the main drop down. This flag could be removed later to be ‘activated’ in the main list.

3. **Fuel Category**
   - **Input Type**: Drop-down menu
   - **User Interaction**: auto-selected locked field if only 1 available category, otherwise user must select, No input allowed

4. **End Use**
   - **Input Type**: Drop-down menu
   - **Options**: from lookup table
   - **User Interaction**: No input allowed

5. **Determining carbon intensity**
   - **Input Type**: Drop-down menu
   - **Options**: from lookup table
   - **User Interaction**: No input allowed

6. **Fuel code**
   - **Input Type**: Drop-down menu
   - **Options**: from lookup table
   - **User Interaction**: Enabled only if Approved fuel code is selected in Determining carbon intensity. No input allowed

7. **Quantity Supplied**
   - **Input Type**: Free form text
   - **Validation**: Numbers only, no decimals, cannot be negative

8. **Units**
   - **Input Type**: Drop-down menu
   - **User Interaction**: auto-selected locked field if only 1 available unit, otherwise user must select, No input allowed

9. ** Compliance Units**
   - **Field Type**: Free form text
   - **User Interaction**: None, auto-calculated value, locked field. Calculated using the Quantity Supplied, uses the formula set out in regulation, No input allowed.
   - **Notes**: Value saved in table as we don’t want these being continually calculated.

9. ** CI limit**
   - **Field Type**: Read only text
[...]

PR #743, opened 2024-07-09
LCFS - Fuel supply, backend data table #700

prv-proton:
- Fuel Supply backend API's (FuelSupply /table-options, /list-all, /save)
- Fuel Supply Model change to include more fields
- roles_required decorator was causing issues with fuel_supply for cyclic import, hence fixed the decorator import statement
- final supply equipments list-all endpoint wasn't exposed to government users, hence fixed it.


backend/lcfs/web/api/fuel_supply/repo.py
AlexZorkin, line 1: These are some complex queries with so many lookup tables! Well done parsing through them all. We will have to monitoring the performance.

AlexZorkin (approved):
Well done on managing the complicated relationships of this model.

Issue #726, opened 2024-07-03
LCFS - Fuel codes, Company and Contact Info


airinggov:
**Description:**
Implement predictive text functionality for Company, Contact name, and Contact email fields with specific behavior and dependencies.

**Wireframe:** N/A
TFRS can be used as an example of Company predictive text functionality to emulate.

**Purpose and benefit to user:**
This functionality enhances user experience by providing relevant suggestions based on existing data, reducing input errors, and improving efficiency when entering company and contact details.

**Acceptance Criteria:**

- [ ] Given I am a user, When I enter 3 characters in the Company field, then predictive text should start suggesting company names that contain those characters anywhere in the string.
- [ ] Given I am a user, When I have not entered a Company, then there should be no predictive text suggestions for Contact name or email.
- [ ] Given I am a user, When I enter a Company, then the Contact name field should be filtered to show only contacts associated with the entered Company.
- [ ] Given I am a user, When I have not entered a Contact name, then there should be no predictive text suggestions for Contact email.
- [ ] Given I am a user, When I enter a Contact name, then the Contact email field should show predictive text suggestions based on the entered Contact name.
- [ ] Given I am a user, When I enter new values in Company, Contact name, or Contact email fields, then the fields should accept these values without forcing predictive text suggestions.

**Development Checklist:**
- [ ] Implement predictive text functionality for the Company field to start after 3 characters are entered and match characters anywhere in the company names.
- [ ] Ensure no predictive text appears in Contact name or email fields until a Company is entered.
- [ ] Filter Contact name suggestions based on the entered Company.
- [ ] Ensure no predictive text appears in the Contact email field until a Contact name is entered.
- [ ] Allow all fields (Company, Contact name, Contact email) to accept new values without forcing predictive text suggestions.

**Notes:**
- Company is a mandatory field, Contact name and email are both optional
- Ensure predictive text functionality is efficient and does not cause performance issues.
- Test the dependencies between Company, Contact name, and Contact email fields to confirm correct behavior.
- Validate that new values can be entered freely in all fields, ensuring user flexibility.

Issue #725, opened 2024-07-03
LCFS - Fuel code numbers


airinggov:
**Description:**
When an IDIR user is adding a new fuel code, specific fuel code generation rules are applied to ensure unique and sequential fuel codes are assigned correctly.

**Wireframe:** 
https://preview.uxpin.com/da68049c0af15edd7ce4e47e4bee431fed40c30d#/pages/168739510?mode=m

**Purpose and benefit to user:**
This functionality ensures consistent and error-free generation of fuel codes, making it easier for IDIR users to manage and track fuel codes efficiently. (Fuel code should not behave as a full select list. Only a new increment is displayed as a single option and must be used.)

**Acceptance Criteria:**
- [ ] Given I am an IDIR user, When I am adding a new fuel code with the default prefix, then the prefix defaults to BCLCF and the code autofills the next available number ending in .0 (e.g., 420.0).
- [ ] Given I am an IDIR user, When I change the prefix to PROXY, then the fuel code updates to the next available number under that prefix ending in .0 (e.g., 003.0).
- [ ] Given I am an IDIR user, When I type an existing code, then the next available increment is forced (e.g., typing 113.5 autofills to 113.6), and certain fields from the previous code are auto-populated (as already implemented). 
- [ ] Given I am an IDIR user, When I click the duplicate button (e.g. 5 times on the 420.0 fuel code), then five rows are generated with sequential codes (e.g., 421.0, 422.0, 423.0, 424.0, 425.0) and certain fields from the original row are auto-populated.
- [ ] Given I am an IDIR user, If I delete the duplicated rows, the numbers will be released and can be used later.
- [ ] Given I am an IDIR user, When a draft fuel code is saved, then the prefix and fuel code fields are locked, but any other field can still be edited.

**Development Checklist:**
- [ ] Ensure the default prefix BCLCF is applied when adding a new fuel code.
- [ ] Implement logic to autofill the next available code ending in .0 for the default prefix BCLCF.
- [ ] Allow users to change the prefix to PROXY and update the fuel code to the next available number ending in .0
- [ ] Implement functionality to force the next available increment when an existing code is typed.
- [ ] Ensure specific fields are auto-populated from the previous code when an existing code is incremented.
- [ ] Enable multiple duplications by clicking the duplicate button and generate sequential codes with certain fields auto-populated.
- [ ] Lock the prefix and fuel code fields after saving a draft, allowing other fields to remain editable.
[...]

Issue #714, opened 2024-06-27
LCFS - FSE date range default


airinggov:
**Description:** Add a default date range for the FSE calendar select component. Have the start date as Jan 1 of the same year as the compliance report (not current year). Have the end date as Dec 31of the same year.

**Purpose and benefit to user:** This will be the most common date range so will make streamline the data entry.

**Acceptance Criteria:**

- [ ] Given I am a BCeID user with Compliance reporting role, When I am entering FSE, then date range selection component defaults to the full calendar year of the report, e.g. 2024 report: 2024-01-01 to 2024-12-31

**Development Checklist:**
- [ ] A  Ensure the FSE calendar select component can accept and display a default date range.
- [ ] B  Implement logic to determine the correct year based on the compliance report year.
- [ ] C  Set the default start date to January 1 of the report year.
- [ ] D  Set the default end date to December 31 of the report year.
- [ ] E  Update the component to dynamically adjust based on the compliance report year, not the current year.
- [ ] F  Test the component with various report years to ensure correct default date ranges are set.
- [ ] G  Ensure the component is accessible and usable for all BCeID users with the Compliance reporting role.

**Notes:**

- Confirm that the default date range does not override any user-selected dates.
- Verify the behavior when the report year is manually changed by the user.

airinggov:
When a user changes the default date range then uses the duplicate row function it should duplicate the user entered date range, not the full year default.

AlexZorkin:
tested and looks good now

airinggov:
Still getting this error so can't test card

![Image](https://github.com/user-attachments/assets/2d1ea030-f595-4cda-896c-91f6c0dc461f)

airinggov:
Changes to the default date range aren't being saved

Grulin:
The new row (4th row) is not saving, even after all the information is entered and validated. If I return to the view page, this input is still not saved:

![Image](https://github.com/user-attachments/assets/c6a7bba8-c4a2-4964-a458-f82cacccb1d9)

PR #766, opened 2024-07-18
Feat: LCFS - Fuel code numbers #725

prv-proton:
#725 - Fuel code numbers ticket.
#726 - Auto suggestions for company and contact details.
#714 - bug fix.
Fuel code data entry ag-grid refinement.

- Created a new version of BCDataGridEditorV2 so that we can have more control over each cell. instead of entire row editing this one chooses a cell by cell approach to perform edits.
- AsynsuggestionEditor is more refined for general use.


frontend/src/components/BCDataGrid/BCDataGridEditorV2.jsx
AlexZorkin, line 1: Do we want to switch out all existing BCDataGridEditors with this new V2 version?

AlexZorkin (changes requested):
- For fuel codes, when entering text into either the contact name, or contact email fields when the company field has nothing in it, the loading text still is showing. It would be nice to only show loading text when company is filled out. Similar functionality for name and email fields if company and name have values. 
- when you search and select a company, and then search a name, but then clear the company field, the search results for name and or email persist. The search result lists should be cleared or just not be shown when company is empty.
- There is a bug when trying to save a new fuel code. It gets stuck in 'updating' and there seems to be a database error. This is with a fresh database.
- Fuel code is required validation error shows when there is a fuel code value present, this is when first loading the new fuel code page

frontend/src/components/BCDataGrid/BCDataGridEditorV2.jsx
prv-proton: - 

> For fuel codes, when entering text into either the contact name, or contact email fields when the company field has nothing in it, the loading text still is showing. It would be nice to only show loading text when company is filled out. Similar functionality for name and email fields if company and name have values

. - **_I've removed the loading text completely_**
- when you search and select a company, and then search a name, but then clear the company field, the search results for name and or email persist. The search result lists should be cleared or just not be shown when company is empty.
- There is a bug when trying to save a new fuel code. It gets stuck in 'updating' and there seems to be a database error. This is with a fresh database.
- Fuel code is required validation error shows when there is a fuel code value present, this is when first loading the new fuel code page
- Do we want to switch out all existing BCDataGridEditors with this new V2 version?

PR #770, opened 2024-07-23
feat: AG Grid refinements 724

### LINKED ISSUE #724: LCFS - AG Grid refinements
**Describe the task**
Further refinements to the refactor work done

**Purpose**
This will bring more consistent and expected functionality for users

**Acceptance Criteria**

https://preview.uxpin.com/da68049c0af15edd7ce4e47e4bee431fed40c30d#/pages/178793572?mode=m

- [ ] Validation column only displays check, spinner or error, no other function, default is blank, greyed out cell, replace to use single checkmark icon
- [ ] No Save option, autosave should function like TFRS and be unnecessary as an option (use TFRS as hi-fi prototype)
- [ ] Action row should have heading ‘Action’, greyed out by default with icon(s) disabled, they become enable when the row is validated/saved
- [ ] Duplicate row icon is only used in Fuel codes and FSE input forms, not on other schedule forms
- [ ] Delete icon is available on all forms
- [ ] Duplicate row is not enabled when row is in error state, Delete row will be enabled
- [ ] Duplicate row will create a new row with certain fields prepopulated, some required fields will be null, validation should not run until the user starts interacting with the data
- [ ] Changelog is future development but shown for consideration of other requirements like tracking features of Add, Edit and Delete

**Additional context**
- Add any other context about the task here.

kevin-hashimoto:
Along with the main adjustments, further refinements were maid with the intention to improve scalability, maintainability and flexibility of the ag grid component and related components

- chore: upgrade ag grid deps
- new validations column for the grid
- refactored actions column for the grid
- new errors logic - uses server side errors to determine what errors to show on the frontend
- BCAlert2 component ( refactor of BCAlert )
  - `triggerAlert` now takes an object `{severity,  message}` directly to set the severity and message for the alert
  - component implementation is simplified
- new `BCGridBase`,  `BCGridViewer`, `BCGridEditor` components
  - removed some logic from the components and made it so that the business logic is written in the parent component.
  - reduced dependencies
  - standardized base component with defaults
  - new editor and viewer components with defaults
  - offers more flexibility and readability when implementing
  - editor has new `onAction` callback that returns which action ('delete', 'duplicate')was clicked so that its functions can be implemented
- `column.js` for common column defs
- renderes are directly used instead of registered components ( uses component instead of string values )
- fix errors in `AutocompleteEditor`
- new `ActionsRenderer2` component
- new `ValidationRenderer2` component
- removed `BCTable` component ( not used )

*implementation of these new components are only done in `AddFuelCode` and `FuelCodes` components*

closes #724


frontend/src/components/BCDataGrid/BCGridEditor.jsx
```diff
@@ -0,0 +1,129 @@
...
+
+  const handleOnCellValueChanged = useCallback(
+    (params) => {
+      if (!isEqual(params.oldValue, params.newValue)) {
+        params.data.modified = true
+      }
+      if (onCellValueChanged) {
+        onCellValueChanged(params)
+      }
+      params.api.refreshCells()
```
prv-proton, line 89: is it necessary to refreshCells here?

frontend/src/views/FuelCodes/AddFuelCode/AddFuelCode.jsx
```diff
@@ -58,284 +39,222 @@ const AddFuelCodeBase = () => {
...
-          (location) => location.fuelProductionFacilityCity === params.newValue
+      if (params.data.fuelProductionFacilityCity !== undefined) {
+        const location = optionsData.fpLocations.find(
+          (location) =>
+            location.fuelProductionFacilityCity ===
+            params.data.fuelProductionFacilityCity
         )
-        const updatedData = {
-          ...params.data,
+        updatedData = {
```
prv-proton, line 101: I feel this logic needs to be moved to valueSetter inside schema, what do you prefer?

frontend/src/components/BCDataGrid/BCGridEditor.jsx
kevin-hashimoto: I dont think we need that anymore. I think that was there to refresh any cell styling when i was experimenting. will remove

frontend/src/views/FuelCodes/AddFuelCode/AddFuelCode.jsx
kevin-hashimoto: Im not opposed to moving it to the valueSetter function.

My reasoning was to keep as much of the business logic in the parent component as possible instead of splitting it up as its sometimes harder to find or map out whats happening.

I think normally how AG Grid does it is that they have the coldefs in the same file as the parent component as a state, making it easy to centralize the business logic. So i guess because we are splitting up the coldefs to a different file, it feels like we are splitting the business logic.

It doesnt look like this block of logic has any dependencies other than the `optionsData` so it would be an easy move. Lets do that.

PR #817, opened 2024-08-13
Feat: Fuel Codes Updated for Aggrid V2 - 811

AlexZorkin:
- Moved add new rows feature into BcGridEditor
- Updated Fuel Code Components to use new ag grid structure
- Removed the new rows button components, and un-used actions components


frontend/src/views/FuelCodes/AddFuelCode/AddFuelCode.jsx
```diff
@@ -265,30 +216,17 @@ const AddFuelCodeBase = () => {
...
           defaultColDef={defaultColDef}
           onGridReady={onGridReady}
           rowData={rowData}
-          gridOptions={gridOptions}
-          onCellEditingStopped={onCellEditingStopped}
           onCellValueChanged={onCellValueChanged}
-          loading={isLoading}
+          onCellEditingStopped={onCellEditingStopped}
           onAction={onAction}
+          stopEditingWhenCellsLoseFocus
```
kevin-hashimoto, line 226: `stopEditingWhenCellsLoseFocus` will cause dialog/modal/popout windows to not function properly on the grid eg. date selection cells. 
It would be nice to have this turned on but I wasnt able to find a suitable solution :(

kevin-hashimoto (changes requested):

frontend/src/views/FuelCodes/AddFuelCode/AddFuelCode.jsx
AlexZorkin: gotcha thanks

Issue #822, opened 2024-08-13
LCFS - BUG - Button typo and refinment


airinggov:
**Describe the Bug**
Fix the typo in compliance
![Image](https://github.com/user-attachments/assets/d355a778-28c0-4c11-aa07-46e1e45ab4a1)

**Expected Behaviour**
The drop down should start at year 2024 and only increment one year past current year. So if current year is 2026 then the list will show 2024-2027. Any year that has an existing report should be greyed out or be inactive.

**Actual Behaviour**
The drop down gives options for 2010-2030. If you try create a duplicate report an error message is provided, which is good but would be better to prevent in the first place.

**Implications**
Reporting starts in LCFS for 2024 so should not go back earlier, we also don't want them creating report too far in the future.

**Steps To Reproduce**
Steps to reproduce the behaviour:
User/Role:

airinggov:
Functionality work looks great...but the typo is still there

Issue #756, opened 2024-07-15
LCFS - Calculate and display available compliance unit balance for the compliance period (Line 17 of the compliance report summary)


Grulin:
**Description:**
Implement functionality to calculate and display the available compliance unit balance for the compliance period. This involves pulling data from previous compliance reports to account for positive compliance units carried forward, subtracting any future debits recorded in the system, and ensuring the final balance is not negative.

**Wireframe:** [Link](https://preview.uxpin.com/da68049c0af15edd7ce4e47e4bee431fed40c30d#/pages/179634499/simulate/sitemap)

**Purpose and benefit to user:**
To provide users with an accurate and clear view of their available compliance unit balance, ensuring they can monitor and manage their compliance obligations effectively. This helps avoid penalties by ensuring users are aware of their unit balance status.

**Acceptance Criteria:**

- [ ] Given I am a BCeID user, When I view the compliance report, then I see the available compliance unit balance displayed in Line 17.
- [ ]  Given I am a BCeID user, When the system pulls data from previous compliance reports, then it accurately adds positive compliance units carried forward year by year and subtracts any future debits recorded in the system.
- [ ]  Given I am a BCeID user, When there is no previous compliance report available, then the system calculates the balance by taking the current positive compliance units reported minus the current negative compliance units reported and displays the remaining balance.
- [ ]  Given I am a BCeID user, When I view Line 17, then the balance displayed is never negative. If the calculation results in a negative balance, the system will instead display zero and trigger an automatic penalty for the current period.

**Development Checklist:**

- [ ]  Implement logic to pull data from previous compliance reports to account for positive compliance units carried forward and subtract any future debits recorded in the system.
- [ ]  Develop a fallback mechanism to calculate the balance using current compliance unit data if no previous report is available.
- [ ]  Ensure the balance displayed in Line 17 is never negative. If the balance is negative, display zero and trigger an automatic penalty.
- [ ]  Test data retrieval, calculation logic, and display to ensure accuracy and proper functionality.

**Notes:**

- Confirm the data retrieval process from previous compliance reports for accurate calculation.
- Ensure the system can handle cases with no previous compliance reports by correctly calculating the balance using current data.

Grulin:
Tested and this is currently still reflecting the balance of the total balance, rather than the available balance for this specific year (I was testing 2024). The available balance should only reflect what is available for this year (2024 for this example).

 This includes:
+ validations or compliance unit balance changes from assessments listed with a 2024 or prior compliance period
[...]

PR #857, opened 2024-08-27
Feat: Line-17 Calculate and display available compliance unit balance for the compliance period.

prv-proton:
#756 
#822 typo fix


AlexZorkin (approved):
Great calculation 👍

PR #862, opened 2024-08-28
Update Base Image + Windows Fixes

dhaselhan:
* Change scripts to use sh instead of bash as bash may not be present on the image
* Remove deprecated version tag from docker-compose


AlexZorkin (approved):
Builds great on mac too 👍

Issue #853, opened 2024-08-23
LCFS - Relocate the Compliance units column to be the first column in Fuel supply


airinggov:
**Description**:

Relocate the ‘Compliance units’ column to be the first column in both the ‘Fuel supply’ input and display tables.

Wireframe: 
https://preview.uxpin.com/da68049c0af15edd7ce4e47e4bee431fed40c30d#/pages/179802796/simulate/no-panels?mode=m

**Purpose and Benefit to User**:

Reordering the columns will make the 'Compliance units' information more prominent and easily accessible, aligning with its importance in the reporting process. This change simplifies navigation for users by placing key information upfront, thereby improving the overall usability of the tables.

**Acceptance Criteria**:

- [ ]  Given I am a user entering data in the ‘Fuel supply’ input table, when I view the table, then the ‘Compliance units’ column is positioned as the first column.
- [ ]  Given I am a user viewing the ‘Fuel supply’ display table, when I view the table, then the ‘Compliance units’ column is positioned as the first column.
- [ ]  Given I am a user working with the ‘Fuel supply’ tables, when I navigate through the columns, then the ‘Compliance units’ column appears before any other columns, maintaining its position consistently.

**Development Checklist**:

- [ ]  Identify the current position of the ‘Compliance units’ column in both the input and display tables for ‘Fuel supply’.
- [ ]  Update the table structure to relocate the ‘Compliance units’ column to the first position in the input table for ‘Fuel supply’.
- [ ]  Update the table structure to relocate the ‘Compliance units’ column to the first position in the display table for ‘Fuel supply’.
- [ ]  Test the changes to ensure that data entry and display functionalities are not affected by the reordering.

**Notes**:

- Ensure that the column relocation does not impact any sorting, filtering, or calculation functionalities associated with the 'Compliance units' data.

PR #869, opened 2024-08-29
Move compliance units to first column

dhaselhan:
* Fix error on initial load that data is not iterable

Closes https://github.com/bcgov/lcfs/issues/853


AlexZorkin (approved):
Looks good.

PR #880, opened 2024-09-04
Add Logout to Mobile Hamburger

dhaselhan:
* Add new route that fires log out keycloak function
* Add new route to the navbar that is only enabled on mobile size
* Add tests


AlexZorkin (approved):
Thanks for noticing and fixing this one!

Issue #852, opened 2024-08-23
LCFS - Relabel the CI limit column heading to Target CI


airinggov:
**Description**:

Relabel the ‘CI limit’ column heading to ‘Target CI’ in both input and display tables for the ‘Fuel supply’ as outlined in the wireframe.

**Wireframe**: 
https://preview.uxpin.com/da68049c0af15edd7ce4e47e4bee431fed40c30d#/pages/179802796/simulate/no-panels?mode=m
https://preview.uxpin.com/da68049c0af15edd7ce4e47e4bee431fed40c30d#/pages/182448168/simulate/no-panels?mode=m

**Purpose and Benefit to User**:

This change aims to align the terminology used in the new legislation and regulations.

**Acceptance Criteria**:

- [ ]  Given I am a user entering data in the ‘Fuel supply’ input table, when I view the columns, then I see the column labeled as ‘Target CI’ instead of ‘CI limit’.
- [ ]  Given I am a user viewing the ‘Fuel supply’ display table, when I view the columns, then I see the column labeled as ‘Target CI’ instead of ‘CI limit’.

 **Development Checklist:**

- [ ]  Update the column heading from ‘CI limit’ to ‘Target CI’ in the input table for ‘Fuel supply’.
- [ ]  Update the column heading from ‘CI limit’ to ‘Target CI’ in the display table for ‘Fuel supply’.
- [ ]  Ensure both frontend and backend naming conventions are in sync and if necessary introduce the migrations for table name/ column name changes.
- [ ]  Update test cases to reflect the name change.

**Note**:
This may also apply to the backend data table for Fuel supply and the newly set up Export fuel table

PR #884, opened 2024-09-05
feat: Rename CI Limit to Target CI

dhaselhan:
* Rename in models, generate migration
* Update label and rename in front end

Closes: https://github.com/bcgov/lcfs/issues/852


AlexZorkin (approved):
Looks good.

Issue #848, opened 2024-08-22
LCFS - Compliance report schedules refinements


airinggov:
Enhance the user experience for all compliance report schedules by implementing the following features:

Add a ‘Save & return to report’ button to all compliance report schedules. The button will return users to the main report page. If any input table row is in an error state, display a pop-up modal with the message, “Incomplete data will not be saved, cancel to fix.” The modal should have two buttons: Cancel and Return to report.

Update the Add row feature to delay validation. The red error icon should not be displayed immediately upon adding a new row. Validation should only occur for each row as the user interacts with the data within that row.

**Purpose and Benefit to User:**

These updates will improve the user experience by preventing premature validation errors and providing clear guidance when attempting to save incomplete data. This approach ensures that users can focus on completing each row without unnecessary interruptions, thereby reducing frustration and potential data entry errors.

- Acceptance Criteria:

- [ ]  Given I am a user, when I click ‘Save & return to report’ on any compliance report schedule, then I am returned to the main report with any new information displayed.
- [ ]  Given I am a user, when I click ‘Save & return to report’ on any compliance report schedule, then I am prompted with a modal if there is an error in any table row, warning that incomplete data will not be saved.
- [ ]  Given I am a user, when I see the pop-up modal, then I can choose between Cancel (to fix the errors) or Return to report (to proceed without saving the incomplete data) and return to report.
- [ ]  Given I am a user, when I use the Add row feature in any schedule input table, then the red error icon is not displayed immediately, and validation only occurs after I interact with the row data.
- [ ]  Given I am a user, when I interact with a row and it contains errors, then the validation process triggers and displays the red error icon.

**Development Checklist:**

- [ ]  Add a ‘Save & return to report’ button to all compliance report schedules.
- [ ]  Implement a pop-up modal that appears when saving with incomplete data, including the specified message and buttons.
- [ ]  Update the Add row feature to delay error validation until the user interacts with the data in the row.
- [ ]  Ensure that validation logic is properly implemented for each row based on user interaction.
- [ ]  Test the new features across all compliance report schedules to confirm correct behavior and error handling.

**Notes:**

Ensure the pop-up modal is consistent in design with other modals in the application.

PR #888, opened 2024-09-06
feat: Add Save Return to Compliance Pages

dhaselhan:
* Add a save and return button beneath all tables in Compliance Reporting
* Expand BCGrid to support this functionality since validity of rows impacts how it works
* Remove Callable Model, looks  to be an old dead modal built using bootstrap
* Fix incorrect route constant
* Remove deprecated reactiveCustomComponents attribute

Closes: https://github.com/bcgov/lcfs/issues/848


frontend/src/components/BCDataGrid/BCGridBase.jsx
```diff
@@ -33,7 +33,6 @@ export const BCGridBase = forwardRef(({ autoSizeStrategy, ...props }, ref) => {
         suppressDragLeaveHidesColumns
         suppressMovableColumns
         suppressColumnMoveAnimation={false}
-        reactiveCustomComponents
```
AlexZorkin, line 36: We have to keep this option to allow our inputs to commit values. Def not ideal but until we find a work around lets leave this here.

AlexZorkin (changes requested):
Looks good, just one minor change to include the reactiveCustomComponents option. Please also merge release into this branch again. ty

frontend/src/components/BCDataGrid/BCGridBase.jsx
dhaselhan: That's odd, its marked as deprecated in their documentation as default on. Are we on an older version?

AlexZorkin: we're running v32.0.2 and there is a v32.1.0
I have tested without this option and the inputs break so we'll have to investigate further. For this ticket lets just leave it in for now.

dhaselhan: Done, added back

AlexZorkin (approved):

Issue #864, opened 2024-08-28
LCFS - Fuel supply refinements


airinggov:
**Description**:
1. The "Fuel type other," "Fuel codes," and "End use" columns are not saving correctly.

2. A new error validation is required to flag duplicate fuels (with the same Type, Category, and Determine CI/Fuel codes). These duplicates need to be identified and flagged as an error, with the following message displayed: “There are duplicate fuel entries, please combine the quantity into a single value on one row.”

**Purpose and benefit to user**:
Ensuring that the "Fuel type other," "Fuel codes," and "End use" columns save correctly will prevent data loss and improve data integrity.

Flagging duplicate fuel entries will help users avoid errors and ensure accurate reporting by combining quantities into a single entry.

**Acceptance Criteria**:

- [ ]  Given I am a BCeID user, When I enter data into the "Fuel type other," "Fuel codes," or "End use" columns, then the data should save correctly without any loss.
- [ ]  Given I am a BCeID user, When I enter duplicate fuels with the same Type, Category, and Determine CI/Fuel codes, then an error message should be displayed: “There are duplicate fuel entries, please combine the quantity into a single value on one row.”

**Development Checklist**:

- [ ]  Investigate and fix the issue preventing data in the "Fuel type other," "Fuel codes," and "End use" columns from saving correctly.
- [ ]  Implement validation to detect duplicate fuels based on Type, Category, and Determine CI/Fuel codes.
- [ ]  Display the specified error message when duplicate fuels are detected.
- [ ]  Test the functionality to ensure that data is saving correctly and duplicates are flagged appropriately.

PR #892, opened 2024-09-10
feat: Improve Fuel Codes and Tables

dhaselhan:
* Fix saving issues with Fuel Code and Fuel Code Other
* Update Fuel codes code to be a suffix, since the final fuel code is calculated
*Add a fuel code that can be used for display
* Use new display property to align dropdown options and values
* Better names for Schema validation for Fuel Supply
* Add missing required stars to Add Fuel Code table
* Clean up and refactor row error validation to a more standard format
* Add logic to handle and display warnings

Closes: #864


AlexZorkin (changes requested):
I added a few fixes on the details object and various schedules. There's one more bug we have to fix for allocation agreements, where the fuel_code field is sending back the prefix+suffix, but is then breaking when it looks up the fuel_code by suffix only.

/backend/lcfs/web/api/allocation_agreement/services.py
        fuel_code = await self.fuel_repo.get_fuel_code_by_name(
            allocation_agreement.fuel_code
        )
We will have to parse out the suffix from the prefix+suffix to lookup by name.

Otherwise this is looking done. ty

dhaselhan:
> I added a few fixes on the details object and various schedules. There's one more bug we have to fix for allocation agreements, where the fuel_code field is sending back the prefix+suffix, but is then breaking when it looks up the fuel_code by suffix only.
> 
> /backend/lcfs/web/api/allocation_agreement/services.py fuel_code = await self.fuel_repo.get_fuel_code_by_name( allocation_agreement.fuel_code ) We will have to parse out the suffix from the prefix+suffix to lookup by name.
> 
> Otherwise this is looking done. ty

Got the query updated and fixed some issues with the errors fields.

AlexZorkin (approved):
Thank you for the updates. 👍

PR #901, opened 2024-09-11
fix: Run Black Format

dhaselhan:
* Format the backend python


AlexZorkin (approved):
thanks for running this.

PR #918, opened 2024-09-16
fix: Add Missing Required Headers

dhaselhan:
* Adjust schema order to match UI so validation follows left to right
* Rename Header component to RequiredHeader as that is currently the difference

Closes: #915


AlexZorkin (approved):
thanks for tracking down the missing required fields.

PR #919, opened 2024-09-16
fix: currency dollar symbol line 11

### LINKED ISSUE #738: LCFS - Calculate and Display Non-Compliance Penalty Payable (Line 11 of report summary)
Image

**Description:**
Implement functionality to calculate and display the automatic non-compliance penalty payable for each fuel class (Gasoline, Diesel, Jet Fuel) in the compliance report. This involves applying the prescribed penalty rates to the difference between the required and net volume of renewable fuel supplied for each class.

Will need to add total renewable requirement penalty at some point. Will be dealt with in another card. 

**Wireframe:** [link](https://preview.uxpin.com/da68049c0af15edd7ce4e47e4bee431fed40c30d#/pages/169090266/simulate/sitemap)

To provide users with a clear calculation of the automatic non-compliance penalty payable, ensuring they are aware of the financial implications of not meeting renewable fuel obligations.

**Acceptance Criteria:**
- [ ]  Given I am a BCeID user, When I view the compliance report summary, then cells G11, D11, and J11 display the non-compliance penalty payable for the Gasoline, Diesel, and Jet Fuel classes, respectively.
- [ ]  Given I am a BCeID user, When the formula is applied, then cell G11 correctly calculates the penalty using the formula (G4 - G10) x $0.30.
- [ ]  Given I am a BCeID user, When the formula is applied, then cell D11 correctly calculates the penalty using the formula (D4 - D10) x $0.45.
- [ ]  Given I am a BCeID user, When the formula is applied, then cell J11 correctly calculates the penalty using the formula (J4 - J10) x $0.50.

**Development Checklist:**
- [ ]  Implement the calculation formula for the Gasoline class in cell G11: (G4 - G10) x $0.30.
- [ ]  Implement the calculation formula for the Diesel class in cell D11: (D4 - D10) x $0.45.
- [ ]  Implement the calculation formula for the Jet Fuel class in cell J11: (J4 - J10) x $0.50.
- [ ]  Ensure that the calculations are accurate and the results are correctly displayed in the respective cells.
- [ ]  Test the calculations thoroughly to ensure they accurately reflect the penalty amounts based on the provided formulas.

**Notes:**
- Confirm the data in the referenced cells (G4, G10, D4, D10, J4, J10) is accurate and up-to-date.
- Coordinate with the development team to ensure the formula implementation aligns with the overall compliance reporting structure.

kevin-hashimoto:
- adds currency symbol for cells that are currency format
- adds the 'currency' format to line 11, closes #738

### DIFF
--- a/frontend/package-lock.json
+++ b/frontend/package-lock.json
@@ -2344,70 +2344,6 @@
         "node-hook": "^1.0.0"
       }
     },
-    "node_modules/@badeball/cypress-configuration/node_modules/@esbuild/aix-ppc64": {
-      "version": "0.19.12",
-      "resolved": "https://registry.npmjs.org/@esbuild/aix-ppc64/-/aix-ppc64-0.19.12.tgz",
-      "integrity": "sha512-bmoCYyWdEL3wDQIVbcyzRyeKLgk2WtWLTWz1ZIAZF/EGbNOwSA6ew3PftJ1PqMiOOGu0OyFMzG53L0zqIpPeNA==",
-      "cpu": [
-        "ppc64"
-      ],
-      "dev": true,
-      "optional": true,
-      "os": [
-        "aix"
-      ],
-      "engines": {
-        "node": ">=12"
-      }
-    },
-    "node_modules/@badeball/cypress-configuration/node_modules/@esbuild/android-arm": {
-      "version": "0.19.12",
-      "resolved": "https://registry.npmjs.org/@esbuild/android-arm/-/android-arm-0.19.12.tgz",
-      "integrity": "sha512-qg/Lj1mu3CdQlDEEiWrlC4eaPZ1KztwGJ9B6J+/6G+/4ewxJg7gqj8eVYWvao1bXrqGiW2rsBZFSX3q2lcW05w==",
-      "cpu": [
-        "arm"
-      ],
-      "dev": true,
-      "optional": true,
-      "os": [
-        "android"
-      ],
-      "engines": {
-        "node": ">=12"
-      }
-    },
-    "node_modules/@badeball/cypress-configuration/node_modules/@esbuild/android-arm64": {
-      "version": "0.19.12",
-      "resolved": "https://registry.npmjs.org/@esbuild/android-arm64/-/android-arm64-0.19.12.tgz",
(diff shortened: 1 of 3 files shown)


frontend/src/views/ComplianceReports/components/SummaryTable.jsx
```diff
@@ -195,7 +195,7 @@ const SummaryTable = ({
                       }}
                     >
                       {row?.format &&
-                      row[column.id] &&
+                      row[column.id] !== undefined &&
```
dhaselhan, line 198: question: Looks like this change is to enable rows with null ids to use the formatters. Why do we want to prevent formatting on undefined IDs but allow it on null?

frontend/package-lock.json
dhaselhan, line 1: I don't see any comments in the MR about doing an upgrade or other commands. Why is the package lock changing?

kevin-hashimoto: was doing a fresh rebuild of my containers since it was behind by a few weeks after vacay and my frontend docker build kept crashing after rebuilding unless I did a lock update. the fix would suggest there was a problem with a third party lib

frontend/src/views/ComplianceReports/components/SummaryTable.jsx
kevin-hashimoto: 0 values come in as falsy so it never triggered the currency formatting ( this is what was preventing currency formatting for all cells that were currency cells ). we can probably change this to check for 0 values instead of a blanket statement of undefined. thoughts?


`(row[column.id] || row[column.id] === 0)`

dhaselhan: I like that a lot better, the 0 is much clearer for what we are trying to avoid. I thought it was for null but it was 0 all along.

frontend/package-lock.json
dhaselhan: Ah odd, good to know.

dhaselhan (approved):
LGTM 👍

frontend/src/views/ComplianceReports/components/SummaryTable.jsx
kevin-hashimoto: changed to `(row[column.id] || row[column.id] === 0)`

PR #961, opened 2024-09-23
fix: Remove fixed width of NotionalTransfer and OtherUses

dhaselhan:
* Only two tables with fixed widths, that forces them off the page at smaller screen size
* Update them to use translation file
* Fix Commmon -> Common *


AlexZorkin (approved):
Nice clean up 👍

PR #964, opened 2024-09-23
Chore: Squash Migrations - 936

AlexZorkin:
- added constraint naming conventions and ensured all constraints have defaults names when using autogenerate
- added and exclude list to alembic env so that materialized views and views can be ignored during autogenerated revisions
- squashed all migrations down into initial migration
- created a separate migration to handle all current materialized views and triggers


backend/lcfs/db/migrations/env.py
```diff
@@ -22,23 +22,32 @@
...
-# from myapp import mymodel
 target_metadata = Base.metadata
 
-# other values from the config, defined by the needs of env.py,
-# can be acquired:
-# my_important_option = config.get_main_option("my_important_option")
-# ... etc.
-
+# Exclude specific tables (views and materialized views) from autogenerate
+exclude_tables = [
```
dhaselhan, line 30: praise: Yay! Cleaner migrations for everyone!

dhaselhan (approved):
LGTM 👍

Issue #895, opened 2024-09-11
LCFS - Refine the renewable requirement summary section to exclude certain fuel types


Grulin:
**Describe the task**
Refine the Renewable requirement summary section of the compliance report to ensure that certain fuel types are excluded from the calculations. Specifically, lines 1, 2, 5, 6, and 7 should not include the following fuel types: Electricity, Hydrogen, Propane, CNG (Compressed Natural Gas), and LNG (Liquefied Natural Gas), as these are not subject to the renewable requirement under the Low Carbon Fuels Act. Instead, these fuel types are only subject to the low carbon fuel requirement.

**Purpose**
To ensure the Renewable requirement summary accurately reflects the fuel types that are subject to the renewable requirement and correctly distinguishes those that are only subject to the low carbon fuel requirement, in compliance with the Low Carbon Fuels Act.

**Acceptance Criteria**

- [ ]  Line 1 does not include Electricity, Hydrogen, Propane, CNG, and LNG.
- [ ]  Line 2 does not include Electricity, Hydrogen, Propane, CNG, and LNG.
- [ ] Line 5 does not include Electricity, Hydrogen, Propane, CNG, and LNG.
- [ ]  Line 6 does not include Electricity, Hydrogen, Propane, CNG, and LNG.
- [ ]  Line 7 does not include Electricity, Hydrogen, Propane, CNG, and LNG.
- [ ]  The summary clearly indicates that these fuel types are only subject to the low carbon fuel requirement.
- [ ]  The summary is reviewed to ensure alignment with the Low Carbon Fuels Act.

**Additional context**

dhaselhan:
@Grulin How do we filter Line 5 (notional transfers)? Currently notional transfers do not have a fuel type associated to them only a category.

dhaselhan:
@Grulin For the last requirement "The summary clearly indicates that these fuel types are only subject to the low carbon fuel requirement." Where in the UI do we want to indicate which types are subjected?

Issue #896, opened 2024-09-11
LCFS - Refine line 6 of the compliance report summary conditions


Grulin:
**Describe the task**
Update line 6 of the "Compliance Report Summary" section to allow for input under the following condition:
a) When an organization has supplied specific types of renewable fuels, categorized as:

- Gasoline Category: Renewable Gasoline, Ethanol, Renewable Naphtha.
- Diesel Category: Biodiesel, HDRD (Hydrotreated Renewable Diesel), Other Diesel Fuel.
- Jet Fuel Category: Alternative Jet Fuel.

and 

b) When an organization has exceeded the renewable volume requirement on line 4 of the summary

**Purpose**
To enhance line 6 of the "Compliance Report Summary" to allow a supplier to retain a surplus of renewable fuel content, up to 5% of line 4, provided that they have exceeded the renewable requirement specified on line 4 of the summary. 

**Acceptance Criteria**

- [ ]  Line 6 allows for input only when the organization has exceeded the renewable requirement specified on line 4 of the summary.
- [ ]  Line 6 provides an option for suppliers to retain a surplus of renewable fuel content, up to a certain percentage.
- [ ]  Ensure input fields are correctly categorized and validated for compliance reporting purposes.
- [ ]  The updated line 6 is reviewed and confirmed to meet regulatory requirements and aligns with the Low Carbon Fuels Act.

**Additional context**
Validate that exceeding the renewable requirement on line 4 triggers the input availability and retaining options on line 6.

Grulin:
Card #895 should be completed first

dhaselhan:
@Grulin Do I need to exceed the requirements by the amount I am going to retain? e.g. Required (line 4) is 1,000 and I exceed it (line 5) with 1,020. Can I still enter 50 in line 6 or is line 6 capped at 20? (e.g. Line 6 is capped at the minimum of Line 5 - Line 4 or 5% of line 4)

PR #976, opened 2024-09-24
feat: Enable saving of Line 6 without overwriting

dhaselhan:
Previously any changes to fuel supply or other fields would cause line 6 to be overwritten. The endpoint was also being called twice.

This change aims to start fixing the issue, by updating the get endpoint to IDEALLY only re-calculate if something has changed however this is not currently working.

Changes:
* Unify get and put to have the same structure, Report -> Summary is 1:1 so there is no need to pass in ID
* Remove  need to pass in summary_id for get as there can only ever be one per report
* Begin to clean up line calculation code to use less / clear checks
* Update useComplianceReport to not invalidate its own data and instead set it
* Fix typos

closes: #895 #896

### DIFF
--- a/backend/lcfs/web/api/compliance_report/repo.py
+++ b/backend/lcfs/web/api/compliance_report/repo.py
@@ -151,12 +151,14 @@ class ComplianceReportRepository:
                 joinedload(ComplianceReport.compliance_period),
                 joinedload(ComplianceReport.current_status),
                 joinedload(ComplianceReport.summary),
-                joinedload(
-                    ComplianceReport.fuel_supplies
-                ),
+                joinedload(ComplianceReport.fuel_supplies),
                 joinedload(ComplianceReport.other_uses),
-                joinedload(ComplianceReport.history).joinedload(ComplianceReportHistory.status),
-                joinedload(ComplianceReport.history).joinedload(ComplianceReportHistory.user_profile),
+                joinedload(ComplianceReport.history).joinedload(
+                    ComplianceReportHistory.status
+                ),
+                joinedload(ComplianceReport.history).joinedload(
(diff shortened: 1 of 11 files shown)


backend/lcfs/web/api/compliance_report/repo.py
```diff
@@ -541,32 +551,40 @@ async def add_compliance_report_summary(
...
-            raise ValueError(f"No summary found with ID {summary_id}")
+            raise ValueError(f"No summary found with report ID {report_id}")
 
         # Update renewable fuel target summary
         for row in summary.renewable_fuel_target_summary:
             line_number = row.line
             for fuel_type in ["gasoline", "diesel", "jet_fuel"]:
                 column_name = f"line_{line_number}_{row.field.lower()}_{fuel_type}"
-                setattr(summary_obj, column_name, int(getattr(row, fuel_type) or 0))
+                setattr(
```
dhaselhan, line 574: Main change of this MR, these would before default to 0. Now we copy over the users previously entered value and then the calculate function will reset it to the max if the max has changed.

backend/lcfs/web/api/compliance_report/summary_service.py
```diff
@@ -360,7 +398,7 @@ def calculate_renewable_fuel_target_summary(
 
         # line 4
         if 2024 <= compliance_period <= 2027:
-            jet_fuel_percentage = 0 / 100
```
AlexZorkin, line 363: This is a fun fix.

AlexZorkin (approved):
Looks good. As discussed it would be great to refactor the line prefixes and rethink summary calculations a bit. Please make some notes about the known issues so we can draft up some tech debt cards. tyty

Issue #886, opened 2024-09-05
LCFS - Task - Export Organizations and Users from TFRS into LCFS


airinggov:
**Describe the task:** 
Using Apache Nifi, build a reusable script to export Organization and User data (BCeID & IDIR) from TFRS (Transportation Fuel Reporting System) to LCFS (Low Carbon Fuel System).

**Purpose:**
To facilitate the migration of Organization and User data from TFRS to LCFS, ensuring that relevant data is mapped correctly while maintaining historical consistency with GUIDs and unique IDs. 

**Acceptance Criteria:**
**Organization data**
- [ ] Write a reusable script in Apache Nifi to export Organization data from TFRS to LCFS.
- [ ] Ensure Organization data fields map correctly, noting that LCFS-Operating Name, LCFS-Email Address, and LCFS-Phone Number will be null as there is no corresponding data in TFRS.
- [ ] Map TFRS-Head Office Address to LCFS-Address for Service. (Street address, Address other, City, Province, Country, Postal code all map over directly 1 to 1.)
- [ ] Map TFRS-Corporation/Attorney in B.C. Address to LCFS-Address in B.C. (Street address, Address other, City, Province, Country, Postal code all map over directly 1 to 1.)
- [ ] Exclude TFRS-Name of Representative from LCFS, it doesn't exist in LCFS.
- [ ] Retain Organization GUID or unique ID to maintain historical data.

**BCeID User data**
- [ ] Write a reusable script in Apache Nifi to export BCeID Users (both Active and Inactive) and roles from TFRS to LCFS.
- [ ] Map user data as follows:
- TFRS-First Name/ LCFS-First name 
- TFRS-Last Name/ LCFS-Last name 
- TFRS-Email address associated with the BCeID/ LCFS-Email address associated with the BCeID
- TFRS-Alternate email address/ LCFS-Alternate email
- TFRS-BCeID/ LCFS-BCeID Userid
- TFRS-Title/ LCFS-Job title
- TFRS-Work Phone/ LCFS-Phone
- TFRS-Mobile Phone/ LCFS-Mobile phone
- TFRS-Status/ LCFS-Status
- [ ] Map the BCeID roles data as follows:
- TFRS-Managing Users/LCFS-Manage Users
- TFRS-Credit Transfers/LCFS-Transfer
- TFRS-Signing Authority/LCFS-Signing Authority
- TFRS-Compliance Reporting/LCFS-Compliance Reporting
[...]

PR #981, opened 2024-09-26
Feat: LCFS - Task - Export Organizations and Users from TFRS into LCFS #886

prv-proton:
#886

- added network registry and zookeeper to persist the flow, so that we don't need to import the templates everytime. All the flows will be loaded as soon as the application is up and running.
- changed the docker-compose LCFS and TFRS database from the main application compose file, instead of creating a separate database.
- implement data transfer logic from openshift database to local and vice versa.


docker-compose.yml
```diff
@@ -1,3 +1,4 @@
+version: '3'
```
dhaselhan, line 1: Version is deprecated

docker-compose.yml
```diff
@@ -69,11 +76,17 @@ services:
...
   postgres_data:
     name: lcfs_postgres_data
   redis_data:
     name: lcfs_redis_data
   node_modules:
-    name: lcfs_node_data
\ No newline at end of file
+    name: lcfs_node_data
+
+networks:
```
dhaselhan, line 90: question: Why is this being changed?

etl/nifi/conf/bootstrap-aws.conf
```diff
@@ -0,0 +1,27 @@
+#
```
dhaselhan, line 1: Why do we need an AWS config file?

dhaselhan (changes requested):
Can you please add some comments to your PR to explain your changes? There are a lot of odd files inlcuding .gz and binary files and its not clear to me why we need them.

docker-compose.yml
prv-proton: sorry missed. Now updated the PR comments detailing all the changes.

etl/nifi/conf/bootstrap-aws.conf
prv-proton: These are some config file which I copied directly from the nifi container, so that we can persist the changes. Not all config file will be required, but while copying just copied everything to the repo. Please let me know if you want me to remove it.

dhaselhan (approved):
LGTM 👍

PR #1001, opened 2024-10-01
Add new developer to seed files

areyeslo:
- Add Arturo Reyes Lopez as new developer in seed files.


dhaselhan (approved):
LGTM 👍

dhaselhan:
One small comment, for branches we use following naming convention:
name/feat-description-ticket 
For example, arturo/feat-add-myself-to-seed-915 if your ticket number was 915.
If its a fix you replace feat with fix.
For an MR like this not really a concern since it has no ticket but wanted to mention it.

PR #1007, opened 2024-10-03
LCFS-968: Bug fix in Allocation agreements for Fuel code and CI of fuel

areyeslo:
Fix bug in allocation agreements in the following scenarios:
- When changing the value of `Determining Carbon Intensity` field, `Fuel code` field value should be reset:
image

- Formatting `CI of fuel` field to show 2 decimal places in the UI. We keep storing the original value to avoid losing precision.
image


dhaselhan (approved):
LGTM 👍

PR #1010, opened 2024-10-04
fix: Reset line 6/8 if line 4 changes

dhaselhan:
* Change to new requirement resulting in cleaner logic
* Add lots of tests


AlexZorkin (approved):
Looks good 👍

PR #1011, opened 2024-10-04
fix: Auto-size text columns to keep nice grid size

dhaselhan:
* Add flex to longer columns so they grow to expand the space instead of the grey bar at the end
* Remove debugger


hamed-valiollahi:
Looks good to me! Thanks 👍

hamed-valiollahi (approved):

Issue #974, opened 2024-09-24
LCFS - TASK - Compliance report format


airinggov:
Following this mockup
https://preview.uxpin.com/da68049c0af15edd7ce4e47e4bee431fed40c30d#/pages/175233595

1. Move the Summary and declaration text outside the accordion and place directly above, the same format as the Report details heading above. The current Summary and declaration text should be changed to just Summary and the same format as other accordion headings.![Image](https://github.com/user-attachments/assets/7e99cbeb-47ac-4a2b-bc9c-e7b1bac9e4ee)

2. The Signing authority declaration text, checkbox and Submit report button should all be moved inside the accordion below the Summary tables. If there is a problem moving this into the accordion then it should be place directly under the Summary accordion above the Introduction accordion. I think the Save draft report button can be removed, the user input fields in the Summary tables should auto save. Also the declaration section is appearing on a new report page before even a draft is saved, it should not appear until the something is added to a schedule and the Summary accordion appears.

![Image](https://github.com/user-attachments/assets/43d66ddc-bf7a-4d63-a705-689c5d8e3782)

PR #1016, opened 2024-10-07
Refactor: LCFS - TASK - Compliance report format #974

prv-proton:
- #974
- Floating alert for AddEdit Fuel supplies. (After discussion can use across other components).


AlexZorkin (approved):
New Alert component looks great. I would suggest getting sign off from product owner and then we should implement it in the other schedules.

Issue #975, opened 2024-09-24
LCFS - BUG - Missing report type in index


airinggov:
**Describe the Bug**
The report type text is missing. Logged in as LCFS2 /LCFS Org 2

![Image](https://github.com/user-attachments/assets/80380b0b-bf67-4a27-9801-644446f94495)

**Expected Behaviour**
Should see the text Compliance report

PR #1017, opened 2024-10-07
Refactor: Relocate Validation Alert and Fix Missing Report Type - 973, 975

### LINKED ISSUE #973: LCFS - BUG - Validation alert, screen real estate
**Describe the Bug**
Can we reserve the screen real estate for the validation/error alert? Presently when the bar disappears the table bumps up. This would be on all schedules and forms with validation.

Relocate alert to directly under the table (above New row and Save button).

Image

**Expected Behaviour**

hamed-valiollahi:
This PR relocates the validation alert directly under the table to prevent UI shifting and fixes the missing compliance report type text in the index page.

Closes #973, #975

### DIFF
--- a/frontend/src/views/ComplianceReports/components/_schema.jsx
+++ b/frontend/src/views/ComplianceReports/components/_schema.jsx
@@ -22,7 +22,7 @@ export const reportsColDefs = (t, bceidRole) => [
     field: 'type',
     headerName: t('report:reportColLabels.type'),
     flex: 2,
-    valueGetter: () => 'Compliance report'
+    valueGetter: () => t('report:complianceReport')
   },
   {
     field: 'status',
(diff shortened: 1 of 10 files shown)


frontend/src/views/ComplianceReports/components/_schema.jsx
```diff
@@ -22,7 +22,7 @@ export const reportsColDefs = (t, bceidRole) => [
     field: 'type',
     headerName: t('report:reportColLabels.type'),
     flex: 2,
-    valueGetter: ({ data }) => data.type?.description || ''
+    valueGetter: () => 'Compliance report'
```
dhaselhan, line 25: suggestion: Update this to use the translation file

dhaselhan (changes requested):
Small change for translation. I believe we should be putting all UI strings there, but correct me if I'm wrong.

hamed-valiollahi:
> Small change for translation. I believe we should be putting all UI strings there, but correct me if I'm wrong.

I’ve moved all the UI strings to the locale file. Thanks for reviewing the code, Daniel! 👍

dhaselhan (approved):
LGTM 👍

PR #1021, opened 2024-10-07
fix: Change NotionalTransfers and OtherUses to use fitGridWidth

dhaselhan:
* Also move config from memo to direct set since its constant


kevin-hashimoto (approved):
Grid spans width, working as expected 👍🏻

PR #1023, opened 2024-10-08
Feat: Parenthesis for Negative Numbers - 934

AlexZorkin:
- Added an optional parameter useParentheses (default: false) to numberFormatter method
  - When useParentheses is true, negative numbers are wrapped in parentheses
  - When false (default), negative numbers use a minus sign prefix
- Summary Table now displays parenthesis on negative numbers
- Updated formatter tests


dhaselhan (approved):
LGTM 👍

PR #1024, opened 2024-10-08
Fix: Director Transfer Actions - 921

AlexZorkin:
- Updated what components are visible for Directors on submitted Transfers.


dhaselhan (approved):
LGTM 👍

PR #1025, opened 2024-10-08
feat: Update Initiative Agreement Tests

dhaselhan:
* Convert existing tests to view test
* Create more tests for repo and services
* Use enum in more places


AlexZorkin (approved):
Thanks for swapping out the raw text for the enum values. 👍

PR #1026, opened 2024-10-09
Feat: Update PyTests for Internal Comment - 942

### LINKED ISSUE #942: LCFS - Update PyTests for Internal Comment and Align Test Filenames
**Purpose:**  
To ensure the tests for the Internal Comment feature are up to date with the latest changes in the Internal Comment files and that each test file matches the corresponding file it is testing in the `internal_comment/` directory.

**Acceptance Criteria:**
- [ ] Review the current Internal Comment files to identify changes in methods and logic.
- [ ] Update the test files to reflect the correct methods and signatures from the latest code.
- [ ] Ensure mock data and assertions align with the current implementation.
- [ ] Verify all Internal Comment tests pass after updates.
- [ ] Rename test files to match the corresponding files in the `internal_comment/` directory.

**Additional Context:**
- The previous test files might include references to outdated methods, which should be updated to reflect the latest logic.
- Test logic should be rewritten to account for any changes in method arguments or return values.
- Ensure all test filenames align one-to-one with the corresponding files for better organization.

hamed-valiollahi:
This PR updates the PyTests for the internal comment feature.

Closes #942


backend/lcfs/conftest.py
```diff
@@ -143,7 +143,7 @@ async def dbsession_factory(
 def fastapi_app(
     dbsession: AsyncSession,
     fake_redis_pool: ConnectionPool,
-    set_mock_user_roles,  # Fixture for setting up mock authentication
```
dhaselhan, line 146: I see you renamed this from set_mock_user_roles to set_mock_user, however the function takes roles and not a user. Why do you think this was more clear?

backend/lcfs/tests/internal_comment/test_internal_comment.py
```diff
@@ -0,0 +1,461 @@
+import pytest
```
dhaselhan, line 1: I believe the goal of the ticket is to have one test file to one python file, but it looks like you combined them. Can you split them back up?

dhaselhan (changes requested):

backend/lcfs/tests/internal_comment/test_internal_comment.py
hamed-valiollahi: I had the same doubt. After discussing with Alex before starting this ticket, we agreed that since these endpoints are simple CRUD, we don’t need to write tests everything (like repos, services, and views).

backend/lcfs/conftest.py
hamed-valiollahi: In `test_internal_comment.py`, there are a few tests that require a specific Keycloak usernames. I needed to mock these tests with specific user details, for example, in line 367:

```python
set_mock_user(
         fastapi_app,
         ["GOVERNMENT"],
         user_details={
             "keycloak_username": "IDIRUSER"
         }
     )
```

To handle this, I improved `set_mock_user_roles` to support both setting roles and adding user details.

dhaselhan (approved):
LGTM 👍

PR #1029, opened 2024-10-09
feat: Add Prefix to Transaction Export

dhaselhan:
* Fix and update transaction tests
* Fix user role check to use correct values intsead of the name of the enum


backend/lcfs/db/models/transaction/Transaction.py
```diff
@@ -6,6 +6,18 @@
 
 from lcfs.db.base import BaseModel, Auditable, EffectiveDates
 
+transaction_type_to_id_prefix_map = {
```
AlexZorkin, line 9: I would like to move these into the /utils/constants.py file. Transactions are not directly related to entities on the backend so better to keep these references separate. This is one of those annoying moments where Transactions on the backend don't necessarily mean the same thing as on the frontend.

AlexZorkin (changes requested):
Looks great, let me know what you think of moving the mappings.

backend/lcfs/db/models/transaction/Transaction.py
dhaselhan: Done! Moved to constants.py. Wasn't sure if I should make it a class / uppercase it so let me know.

AlexZorkin (approved):
Mappings look great ty.

PR #1032, opened 2024-10-10
fix: Don't unload components when RequireAuth is loading

dhaselhan:
* Token refresh causes loading to go true, this clears all components from the page and resets their status
* By not loading, components on the page keep their state


AlexZorkin (approved):
Excellent fix tyty 👍

PR #1041, opened 2024-10-11
Remove duplicate from most tables

dhaselhan:
* Remove duplicate from everything other than FSE and Fuel Code tables


AlexZorkin (approved):
Looks good approved here. Just want to double check if there's any tests that need to be updated with removing the duplicate action?

dhaselhan:
Checked and it looks like no tests were impacted by removing duplicate

PR #1043, opened 2024-10-11
Remove Link from Status Renderer

dhaselhan:
* Remove link so it no longer has a hover cursor
* Fix permission checks in backend to use correct user_roles

### DIFF
--- a/frontend/src/utils/grid/cellRenderers.jsx
+++ b/frontend/src/utils/grid/cellRenderers.jsx
@@ -33,35 +33,30 @@ export const LinkRenderer = (props) => {
 export const StatusRenderer = (props) => {
   const location = useLocation()
   return (
-    <Link
-      to={props.node && location.pathname + '/' + props.node.id}
-      style={{ color: '#000' }}
+    <BCBox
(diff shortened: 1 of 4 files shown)


frontend/src/utils/grid/cellRenderers.jsx
```diff
@@ -33,35 +33,30 @@ export const LinkRenderer = (props) => {
 export const StatusRenderer = (props) => {
   const location = useLocation()
   return (
-    <Link
-      to={props.node && location.pathname + '/' + props.node.id}
-      style={{ color: '#000' }}
+    <BCBox
```
prv-proton, line 36: FYI Since status Renderer component is used in below table as well, removing the Link could cause inability to have the redirect link available here, i.e., it offers the users to right click and select "Open Link in New Tab/Window".
image

AlexZorkin (approved):
👍

frontend/src/utils/grid/cellRenderers.jsx
dhaselhan: Is this a requirement? Should I fix this?

PR #1050, opened 2024-10-15
fix: Refresh compliance balance when changing

dhaselhan:
* Add refresh call to re-load query balance
* Remove existing hacky approch
* Add ? to alertRef in case its not defined, this allowed onSettled to actually fire


frontend/src/constants/langEnUs.js
AlexZorkin, line 1: Are we not using this file?

AlexZorkin (approved):
👍

frontend/src/constants/langEnUs.js
dhaselhan: Not that my IDE can see, we normally use .json files in the locale folder so I'm not sure where this one came from

PR #1054, opened 2024-10-15
Small Test Fixes

dhaselhan:
* Fix mock user roles on initiative agreements
* Add an error message on director approval when roles do not match and only check for director role
* Disable auditing tests as auditing is not working, filed a ticket to fix

### DIFF
--- a/backend/lcfs/tests/conftest.py
+++ b/backend/lcfs/tests/conftest.py
@@ -1,4 +1,6 @@
 import pytest
+
+from lcfs.db.models.user.Role import RoleEnum
 from lcfs.web.api.base import PaginationRequestSchema, FilterModel, SortOrder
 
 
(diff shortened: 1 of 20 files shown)


backend/lcfs/tests/conftest.py
```diff
@@ -32,26 +34,23 @@ def pagination_request_schema():
...
-                self.last_name = "Doe"
-                self.keycloak_username = "johndoe"
-                self.organization_id = organization_id
-                self.email = email
-                self.role_names = [role]
-
-            def get_role_names(self):
-                return self.role_names
+def mock_user_profile(
+    role=RoleEnum.GOVERNMENT,
```
AlexZorkin, line 38: I think we have a few more places to update with the enum checks. If you run a find all on the all caps role names, see if you can find any others to fix.

AlexZorkin (approved):

backend/lcfs/tests/conftest.py
dhaselhan: Done! Looks like the were ~100 or so but I think I got them.

PR #1061, opened 2024-10-16
LCFS-985: Fixing filters in transaction and compliance report tables

areyeslo:
- Filters in transaction table and compliance report are working propertly backend. 
- Inputs are not case sensitive, user can type lower or capital letters
- Date picker is added
- Configuration in the front end to allow filters, including special configuration for Enum values
- Add clear button to reset individual filters.

### DIFF
--- a/frontend/src/components/BCDataGrid/BCDataGridServer.jsx
+++ b/frontend/src/components/BCDataGrid/BCDataGridServer.jsx
@@ -181,16 +181,26 @@ const BCDataGridServer = ({
     const filterModel = gridRef?.current?.api.getFilterModel()
     const filterArr = [
       ...Object.entries(filterModel).map(([field, value]) => {
+        // Check if the field is a date type and has an 'inRange' filter
+        if (value.filterType === 'date' && value.type === 'inRange') {
+          return {
+            field,
+            filterType: value.filterType,
+            type: value.type,
+            dateFrom: value.dateFrom,
+            dateTo: value.dateTo
+          }
+        }
         return { field, ...value }
       }),
       ...defaultFilterModel
     ]
     setFilterModel(filterArr)
     // save the filter state in browser cache.
-    localStorage.setItem(
-      `${gridKey}-filter`,
-      JSON.stringify(gridRef.current.api.getFilterModel())
-    )
+    // localStorage.setItem(
+    //   `${gridKey}-filter`,
+    //   JSON.stringify(gridRef.current.api.getFilterModel())
+    // )
   }, [])
 
   // Callback for grid sort changes.
--- a/frontend/src/components/BCDataGrid/BCDataGridServer.jsx
+++ b/frontend/src/components/BCDataGrid/BCDataGridServer.jsx
@@ -17,6 +17,7 @@ import BCAlert from '@/components/BCAlert'
 import BCBox from '@/components/BCBox'
 import DataGridLoading from '@/components/DataGridLoading'
 import { BCPagination } from './components'
+import { useTranslation } from 'react-i18next' // Import useTranslation
 // Register the required AG Grid modules for row model and CSV export functionality
 ModuleRegistry.registerModules([ClientSideRowModelModule, CsvExportModule])
 
@@ -74,6 +75,8 @@ const BCDataGridServer = ({
   suppressPagination,
   ...others
 }) => {
+  const { t } = useTranslation(['report'])
+
   // State declarations
   const [page, setPage] = useState(1)
   const [size, setSize] = useState(paginationPageSize)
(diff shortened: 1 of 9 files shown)


frontend/src/components/BCDataGrid/BCDataGridServer.jsx
```diff
@@ -74,6 +75,8 @@ const BCDataGridServer = ({
   suppressPagination,
   ...others
 }) => {
+  const { t } = useTranslation(['report'])
```
dhaselhan, line 78: I see translation imported here but I can't see it used anywhere, did I miss it?

frontend/src/views/Transactions/_schema.js
```diff
@@ -24,28 +24,48 @@ export const transactionsColDefs = (t) => [
...
+    },
     width: 222
   },
   {
     colId: 'fromOrganization',
     field: 'fromOrganization',
     headerName: t('txn:txnColLabels.organizationFrom'),
     minWidth: 300,
-    flex: 2
+    width: 300,
```
dhaselhan, line 55: Are you sure we want to use a fixed width? Can you make sure the table still grows / shrinks appropriately?

frontend/src/views/ComplianceReports/components/_schema.jsx
```diff
@@ -9,11 +9,11 @@ export const reportsColDefs = (t, bceidRole) => [
     field: 'compliancePeriod',
     headerName: t('report:reportColLabels.compliancePeriod'),
     width: 210,
-    cellRenderer: LinkRenderer,
-    cellRendererParams: {
-      url: ({ data }) => `${data.compliancePeriod?.description}/${data.complianceReportId}`,
-    },
-    valueGetter: ({ data }) => data.compliancePeriod?.description || ''
+    suppressFloatingFilterButton: true,
```
dhaselhan, line 12: Was it intentional to remove the cellRenderer: LinkRenderer here? Will that make the columns look different than before?

frontend/src/components/BCDataGrid/BCDataGridServer.jsx
areyeslo: Used to retrieve the constant value "complianceReport": "Compliance report":
```
localFilteredData = localFilteredData.filter((row) => {
        const typeLiteral = t('report:complianceReport').toLowerCase()
        return typeLiteral.includes(filterText)
      })
```

frontend/src/views/Transactions/_schema.js
areyeslo: Having flex property is not providing this elasticity. We might need to add the corresponding styling based on the view port and have the application more compatible with different devices.

frontend/src/views/ComplianceReports/components/_schema.jsx
areyeslo: You are making me wonder if I should get rid of the other cellRenderers in organization and status. What is the purpose of this if it is not to allow the user to click on the row and take the user to the details? This is done in ComplianceReports.jsx:
```
const handleRowClicked = useCallback(
    ({ data }) => {
      navigate(
        ROUTES.REPORTS_VIEW.replace(
          ':compliancePeriod',
          data.compliancePeriod.description
        ).replace(':complianceReportId', data.complianceReportId)
      )
    },
    [navigate]
  )
```

frontend/src/views/Transactions/_schema.js
```diff
@@ -88,7 +116,26 @@ export const transactionsColDefs = (t) => [
...
     headerName: t('txn:txnColLabels.updateDate'),
     valueFormatter: dateFormatter,
-    width: 190
+    width: 190,
+    filter: 'agDateColumnFilter',
+    filterParams: {
+      filterOptions: ['inRange', 'equals', 'lessThan', 'greaterThan'],
+      defaultOption: 'inRange',
+      comparator: (filterDate, cellValue) => {
+          const cellDate = new Date(cellValue).setHours(0, 0, 0, 0);
```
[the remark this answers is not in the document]
areyeslo: Not sure to understand, are you saying that transactions and compliance reports should handle in the same way `updatedDate`?

frontend/src/components/BCDataGrid/BCDataGridServer.jsx
dhaselhan: Ah thanks! Just could not see it and CTRL+F with t is not a great option.

frontend/src/views/ComplianceReports/components/_schema.jsx
dhaselhan: I'm not exactly sure to be honest, might be a better question for @AlexZorkin or maybe its safe to delete? I imagined it was to make it look like a blue link.

frontend/src/views/Transactions/_schema.js
dhaselhan: I'm asking as I recently added those flex attributes to allow the table to grow correctly on larger screens and was wondering if this broke that.

frontend/src/views/Transactions/_schema.js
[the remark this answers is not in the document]
dhaselhan: Why is one like this: new Date(cellValue).setHours and the other just filterDate.setHours. It looks like you are modifying the passed in filterDate, so I was thinking it would also make sense to have new Date(filterDate).setHours()

areyeslo:
Thanks a lot for the review @dhaselhan

PR #1080, opened 2024-10-18
feat: Move FuelSupply calculations to the back end

dhaselhan:
* Instead of using front end, calculate inside the update method
* Add logic to load the associated entities and perform calculations
* Change energy from float -> int as we never want it to have decimals
* Fix Precision Syntax and migrate Real columns -> Precision


AlexZorkin (approved):
Great cleanup 👍

PR #1081, opened 2024-10-18
Fix: Transfer Comment Inputs Validation

AlexZorkin:
- added validation to transfer comment fields for max length matching db values


dhaselhan (approved):
LGTM 👍

PR #1088, opened 2024-10-19
hotfix-LCFS-985-BugsFiltersTransactionComplianceTable

areyeslo:
Change in currencyFormatter affected to ComplianceReportSummary when passing a value in parameters. The issue was that the logic needs to take into consideration that transactions can have params.value with null value so this was displaying wrong information in the grid table.

This change is aimed to address the issue in ComplianceReportSummary while providing a solution for transactions.

### DIFF
--- a/frontend/src/utils/formatters.js
+++ b/frontend/src/utils/formatters.js
@@ -32,21 +32,15 @@ export const numberFormatter = (params, useParentheses = false) => {
 }
 
 export const currencyFormatter = (params) => {
-  if (params.value !== null && !isNaN(params.value)) {
-    return params.value.toLocaleString('en-CA', {
+  const cellValue = Object.hasOwn(params, 'value') ? params.value : params
+  if (cellValue !== null && !isNaN(cellValue)) {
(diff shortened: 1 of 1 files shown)


dhaselhan (changes requested):
I think this is a cleaner version that accomplishes the same goal:
```
export const currencyFormatter = (params) => {
  const cellValue = params.value ?? params;
  if (cellValue !== null && !isNaN(cellValue)) {
    return cellValue.toLocaleString('en-CA', {
      style: 'currency',
      currency: 'CAD'
    })
  }
  return cellValue
}
```

dhaselhan:
And after testing turns out that won't render null as empty, so I had to change to this:
```
export const currencyFormatter = (params) => {
  const cellValue = Object.hasOwn(params, 'value') ? params.value : params
  if (cellValue !== null && !isNaN(cellValue)) {
    return cellValue.toLocaleString('en-CA', {
      style: 'currency',
      currency: 'CAD'
    })
  }
  return cellValue
}
```

dhaselhan (approved):
LGTM 👍

Issue #1014, opened 2024-10-04
LCFS - Can access data for other suppliers and internal data


airinggov:
**Describe the Bug**
As a BCeID user viewing a compliance report or a transaction, it is possible to change the ID number in the URL and access reports/transactions from other suppliers. This presents a major security vulnerability, allowing unauthorized access to sensitive data.

I can also change URL id numbers in org pages and view any user including IDIR user info.

![Image](https://github.com/user-attachments/assets/ac4b29d0-ddbb-4c66-82c1-8f33952eaebf)

**Expected Behaviour**
A BCeID user should only be able to view compliance reports associated with their own account or permissions. Access to reports from other suppliers should be restricted and result in an error or access denied message.

Automated tests should be in place to ensure this never happens.

**Actual Behaviour**
By manually changing the ID number in the URL, a BCeID user can access compliance reports from other suppliers, which should be restricted to authorized users only.

This occurs with:
- admin adjustments
- initiative agreements
- compliance reports
- organization users (but not the transaction data)

It doesn't appear to happen with:
- credit transfers - although it throws an ugly Unexpected Application Error! that should be improved on.

- A blank screen within the standard header/footer would be acceptable for all instances (as and Unauthorized message may actually give the user some degree of info).

**Implications**
This issue poses a serious security risk, allowing users to view confidential reports from other suppliers. It violates data privacy regulations and could lead to significant data breaches.

**Steps To Reproduce**
User/Role: BCeID User
1. Log in with a BCeID account.
2. Navigate to one of your compliance reports.
3. In the URL, manually change the report ID to another supplier's report ID.
4. Observe that the other supplier's compliance report is accessible.

Grulin:
Hi @airinggov. This seems to be fixed, but in testing I noted that the error message says "You do not have permission to view this user's information"

![Image](https://github.com/user-attachments/assets/ef573407-c684-4b1e-9a75-19b0e0ef0877)

You noted above a requirement that the screen should not give an error message and just show as blank. Is that still a requirement you prefer?

airinggov:
@prv-proton @Grulin yes I think it would be better to just be blank, so give no error at all. An error message like this may provide information that we don't want to be giving out.

prv-proton:
Thanks @airinggov for consignation. Should this be across or just for user view?

airinggov:
Yes, I think for all unauthorized type error on the BCeID side. We don't display any error

Grulin:
[...]

PR #1089, opened 2024-10-21
Fix: Can access data for other suppliers and internal data

prv-proton:
- #1014 
- remove redundant codes
- add/update test cases for the changes.
- update error messages and components.


backend/lcfs/web/api/compliance_report/repo.py
```diff
@@ -142,7 +139,7 @@ async def get_compliance_period(self, period: str) -> CompliancePeriod:
         return result
 
     @repo_handler
-    async def get_compliance_report(
+    async def check_compliance_report(
```
dhaselhan, line 142: question: Why rename this to check_compliance_report when it still returns a compliance report?

backend/lcfs/web/api/initiative_agreement/validation.py
```diff
@@ -40,3 +43,27 @@ async def validate_initiative_agreement_update(
...
+    async def validate_organization_access(self, initiative_agreement_id: int):
+        initiative_agreement = await self.service.get_initiative_agreement(initiative_agreement_id)
+        if not initiative_agreement:
+            raise HTTPException(
+                status_code=status.HTTP_404_NOT_FOUND,
+                detail="Transaction not found.",
+            )
+
+        organization_id = initiative_agreement.to_organization.organization_id
+        user_organization_id = (
```
dhaselhan, line 56: suggestion: This block is repeated a few times, does it make sense to DRY this up?

dhaselhan (approved):
LGTM 👍

backend/lcfs/web/api/compliance_report/repo.py
prv-proton: unlike in get_compliance_report_id, where it joins and returns all related data, this method is only used to check the existence of the compliance report id in the main table. And since it's used in only validations, so had it renamed.

backend/lcfs/web/api/initiative_agreement/validation.py
prv-proton: yes, since the structure is different for each of the feature type, had to do this way. But will look if I can generalize this.

PR #1095, opened 2024-10-22
feat: Add Compare Table Test Page

dhaselhan:
* New compare table component with new page that allows comparing two compliance reports
* Clean up getRowID to return a string
* Always specify formatters instead of assuming None is Number
* Clean up formatters and add unit tests


frontend/src/utils/formatters.js
AlexZorkin, line 1: Nice commenting 👍

AlexZorkin (changes requested):
Look great, there is still a migration conflict on this branch that is resolved on release-0.2.0. Just merge release and then you're gtg. 👍

PR #1103, opened 2024-10-23
fix: fuel code seeder sequencing

AlexZorkin:
- removed explicit fuel_code ids so auto sequencing could auto set them


kevin-hashimoto (approved):
LGTM

PR #1104, opened 2024-10-23
Chore: hiding cypress from test pipeline for now

AlexZorkin:
- disabling cypress end to end tests in the dev test pipeline
- we can enable this for deployments to test environment and production


dhaselhan (approved):
LGTM 👍

PR #1116, opened 2024-10-25
Feat: Supplemental Reporting & Schedule Versioning

AlexZorkin:
- **New Feature: Supplemental Reporting**
  - Introduced the ability for users to create supplemental reports for compliance.
  - Supplemental reports are versioned to track changes across multiple report submissions.

- **New Feature: Versioned Record Handling**
  - Implemented versioning system for fuel_supply, fuel_export, and other report-related records.
  - When records are edited in a supplemental report, new versioned records are created with an `UPDATE` change type.
  - Original records retain their `CREATE` change type for historical tracking.
  - If a record is marked for deletion in a supplemental report, a new `DELETE` record is created instead of physically removing the data.
  - Supplemental versions maintain a clear distinction between the original and modified records for easy rollback or auditing.

- **Backend Enhancements**
  - Updated fuel_supply and fuel_export services to support version-controlled record creation, updates, and deletion handling.
  - Introduced action services for handling version-specific logic such as creating new versions for edits or managing delete actions within supplemental reports.

- **Frontend Integration**
  - Frontend now displays the latest version of records in the compliance report views, including edits from supplemental reports.

- **Database Changes**
  - New data model for Compliance Reports and Schedule tables to support versioning
  - Supplemental reports are stored with unique version identifiers and group uuids linked to their original compliance report.


backend/lcfs/web/api/compliance_report/schema.py
```diff
@@ -114,6 +123,9 @@ class FinalSupplyEquipmentSchema(BaseSchema):
 
 class ComplianceReportBaseSchema(BaseSchema):
     compliance_report_id: int
+    compliance_report_group_uuid: Optional[str]
+    version: Optional[int]
+    supplemental_initiator: Optional[SupplementalInitiatorType]
```
areyeslo, line 128: should be included `reporting_frequency`  as well?

backend/lcfs/web/api/compliance_report/services.py
```diff
@@ -39,20 +45,103 @@ async def create_compliance_report(
...
+        if user.organization_id != current_report.organization_id:
+            raise ServiceException(
+                "You do not have permission to create a supplemental report for this organization."
+            )
+
+        # Validate that the status of the current report is 'Assessed'
+        # if current_report.current_status.status != ComplianceReportStatusEnum.ASSESSED:
+        #     raise ServiceException(
+        #         "A supplemental report can only be created if the current report's status is 'Assessed'."
+        #     )
```
areyeslo, line 108: is this commented out because the status of the compliance report can be in any status to create a supplemental report?

backend/lcfs/web/api/compliance_report/repo.py
```diff
@@ -717,38 +667,31 @@ async def get_issued_compliance_units(
 
     @repo_handler
     async def calculate_fuel_quantities(
-        self, compliance_report_id: int, fossil_derived: bool
+        self, compliance_report: ComplianceReport, fossil_derived: bool
```
areyeslo, line 670: Great! you are passing compliance_report instead of just the id.

backend/lcfs/web/api/compliance_report/repo.py
```diff
@@ -158,37 +166,20 @@ async def check_compliance_report(
                 joinedload(ComplianceReport.organization),
                 joinedload(ComplianceReport.compliance_period),
                 joinedload(ComplianceReport.current_status),
+                joinedload(ComplianceReport.summary),
+                joinedload(ComplianceReport.fuel_supplies),
+                joinedload(ComplianceReport.other_uses),
+                joinedload(ComplianceReport.history).joinedload(
```
dhaselhan, line 172: Does this join history twice with its own different joins? Could we join all of them at once?

frontend/src/views/ComplianceReports/components/__tests__/ImportantInfoCard.test.jsx
```diff
@@ -0,0 +1,161 @@
...
+
+    // Mock useCreateSupplementalReport
+    useCreateSupplementalReport.mockReturnValue({
+      mutate: createSupplementalReportMock,
+      isLoading: false
+    })
+  })
+
+  // Custom render function to include ThemeProvider
+  const customRender = (ui) => {
```
dhaselhan, line 48: suggestion: Use the standard test wrapper instead of making a new custom render

dhaselhan (approved):
I like the # Step comments for the comples queries, LGTM 👍

backend/lcfs/web/api/compliance_report/services.py
AlexZorkin: good catch, this was just for testing. fixed.

backend/lcfs/web/api/compliance_report/schema.py
AlexZorkin: yes added ty.

backend/lcfs/web/api/compliance_report/repo.py
AlexZorkin: 👍

backend/lcfs/web/api/compliance_report/repo.py
AlexZorkin: The problem is that they're separate relationships so i don't know of a way to chain them from the one parent. I get errors like this when i tried it: "ORM mapped entity or attribute "ComplianceReportHistory.user_profile" does not link from relationship"

Let me know if there's a more optimized way to do this.

frontend/src/views/ComplianceReports/components/__tests__/ImportantInfoCard.test.jsx
AlexZorkin: thank you, updated.

PR #1120, opened 2024-10-25
fix: Transfers Fixes

dhaselhan:
* Fix date formatter not accepting null date, breaking the page while its loading
* Add an if check on transferId to prevent the InternalComment components from throwing an exception when it receives null
* Fix the transfer.json text to have the full correct message


AlexZorkin (approved):
👍

PR #1121, opened 2024-10-28
Fix: Error messages for permission access

prv-proton:
#1014
> @prv-proton @Grulin yes I think it would be better to just be blank, so give no error at all. An error message like this may provide information that we don't want to be giving out.

Fix for above, currently the error message is changed to Not found for all 403 error messages.

Need to confirm with Al, before the changes are approved.


AlexZorkin (approved):
Looks good on this end. Just confirm with product owner 👍

prv-proton:
Confirmed with @airinggov to go ahead with "Not found" messages in case of permission issues.

Issue #1005, opened 2024-10-02
LCFS - Compliance report, internal workflow, Recommended by manager status, IDIR Director role


airinggov:
Compliance report, internal workflow, Recommended by manager status, IDIR Director role

**Description:**  
The report will have two buttons to Issue assessment and Return to compliance manager. IDIR users in Analyst or Compliance manager roles will not see these buttons.

Once the Director has clicked the Issue assessment button the report is put into Assessed status, this transaction is recorded in the report history section above the compliance manger recommend info as per mock-up. Note the BCeID version displays different text.
IDIR version
•	Assessed [date time stamp] PST by [IDIR director first name, last name].

BCeID version (government staff names are not displayed on BCeID version, the Recommended history is also not displayed, see card #1006 )
•	Assessed [date time stamp] PST by the director under the Low Carbon Fuels Act.

The transaction ledger should be updated to capture the change in the suppliers compliance unit balance.

If the director clicks the return button the report will be put back into Recommended by analyst status for the compliance manger to action further. This action is not recorded in the report history. When the compliance manager recommends the report for the second time the original recommended history entry will be updated. This could be a different compliance manager so the name should also be updated if required.

**Wireframe:**  
https://preview.uxpin.com/e0e2f15fc1b52987fd0287ad8c8103452dd16d60#/pages/186169277

**Purpose and benefit to user:**  
The purpose of this workflow is to facilitate the review process for compliance reports by adding decision-making capabilities for the IDIR Director. The Director can either assess or return reports to a compliance manager, and the history of these actions will be appropriately recorded. This improves auditability and ensures that compliance reports are handled efficiently.

**Acceptance Criteria:**  
- [ ] Two buttons are available on the compliance report for IDIR Director role: Issue assessment and Return to compliance manager.
- [ ] The two buttons are only visible to IDIR Directors and not to users with Analyst or Compliance manager roles.
- [ ] Clicking the Issue assessment button changes the report status to "Assessed."
- [ ] A transaction is recorded in the report history section above the compliance manager recommendation, showing:  
  `Assessed <date time stamp> PST by <IDIR director first name, last name>.`
- [ ] Clicking the Return to compliance manager button changes the report status to "Recommended by analyst."
- [ ] No transaction is recorded in the report history for the return action.
- [ ] If the Compliance manager recommends the report again after it was returned, the original recommendation entry is updated, including the compliance manager’s name (if different).

**Development Checklist:**  
[...]

PR #1122, opened 2024-10-28
Fix: Bug fix for compliance report director assessment

prv-proton:
#1005
> Everything works great but the Issue assessment button gives a 500 error

Since the reports were already created in dev environment, without creating a transaction record in RESERVED status, existing reports were failing. Hence have made changes to create a new transaction as ADJUSTMENT for the final report assessment in case there is no existing transaction.

### DIFF
--- a/backend/lcfs/web/api/compliance_report/update_service.py
+++ b/backend/lcfs/web/api/compliance_report/update_service.py
@@ -219,6 +219,7 @@ class ComplianceReportUpdateService:
             raise HTTPException(status_code=403, detail="Forbidden.")
         # Update the transaction to assessed
         report.transaction.transaction_action = TransactionActionEnum.Adjustment
+        report.transaction.update_user = str(user.user_profile_id)
         await self.repo.update_compliance_report(report)
 
     async def handle_reassessed_status(self, report: ComplianceReport):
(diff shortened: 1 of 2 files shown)


AlexZorkin (changes requested):
Instead of adding this temporary code we should just add the records manually in the database. There's a chance we forget about this so I think its better to just fix the source of the issue.

prv-proton:
Thanks for the review @AlexZorkin, updated the data in dev and removed the temporary code.

AlexZorkin (approved):
👍

backend/lcfs/web/api/compliance_report/update_service.py
```diff
@@ -219,6 +219,7 @@ async def handle_assessed_status(self, report: ComplianceReport):
             raise HTTPException(status_code=403, detail="Forbidden.")
         # Update the transaction to assessed
         report.transaction.transaction_action = TransactionActionEnum.Adjustment
+        report.transaction.update_user = str(user.user_profile_id)
```
dhaselhan, line 222: update_user and create_user should be handled correctly but the auditing dependencies. It is also set to the users username, not their ID.

prv-proton: modified to set the username instead of user id

PR #1123, opened 2024-10-28
feat: Add More Fuel Codes

dhaselhan:
* Add 3 more fuel codes, so each fuel type (other than fossil-dreived) has at least one code


hamed-valiollahi (approved):
Looks good! 👍 Thank you!

PR #1128, opened 2024-10-28
fix: Change audit columns to use default and onUpdate

dhaselhan:
* Store user in a context variable
* Use context var for update / create user


backend/lcfs/db/base.py
```diff
@@ -17,6 +29,16 @@
 Base = declarative_base(metadata=metadata)
 
 
+# Define a context variable to store the user
+current_user_var = ContextVar("current_user", default=None)
+
+
+def get_current_user():
```
areyeslo, line 36: where this function is called?

prv-proton (approved):
👍

PR #1134, opened 2024-10-28
fix: Comment out failing tests

dhaselhan:
* Update tests to use vitest instead of mocha
* Update vitest version


AlexZorkin (approved):
Lots of great cleanup here, and thank you for finding that mocha conflict. 👍

PR #1139, opened 2024-10-29
Fix: Compliance report loading issue in dev

prv-proton:
Quick fix for the issue in dev.
getComplianceReport call is disabled if orgID is not provided, added fix to bypass this for IDIR users and async call introduced.


dhaselhan (approved):
LGTM 👍

PR #1159, opened 2024-10-31
fix: Change Energy to Big Int

dhaselhan:
* Integer is not large enough to hold enough digits.
* Increase worker count to 2


AlexZorkin (commented):
This migration change won't run on dev because it's already run. Are you planning to adjust it manually?

AlexZorkin (approved):
👍

Issue #1035, opened 2024-10-10
LCFS - Implement User Login History View for Administrators


AlexZorkin:
**Description:** 
Create a new feature in the administration tab that displays a table view of user login history. This view will be accessible only to administrators and will utilize the AG Grid read-only table to display data from the user_login_history table.

**Purpose and benefit to user:**
This feature will assist administrators in resolving login issues for users by providing a comprehensive view of login attempts, success rates, and error messages. It will improve the efficiency of troubleshooting and support processes.

**Acceptance Criteria:**
- [ ] Given I am an administrator, When I navigate to the administration tab, Then I see a "User Login History" option
- [ ] Given I am an administrator, When I click on "User Login History", Then I see a table view populated with data from the user_login_history table
- [ ] Given I am a non-administrator user, When I access the administration tab, Then I do not see or have access to the "User Login History" view
- [ ] Given I am an administrator viewing the login history, When I interact with the table, Then I can sort, filter, and search the data
- [ ] Given I am an administrator, When I view the login history table, Then I see columns for all relevant fields from the UserLoginHistory model

**Development Checklist:**
- [ ] Implement backend API endpoint to fetch user login history data
- [ ] Create frontend component for the User Login History view using AG Grid
- [ ] Implement role-based access control to restrict view to administrators only
- [ ] Integrate the new component into the administration tab
- [ ] Implement sorting, filtering, and search functionality for the table
- [ ] Implement Pagination with ag grid
- [ ] Add appropriate error handling and loading states
- [ ] Write unit and integration tests for both frontend and backend
- [ ] Update documentation to include information about the new feature

**Notes:**
- The UserLoginHistory model contains the following fields to be displayed in the table:
  - user_login_history_id
  - keycloak_email
  - external_username
  - keycloak_user_id
  - is_login_successful
  - login_error_message
  - create_date
- Ensure that sensitive information is properly handled and displayed according to security best practices

dhaselhan:
Idea:
This could be done as a Metabase report for less development effort

PR #1160, opened 2024-11-01
Feat: LCFS - Implement User Login History View for Administrators #1035

prv-proton:
- #1035
- pytest issue with multipart package, so suppressing the warning message.
- added deprecated mark on BCDataGrid components, using jsdoc package.
- In BCDataGridViewer added reset button to clear caching of filter and sort applied on the grid, in case of error.
- Enclosed User activity component inside Role to prevent loading for unintended users.
- Removed unused code and components.

### DIFF
--- a/frontend/src/components/BCDataGrid/BCDataGridClient.jsx
+++ b/frontend/src/components/BCDataGrid/BCDataGridClient.jsx
@@ -17,6 +17,9 @@ import DataGridLoading from '@/components/DataGridLoading'
 // Register the required feature modules with the Grid
 ModuleRegistry.registerModules([ClientSideRowModelModule])
 
+/**
+ * @deprecated
+ */
 const BCDataGridClient = ({
   gridOptions,
   gridKey,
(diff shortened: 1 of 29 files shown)


frontend/src/components/BCDataGrid/BCDataGridClient.jsx
```diff
@@ -17,6 +17,9 @@ import DataGridLoading from '@/components/DataGridLoading'
 // Register the required feature modules with the Grid
 ModuleRegistry.registerModules([ClientSideRowModelModule])
 
+/**
```
AlexZorkin, line 20: Do we want to remove these deprecated components?

AlexZorkin (approved):
This looks great. We will have to fix the user_login_history being created on every request but that can be handled in a follow up ticket. 👍

frontend/src/components/BCDataGrid/BCDataGridClient.jsx
prv-proton: yes, but there are components that are still using them.

PR #1162, opened 2024-11-02
LCFS-1112: Add FSE columns

areyeslo:
Add new fields to the FSE form:

- kWh usage: Optional number field with 2 decimal places, placed after the date of supply.
- Model: Optional text field, placed after "Make."
- Ports: Optional select field with options: "Single port" and "Dual port," placed after "Level of equipment."
- Intended users: Required multi-select field with options: "Multi-unit residential building," "Fleet," "Public," and "Employee," placed after "Intended use."

[Story](https://github.com/orgs/bcgov/projects/137/views/1?pane=issue&itemId=84734070&issue=bcgov%7Clcfs%7C1112)


dhaselhan (approved):
LGTM 👍

PR #1164, opened 2024-11-04
Fix: Add default uuid values for versioning fields

AlexZorkin:
- Added uuid values in the versioning migration for existing Compliance reports, fuel supply, and fuel export records.


dhaselhan (approved):
LGTM 👍

PR #1169, opened 2024-11-05
feat: Add RabbitMQ for Transactions

dhaselhan:
* Add BaseConsumer that hold standard connect and consume logic
* Add TransactionConsumer to consume messages from the transaction queue
* Add tests for setting up and tearing down the consumers


backend/lcfs/services/rabbitmq/transaction_consumer.py
```diff
@@ -0,0 +1,71 @@
...
+        compliance_units = message_content.get("compliance_units_amount")
+        org_id = message_content.get("organization_id")
+
+        redis = Redis.from_url(
+            str(settings.redis_url), encoding="utf8", decode_responses=True
+        )
+
+        async with AsyncSession(async_engine) as session:
+            async with session.begin():
+                repo = OrganizationsRepository(db=session)
```
AlexZorkin, line 57: Great to see that we can just use all our existing repos/services 👍

backend/lcfs/services/rabbitmq/base_consumer.py
AlexZorkin, line 1: This is a nice service. I'm wondering if we should keep this around even after TFRS is shut down. Might be some use cases for a message queue like this. Real time notifications just for fun?

backend/lcfs/scripts/send_transaction.py
AlexZorkin, line 1: plz add some basic documentation just explaining the purpose of this script and how to test with it.

AlexZorkin (approved):
Nice and simple base class approach thank you. Approved but please add a bit of documentation around the temporary purpose of the rabbitmq implementation.

backend/lcfs/scripts/send_transaction.py
dhaselhan: Done!

PR #1171, opened 2024-11-05
fix: Remove frontend logic for ciOfFuel

dhaselhan:
* Calculation is performed correctly on the backend, frontend just needs to show the field as it is.


hamed-valiollahi (approved):
Looks good to me! 👍

PR #1181, opened 2024-11-06
feat: Consolidate Compliance Report Cards

dhaselhan:
* Remove important info and file upload
* Remove org info
* Merge all into AssessmentCard


frontend/src/views/ComplianceReports/components/AssessmentCard.jsx
```diff
@@ -4,70 +4,245 @@ import { List, ListItem, ListItemText, Stack, Typography } from '@mui/material'
...
-      if (item.status.status === COMPLIANCE_REPORT_STATUSES.ASSESSED && !isGovernmentUser) {
-        item.status.status = "AssessedBy"
-      }
-      return item
-    }).filter(item => item.status.status !== COMPLIANCE_REPORT_STATUSES.DRAFT)
+    return [...history]
+      .sort((a, b) => {
+        return new Date(b.createDate) - new Date(a.createDate)
+      })
+      .map((item) => {
```
prv-proton, line 38: I noticed few things during my local testing. 

- Looks like the new history list expects "Compliance Report:" to be prefixed before every history record.
- When new supplemental report is submitted, assessment card does not show up.
- May need to implement the ability for the user to expand or collapse full history. (I'm not quite sure, but this might be altogether a new ticket).

prv-proton (changes requested):
Overall changes look good, but consider to make changes for the missed out features. Thanks

frontend/src/views/ComplianceReports/components/AssessmentCard.jsx
dhaselhan: Confirmed with Alex that updating history will be its own ticket. Good catch on the issue with supplementals, added a new flag and unit test to cover this case.

prv-proton (approved):
Looks good👍

PR #1184, opened 2024-11-07
fix: Remove non-JSON payload logic from decorators

dhaselhan:
* If the payload is a file, this breaks. As it tries to consume the file buffer that was already consumed.


backend/lcfs/web/core/decorators.py
```diff
@@ -54,11 +54,7 @@ async def get_request_payload(request):
         try:
             request_payload = await request.json()
```
dhaselhan, line 55: This is the actual change, rest is just formatting

hamed-valiollahi (approved):
Looks good to me! 👍

PR #1185, opened 2024-11-07
fix: Filter Report Status on Get Report Endpoint

dhaselhan:
* Re-use existing status mask from list reports on get compliance report
* This correctly masks the status for organizations

### DIFF
--- a/backend/lcfs/web/api/compliance_report/services.py
+++ b/backend/lcfs/web/api/compliance_report/services.py
@@ -201,27 +201,33 @@ class ComplianceReportServices:
 
     @service_handler
     async def get_compliance_report_by_id(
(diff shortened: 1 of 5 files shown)


backend/lcfs/web/api/compliance_report/services.py
```diff
@@ -201,27 +201,33 @@ def _mask_report_status(self, reports: List) -> List:
 
     @service_handler
     async def get_compliance_report_by_id(
-        self, report_id: int, bceid_user: bool = False
+        self, report_id: int, apply_masking: bool = False
```
AlexZorkin, line 204: its always nice to have context related variable names 👍

AlexZorkin (approved):
Looks good 👍

PR #1201, opened 2024-11-08
Update Auth Tests

dhaselhan:
* Able to get some working but cannot figure out mocking AsyncContextManager


AlexZorkin (approved):
Who needs async context managers anyway. 👍

PR #1205, opened 2024-11-09
fix: New AutocompleteCellEditor

dhaselhan:
* Delete old unused
* Rename existing to cell editor
* Update component to use a forwardRef and properly call stop editing and onValueChange
* Bind onBlur to the internal input when freeSolo is true to fix the bug


AlexZorkin (commented):
This is looking good, but there's a minor bug on get compliance report. Not sure if this is fixed in a separate branch but i got this error on:
    return await report_service.get_compliance_report_by_id(report_id, bceid_user=True)

bceid_user should now apply_masking, just a minor fix

dhaselhan:
> This is looking good, but there's a minor bug on get compliance report. Not sure if this is fixed in a separate branch but i got this error on: return await report_service.get_compliance_report_by_id(report_id, bceid_user=True)
> 
> bceid_user should now apply_masking, just a minor fix

Rebased to get the fix for this in. Kevin also tested it and found some issues that needed to be fixed so I marked it Draft while I fix them.

kevin-hashimoto (approved):
Nice! Autocomplete cells now trigger onblur when clicking into another cell after an edit!
LGTM 👍🏻

PR #1212, opened 2024-11-12
fix: Fix some tests

dhaselhan:
* Fix various tests
* Small cleanup on shadowed var and orgId not being camel case


AlexZorkin (approved):
Looks good 👍

PR #1214, opened 2024-11-13
Feat: Updated styling for compliance report views - 1140

AlexZorkin:
- Fix summary table column widths to match each other closer at max width
- added disable logic to signing authority checkbox with tooltip
- move signing authority button to flex left
- updated floating selection boxes to automatically fit to column width


areyeslo (approved):
LGTM. Looks great the styling!

PR #1245, opened 2024-11-15
LCFS-1197: BUG IDIR users can see compliance reports in "draft' status

areyeslo:
- IDIR users are not allowed to see `Draft` compliance reports
- BCEID users are not allowed to se internal status.

[Story](https://github.com/orgs/bcgov/projects/137/views/1?pane=issue&itemId=86543645&issue=bcgov%7Clcfs%7C1197)

### DIFF
--- a/backend/lcfs/web/api/compliance_report/views.py
+++ b/backend/lcfs/web/api/compliance_report/views.py
@@ -9,13 +9,7 @@ GET: /reports/<report_id> - retrieve the compliance report by ID
 import structlog
 from typing import List
 
-from fastapi import (
-    APIRouter,
-    Body,
-    status,
-    Request,
-    Depends,
-)
+from fastapi import APIRouter, Body, status, Request, Depends, HTTPException
 
 from lcfs.db.models.user.Role import RoleEnum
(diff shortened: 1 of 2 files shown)


backend/lcfs/web/api/compliance_report/views.py
```diff
@@ -70,7 +66,13 @@ async def get_compliance_reports(
     pagination.filters.append(
         FilterModel(field="status", filter="Draft", filter_type="text", type="notEqual")
     )
-    return await service.get_compliance_reports_paginated(pagination)
+
+    try:
```
dhaselhan, line 70: The @view_handler should be mapping the DataNotFoundException to a 404 already.
```
            except DataNotFoundException:
                raise HTTPException(status_code=404, detail="Not Found")
```
Can this be removed?

dhaselhan (changes requested):
Remove redundant code.

backend/lcfs/web/api/compliance_report/views.py
areyeslo: Thank you for catching this!

dhaselhan (approved):
LGTM 👍

PR #1248, opened 2024-11-18
feat: Add can_sign to compliance summary

dhaselhan:
* Add new UI for SigningAuthorityDeclaration to have a disabled state with a warning message
* Calculate can_sign based on number of records in each schedule and use to set flag
* Update submit logic to check the same flag and raise an error
* Update compliance summary calculation to only call get effective records one
* Update / Add tests accordingly

### DIFF
--- a/backend/lcfs/web/api/compliance_report/repo.py
+++ b/backend/lcfs/web/api/compliance_report/repo.py
@@ -666,32 +666,27 @@ class ComplianceReportRepository:
         )
         return result or 0
 
-    @repo_handler
-    async def calculate_fuel_quantities(
-        self,
-        compliance_report_id: int,
-        effective_fuel_supplies: List[FuelSupply],
-        fossil_derived: bool,
+    def aggregate_fuel_supplies(
+        self, fuel_supplies: List[FuelSupply], fossil_derived: bool
     ) -> Dict[str, float]:
-        """
-        Calculate the total quantities of fuels, separated by fuel category and fossil_derived flag.
-        """
+        """Aggregate quantities from fuel supplies based on fossil_derived flag."""
         fuel_quantities = defaultdict(float)
 
-        # Filter fuel supplies based on fossil_derived flag
-        filtered_fuel_supplies = [
-            fs
-            for fs in effective_fuel_supplies
-            if fs.fuel_type.fossil_derived == fossil_derived
-        ]
-
-        # Aggregate quantities from fuel supplies
-        for fs in filtered_fuel_supplies:
-            fuel_category = fs.fuel_category.category.lower().replace(" ", "_")
+        # Use a list comprehension to filter and iterate in one step
+        for fs in (
+            fs for fs in fuel_supplies if fs.fuel_type.fossil_derived == fossil_derived
+        ):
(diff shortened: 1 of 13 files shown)


backend/lcfs/web/api/compliance_report/repo.py
```diff
@@ -667,18 +668,16 @@ async def get_issued_compliance_units(
 
     @repo_handler
     async def calculate_fuel_quantities(
```
prv-proton, line 670: Just a suggestion, is it possible to separate the business logic into service component and use repo module only for data extraction?

backend/lcfs/web/api/compliance_report/summary_service.py
```diff
@@ -255,6 +262,7 @@ async def calculate_compliance_report_summary(
         # After the report has been submitted, the summary becomes locked
         # so we can return the existing summary rather than re-calculating
         if summary_model.is_locked:
+            print("LOCKED")
```
prv-proton, line 265: can we have a log message instead.

backend/lcfs/web/api/compliance_report/update_service.py
```diff
@@ -213,9 +220,11 @@ async def handle_assessed_status(self, report: ComplianceReport):
...
         if not has_director_role:
             raise HTTPException(status_code=403, detail="Forbidden.")
-        # Update the transaction to assessed
-        report.transaction.transaction_action = TransactionActionEnum.Adjustment
-        report.transaction.update_user = user.keycloak_username
+
+        if report.transaction:
+            # Update the transaction to assessed
+            report.transaction.transaction_action = TransactionActionEnum.Adjustment
+            report.transaction.update_user = user.keycloak_username
```
prv-proton, line 227: it would be good idea to raise Value error here.
else:
    # Handle the case where the transaction is missing
    raise ValueError(
        f"Transaction is missing for the compliance report with ID: {report.compliance_report_id}."
    )

prv-proton (changes requested):
Thank you for the effort and detail in this implementation! I've provided few suggestions. If any suggestion doesn't align with the current objectives, feel free to disregard or discuss further.
Tested and everything is working as expected. 👍

backend/lcfs/web/api/compliance_report/update_service.py
dhaselhan: For 0 credit compliance reports we no longer create a transaction, I talked to Alex and we agreed that made more sense than making a 0 credit adjustment, so not all reports will have transactions.

backend/lcfs/web/api/compliance_report/summary_service.py
dhaselhan: This is leftover debugging, will remove

dhaselhan: Removed

backend/lcfs/web/api/compliance_report/repo.py
dhaselhan: Broke this into multiple functions. The calculate is now in the service and it uses 3 query methods from the repo for the logic.

prv-proton (approved):
Looks good

PR #1251, opened 2024-11-18
fix: Always enable lines 7 and 9 in 2024

dhaselhan:
* Remove all other logic for controlling the Diesel and Gasoline controls. Aim to remove this code and have it auto-load for 2025 and onwards.


AlexZorkin (approved):
👍

PR #1257, opened 2024-11-19
Feat: Fuel Refactor & Compliance Units Validation - 1176 1198 1217

AlexZorkin:
- Refactored fuel calculations for supply and export to use a common lookup for calculation values
- Added tests to cover compliance unit cases


dhaselhan (approved):
Great cleanup and comments on the calculations, much easier to follow.

LGTM 👍

Issue #1195, opened 2024-11-08
LCFS - FSE Accessibility issue


StuartGALL:
**Title:** Improve Accessibility of Date Range Picker and Cell Filters

**Description:** 
Reformat the "date range picker" and "cell filters" components to improve accessibility. Ensure that both elements are fully navigable by keyboard and compatible with screen readers.

Please refer to the provided screenshots for current UI layout.

**Wireframe Link:**  
N/A

**Purpose and Benefit to the User:**  
Enhancing the accessibility of these components will ensure that users who rely on keyboard navigation or screen readers can effectively interact with and use the date range picker and cell filters. This change will make the UI more inclusive and compliant with accessibility standards.

**Acceptance Criteria:**  
- [ ] Date range picker can be navigated entirely via keyboard.
- [ ] Date range picker is compatible with screen readers, providing relevant labels and focus states.
- [ ] Cell filters are accessible via keyboard and screen readers, with each filter option labeled appropriately.

**Development Checklist:**  
- [ ] Refactor date range picker for full keyboard navigability.
- [ ] Add ARIA labels and roles to date range picker for screen reader support.
- [ ] Refactor cell filters for keyboard and screen reader compatibility.
- [ ] Conduct an accessibility audit and fix any remaining issues in these areas.

**Notes:**  
Please refer to the screenshots provided to understand the current layout of the date range picker and cell filters.

![Image](https://github.com/user-attachments/assets/7223c46c-ec9b-41f1-813d-d6af117d6434)

Issue #1221, opened 2024-11-13
LCFS - BUG - FSE Date range


airinggov:
**Describe the Bug**  
The FSE date range lacks validation to restrict the selected date range within the calendar year of the report. For instance, the date range for a 2024 report should only allow dates from January 1, 2024, to December 31, 2024.

**Expected Behaviour**  
The FSE date range should have validation that restricts the date selection to only fall within the specific calendar year of the report.

**Actual Behaviour**  
Currently, the date range validation allows selections that extend beyond the calendar year of the report, which could lead to data misalignment or inaccurate reporting.

**Implications**  
Without proper date range validation, reports may include date ranges that are outside the intended reporting period, potentially leading to compliance issues or inaccurate data representation.

**Steps To Reproduce**  
User/Role: Report creator or analyst
1. Open the FSE date range selection for a new report (e.g., 2024 report).
2. Attempt to select a date range that starts before January 1, 2024, or ends after December 31, 2024.
3. Observe that there is no restriction or error message preventing this selection.

Issue #1188, opened 2024-11-07
LCFS - Set focus


airinggov:
**Description:**  
For all AG Grid input forms that include a duplicate row icon or "Add row" button, automatically set the focus to the first input field in the newly created row after clicking either action. This includes all compliance report schedules including FSE on and the Fuel codes form.

**Wireframe:**  
No wireframe is necessary as this is a functionality improvement to existing forms.

**Purpose and benefit to user:**  
This feature streamlines data entry by ensuring the user can immediately begin typing in the new row without additional clicks, enhancing the usability and efficiency of the form.

**Acceptance Criteria:**
- [ ] When a new row is created via the duplicate row icon, focus is set to the first input field in the new row.
- [ ] When a new row is created via the "Add row" button, focus is set to the first input field in the new row.
- [ ] The focus action should work consistently across all AG Grid input forms where either button is available.
- [ ] No additional clicks should be required from the user to start typing in the new row.

**Development Checklist:**
- [ ] Identify all AG Grid input forms with the duplicate row icon or "Add row" button.
- [ ] Implement focus setting functionality for the first input field in a new row upon the click event of either button.
- [ ] Test the focus behavior on all applicable forms to ensure consistency.
- [ ] Verify no regressions in form functionality or unexpected interactions.
- [ ] Document the functionality in user documentation or release notes, if applicable.

**Notes:**  
This enhancement addresses a user experience improvement, ensuring a seamless transition to new rows for data entry.

PR #1258, opened 2024-11-19
Fix: FSE date range & accessibility issues, focus issue in ag-grid.

prv-proton:
Fixes: #1195, #1221,  #1188


AlexZorkin (changes requested):
Bug: When inputting manually into the date picker, no text is visible when typing. I don't think the value is being set.

Testing ticket requirements looks good otherwise.
- Tabbing now works into the date field.
- Validation of current year is working.
- Add new row automatically selects first input

AlexZorkin (approved):
Looks good, thanks for adressing the bug. 👍

PR #1260, opened 2024-11-19
Add number formatter to Nameplace Capacity

dhaselhan:


prv-proton (approved):
👍

prv-proton (changes requested):
commas won't gets added while editing, get formatted after editing is complete.

dhaselhan:
> Hi @dhaselhan , thanks for the implementation, Please find my observations below.
> 
> * commas won't get added while editing, instead it gets formatted after editing is complete.
> 
> image image
> * Using NumberEditor component should fix this issue. Also if possible please fix the alignment to be same while editing and not editing.

Good catch Prashanth! Updated the code to use the NumberEditor and it works great. I did look at other places for column alignment and it looks like Fuel Quantity follows the same left/right behavior and I couldn't find any cells that work the way you are suggesting.

prv-proton:
> > Hi @dhaselhan , thanks for the implementation, Please find my observations below.
> > 
> > * commas won't get added while editing, instead it gets formatted after editing is complete.
> > 
> > image image
> > 
> > * Using NumberEditor component should fix this issue. Also if possible please fix the alignment to be same while editing and not editing.
> 
> Good catch Prashanth! Updated the code to use the NumberEditor and it works great. I did look at other places for column alignment and it looks like Fuel Quantity follows the same left/right behavior and I couldn't find any cells that work the way you are suggesting.

I had a ticket similar to this changes in FuelCode carbonIntensity, and Al suggested this change. you can refer here
frontend/src/views/FuelCodes/AddFuelCode/_schema.jsx
`type: 'leftAligned'`. We can confirm this with Al and Stuart tomorrow if these needs to be consistent across. Thanks for the changes, everything looks good.

prv-proton (approved):

PR #1272, opened 2024-11-21
Fix: nifi wf fix for transfers data

prv-proton:
- fix contention with mv_aggregate_transaction (materialized views) during load.
- avoid inserting duplicate entries into history table.
- update the version control flow files.


AlexZorkin (approved):
Thank you for digging into this now. I will create a followup ticket to handle the other transaction types. 👍

PR #1275, opened 2024-11-21
Fix: User Activity Table - 1264

AlexZorkin:
- Refactored user activity table to use new BCGridViewer
- Updated tests


dhaselhan (approved):
LGTM 👍

PR #1277, opened 2024-11-22
feat: Improved User Tracking Logic

dhaselhan:
* Create keycloak wrapper to listen to keycloak events
* Use wrapper to listen to authenticated and send a post to the backend
* Update backend with a track login endpoint
* Use session storage so that page refreshes do not count as logins


frontend/src/components/KeycloakProvider.jsx
AlexZorkin, line 1: Nice wrapper 👍

backend/lcfs/services/keycloak/authentication.py
```diff
@@ -136,21 +135,12 @@ async def authenticate(self, request):
 
                     if user is None:
                         # Handle the case where no user is found
-                        await self.create_login_history(
```
AlexZorkin, line 139: I think we should retain the history creation for failed login attempts here. This is one of the main use cases is so that business area admins can debug supplier login issues by seeing the failed attempt information. Moving to just front end we only track successful logins. Let me know what you think on this.

AlexZorkin (changes requested):

backend/lcfs/services/keycloak/authentication.py
dhaselhan: Added back in error logging for user not found and username mismatch, and tested by changing some fields in the DB by logging back in. Unfortunately authentication.py exists outside of FastAPI, so I was unable to use depends and had to duplicate the insert logic.

AlexZorkin (approved):
👍

PR #1278, opened 2024-11-22
Fix: FS duplicate bug #1216

### LINKED ISSUE #1216: LCFS - BUG - Duplicate row validation with End use
**1. Describe the Bug**
This should not be throwing an error
Image

**Expected Behaviour**
The End use should also be part of the duplicate row validation, having a different End use makes this a unique fuel.

**Actual Behaviour**
It doesn't appear End use is part of the duplicate row validation.

**Implications**
This prevents suppliers from reporting fuels accurately.

**************
**2. Spike**
**Problem Description**
After deleting a duplicate row, trying to input the same information, it is as if the error matching hasn't been cleared and a duplicate row error is given when it shouldn't be

Image

**In order to**
Enable suppliers to report fuel data accurately without encountering erroneous duplicate errors, the system **needs to** validate rows.

**Solution Needs**
- Clear any residual matching errors when entries are deleted or modified to avoid incorrect duplicate error messages.

**Timebox**
Allocate up to 3 points to investigate and prototype a validation update for duplicate entries.

**Outcome**
- Identify whether additional logic or database adjustments are needed to prevent residual error matching.

**Additional Context**
- Current implications affect fuel reporting accuracy, impacting compliance and record-keeping for suppliers.
- Steps to reproduce the current issue may include:
  - Entering a fuel record with specific values.
  - Deleting and re-entering similar records with different values.

kevin-hashimoto:
- fix: a bug that prevents new rows from saving if it is the same as a previously deleted row
- feat: end use is now a unique identifier for the row
closes #1216 

jsx file just a cleanup, no meaningful changes

### DIFF
--- a/backend/lcfs/web/api/fuel_supply/repo.py
+++ b/backend/lcfs/web/api/fuel_supply/repo.py
@@ -298,6 +298,16 @@ class FuelSupplyRepository:
     @repo_handler
     async def check_duplicate(self, fuel_supply: FuelSupplyCreateUpdateSchema):
         """Check if this would duplicate an existing row"""
+
+        delete_group_subquery = (
+            select(FuelSupply.group_uuid)
+            .where(
+                FuelSupply.compliance_report_id == fuel_supply.compliance_report_id,
+                FuelSupply.action_type == ActionTypeEnum.DELETE,
+            )
+            .distinct()
+        )
+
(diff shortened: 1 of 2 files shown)


dhaselhan (approved):
LGTM 👍

PR #1285, opened 2024-11-22
Feat: Other Diesel - 1243

AlexZorkin:
- added new fuel type to seeders called 'Other diesel'
- updated frontend schema logic on dependant columns


frontend/src/views/Transactions/components/__tests__/Comments.test.jsx
```diff
@@ -70,7 +70,7 @@ describe('Comments Component', () => {
     fireEvent.click(toggleButton)
     setTimeout(() => {
```
dhaselhan, line 71: Rather than adjusting these timeouts I would use:   ```await waitFor(() => {}```

frontend/src/views/FuelSupplies/_schema.jsx
```diff
@@ -416,12 +419,13 @@ export const fuelSupplyColDefs = (optionsData, errors, warnings) => [
       params.value ||
       (!params.value && <Typography variant="body4">Select</Typography>),
     suppressKeyboardEvent,
-    editable: (params) => /other/i.test(params.data.fuelType),
+    editable: (params) => params.data.fuelType === FUEL_TYPE_OTHER,
```
dhaselhan, line 422: Hard-coding string constants across multiple frontend places and the backend creates fragile code. My idea would be to add a Boolean to fuel_type table that controls this field instead. Thoughts?

frontend/src/views/FuelSupplies/_schema.jsx
```diff
@@ -175,17 +177,18 @@ export const fuelSupplyColDefs = (optionsData, errors, warnings) => [
     }),
     cellStyle: (params) => {
       const style = StandardCellWarningAndErrors(params, errors, warnings)
-      const conditionalStyle = /other/i.test(params.data.fuelType)
-        ? { backgroundColor: '#fff', borderColor: 'unset' }
-        : { backgroundColor: '#f2f2f2' }
+      const conditionalStyle =
+        params.data.fuelType === FUEL_TYPE_OTHER
```
dhaselhan, line 181: suggestion: DRY this up by wrapping the StandardCellWarningAndErrors into another CellStyle that takes a function to evaluate whether the field is editable or not. This will ensure we have consistent editable / non-editable styles across the product.

dhaselhan (changes requested):
Lets DRY this up!

frontend/src/views/Transactions/components/__tests__/Comments.test.jsx
AlexZorkin: fixed ty

frontend/src/views/FuelSupplies/_schema.jsx
AlexZorkin: I have added your suggestion, please review.

frontend/src/views/FuelSupplies/_schema.jsx
AlexZorkin: I’ve refactored the code so the hard-coded string is only checked in one place within the file. I’m not sure adding a boolean field to the fuel_type table is warranted since it would only apply to the 'Other' record.

frontend/src/views/FuelSupplies/_schema.jsx
AlexZorkin: realized I need to add this to the other schemas

AlexZorkin: now so dry.

dhaselhan (approved):
Nice clean up! The DRYest code ever.
LGTM 👍

PR #1292, opened 2024-11-26
Fix: for Fuel code prefix and asterisk indicator issues

prv-proton:
- Add Fuel code is currently giving error in dev env for PROXY prefix, hence adding fix for it.
- current logic to identify the asterisk indicator is not working in dev, hence adding extra condition.


backend/lcfs/web/api/fuel_code/repo.py
```diff
@@ -548,7 +548,7 @@ async def get_next_available_fuel_code_by_prefix(self, prefix: str) -> str:
                 WHERE fcp.prefix = :prefix
             ),
             all_possible_codes AS (
-                SELECT generate_series(101, COALESCE((SELECT MAX(base_code) FROM parsed_codes), 101)) AS base_code
```
dhaselhan, line 551: Your description doesn't explain how this fixes the issue, why are we adding +1?

prv-proton: Sorry for the confusion , this was the issue I found.
In database we already had 101, 102 prefixes, and since as per the sequence generator it only generates sequence starting from 101 to max prefix value which is 102. Since both sequences were already taken query was giving null as result. Adding +1 to max value would give 103 as the next sequence instead of getting null result. Hope this clears if not happy to connect and clear doubts. Thank you

areyeslo (approved):
Works like a charm! LGMT

backend/lcfs/web/api/fuel_code/repo.py
dhaselhan: Thanks for the explanation! Trying to learn as much as I can :)

PR #1293, opened 2024-11-26
feat: retrieve compliance report chain

kevin-hashimoto:
- feat: Retrieve Full Report Chain and Enhance Frontend Display

### DIFF
--- a/backend/lcfs/web/api/compliance_report/views.py
+++ b/backend/lcfs/web/api/compliance_report/views.py
@@ -19,7 +19,9 @@ from lcfs.web.api.compliance_report.schema import (
     ComplianceReportBaseSchema,
     ComplianceReportListSchema,
     ComplianceReportSummarySchema,
-    ComplianceReportUpdateSchema, ComplianceReportSummaryUpdateSchema,
+    ChainedComplianceReportSchema,
+    ComplianceReportUpdateSchema,
+    ComplianceReportSummaryUpdateSchema,
 )
 from lcfs.web.api.compliance_report.services import ComplianceReportServices
 from lcfs.web.api.compliance_report.summary_service import (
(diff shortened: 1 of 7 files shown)


prv-proton (changes requested):
can you please fix, failing test cases.
https://github.com/bcgov/lcfs/runs/33715871791

frontend/src/views/ComplianceReports/EditViewComplianceReport.jsx
```diff
@@ -92,9 +92,9 @@ export const EditViewComplianceReport = () => {
     complianceReportId
   )
 
-  const currentStatus = reportData?.data?.currentStatus?.status
+  const currentStatus = reportData?.report.currentStatus?.status
```
prv-proton, line 95: reportData?.report seems to be unnecessary, can you please recheck this as it is failing
image

backend/lcfs/web/api/compliance_report/views.py
```diff
@@ -80,12 +82,16 @@ async def get_compliance_report_by_id(
...
     service: ComplianceReportServices = Depends(),
     validate: ComplianceReportValidation = Depends(),
-) -> ComplianceReportBaseSchema:
+) -> ChainedComplianceReportSchema:
     await validate.validate_organization_access(report_id)
 
     mask_statuses = not user_has_roles(request.user, [RoleEnum.GOVERNMENT])
 
-    return await service.get_compliance_report_by_id(report_id, mask_statuses)
+    result = await service.get_compliance_report_by_id(
```
prv-proton, line 90: `service.get_compliance_report_by_id` this function is being called from backend/lcfs/web/api/organization/views.py as well, may need to make changes in it as well.

prv-proton (changes requested):
Thanks for fixing test cases @kevin-hashimoto, But found issues w.r.t BCeID user.
- Please fix the issue from both BCeID user and IDIR user perspective and mask statuses as deemed necessary. 
- Since we're using chain instead of history, is it necessary to have both history and chain data?

kevin-hashimoto:
> Thanks for fixing test cases @kevin-hashimoto, But found issues w.r.t BCeID user.
> 
> * Please fix the issue from both BCeID user and IDIR user perspective and mask statuses as deemed necessary.
> * Since we're using chain instead of history, is it necessary to have both history and chain data?

We will need both as the chain is a list of supplemental reports with the history of events for each report

PR #1299, opened 2024-11-26
Upgrade to v4 artifacts

dhaselhan:
* Main changes involve multiple uploads to the same artifact, which luckily we don't appear to be using


AlexZorkin (approved):
Lets see how if works 👍

dhaselhan:
image
Looks like its working :)

PR #1306, opened 2024-11-28
Chore: Squash migrations and update seeders

AlexZorkin:
- squashed latest migrations
- separate migration for db functions
- updated seeders


prv-proton (approved):
reviewed and tested, everything looks good. Thank you

PR #1308, opened 2024-11-28
feat: Add UCI to Fuel Supply and Fuel Export

dhaselhan:


backend/lcfs/web/api/fuel_code/repo.py
```diff
@@ -798,12 +792,29 @@ async def get_standardized_fuel_data(
...
+            uci=uci.intensity if uci else None,
+        )
+
+    @repo_handler
+    async def get_additional_carbon_intensity(
+        self, fuel_type_id, end_use_type_id
+    ) -> Optional[AdditionalCarbonIntensity]:
+        """Get a single use of a carbon intensity (UCI), returns None if one does not apply"""
+        query = select(AdditionalCarbonIntensity).where(
+            AdditionalCarbonIntensity.end_use_type_id == end_use_type_id,
```
areyeslo, line 815: End use is always "Any" in endpoint /compliance-reporting/:compliancePeriod/:complianceReportId/supply-of-fuel . It is not possible to check how the uci is changing when the user selects different end use. Based on the code should work, but it would be nice to check in UI.

backend/lcfs/db/migrations/versions/2024-11-28-01-09_b4da565bb711.py
```diff
@@ -0,0 +1,48 @@
...
+"""
+
+import sqlalchemy as sa
+from alembic import op
+from alembic_postgresql_enum import TableReference
+from sqlalchemy.dialects import postgresql
+
+# revision identifiers, used by Alembic.
+revision = "b4da565bb711"
+down_revision = "0775a141d335"
```
areyeslo, line 16: Is this correct down_revision?

backend/lcfs/web/api/fuel_code/repo.py
dhaselhan: You are correct that currently there are no possible fuel combinations that use UCI. I modified by own fuel codes to properly test.
image

backend/lcfs/db/migrations/versions/2024-11-28-01-09_b4da565bb711.py
dhaselhan: Good catch. Not at all, I made this MR before the squash. Updated to have the correct revision.

areyeslo (approved):
LGTM

PR #1312, opened 2024-11-28
LCFS-1280: Fuel supply validation (2/2)

areyeslo:
- The default **Quantity supplied** value being 0 should not be allow. 
- Zero should not be a valid amount in a fuel supply record and the row should not save until the user has entered a value.

[Story](https://github.com/orgs/bcgov/projects/137/views/1?pane=issue&itemId=88333582&issue=bcgov%7Clcfs%7C1280)


prv-proton (changes requested):
Thanks for the implementation @areyeslo, but I think this ticket is focused on "Supply of fuel". Can you please confirm with Al, and make necessary changes.

AlexZorkin (commented):
Should this be closed in favour of your newer branch?

Issue #1148, opened 2024-10-31
LCFS - Implement CHES Email Notification System for LCFS Based on User-Selected Actions


AlexZorkin:
**Describe the task**
Develop the backend infrastructure for integrating the BC Gov Common Hosted Email Service (CHES) within the LCFS application. This task involves setting up an email notification service that will send automated emails based on the specific notifications selected by users, as shown in the provided screenshot (e.g., Transfers, Initiative Agreements, Compliance & Supplemental Reports).

**Purpose**
This implementation will enable LCFS to send automated email notifications to users based on their preferences, improving communication and ensuring users stay informed about key actions and decisions in the system.

**Acceptance Criteria**
- [ ] Create a backend email service in LCFS that integrates with CHES for sending email notifications.
- [ ] Implement functionality to authenticate and retrieve a token from CHES.
- [ ] Develop methods for generating email content dynamically based on the type of user-selected notifications (e.g., Transfers, Initiative Agreements, Compliance & Supplemental Reports).
- [ ] Ensure that emails are sent only to users who have opted-in for specific notifications.
- [ ] Implement error handling and logging for better debugging and operational monitoring.
- [ ] Add appropriate environment variables for CHES integration
- [ ] Test the email service to ensure accurate delivery and correct content based on selected notifications.

**Technical Steps**
1. **Configure CHES Integration:**
   - Add environment variables for `EMAIL_SERVICE_CLIENT_ID`, `EMAIL_SERVICE_CLIENT_SECRET`, `CHES_AUTH_URL`, `CHES_EMAIL_URL`, `SENDER_EMAIL`, and `SENDER_NAME`.
   - Create a configuration method to load these variables from the environment.

2. **Implement Token Retrieval Method:**
   - Develop `get_email_service_token()` to request an access token from CHES using client credentials.

3. **Create Email Body Generator:**
   - Implement `generate_email_body(notification_type, details: dict)` to create the content of the email based on the selected notification type (e.g., Transfers, Initiative Agreements).

4. **Develop Email Sending Function:**
   - Implement `send_email(recipients: List[str], notification_type, details: dict)` to construct and send the email using CHES API.
   - Ensure it supports HTML formatting and uses the `Bearer` token for authorization.

5. **Integrate User Notification Preferences:**
   - Use data from user preferences (e.g., Transfers, Initiative Agreements, Compliance & Supplemental Reports) to filter which users receive emails.
   - Develop logic to retrieve and filter user email addresses based on their selected notification types.

[...]

PR #1313, opened 2024-11-29
Feat: Implement CHES email notification

prv-proton:
#1148 
image

secrets can be found here.
[lcfs-email-service-dev](https://teams.microsoft.com/l/message/19:1fc88441-3dc0-4f50-a6e9-80caec052de5_6f100c7e-b1ed-4e23-8db2-ddc1ff45ef68@unq.gbl.spaces/1732745247246?context=%7B%22contextType%22%3A%22chat%22%7D)

note: view component is only for testing.


backend/lcfs/web/api/email/repo.py
```diff
@@ -0,0 +1,71 @@
...
+            .filter(
+                NotificationType.name == NotificationTypeEnum[notification_type.upper()]
+            )
+            .limit(1)  # Fetch only one record
+        )
+
+        # Execute the query
+        result = await self.db.execute(query)
+        template = result.scalar_one_or_none()
+        return template or "default.html"
```
dhaselhan, line 71: question: Is this temporary just for testing until we get some templates implemented?

backend/lcfs/web/api/email/services.py
```diff
@@ -0,0 +1,155 @@
...
+    Service layer for sending email notifications via CHES.
+    Handles CHES-specific logic and coordinates with the repository.
+    """
+
+    def __init__(self, repo: CHESEmailRepository = Depends()):
+        self.repo = repo
+
+        # CHES configuration
+        self.config = {
+            "AUTH_URL": os.getenv("CHES_AUTH_URL", ""),
```
dhaselhan, line 28: Currently all of our ENV vars start with LCFS_ and are mapped into settings.py, should we follow that pattern here?

backend/lcfs/web/api/email/repo.py
dhaselhan: ah nvm, I see how this works. We have a generic template for some notifications but some have a specific template.

backend/lcfs/web/api/email/services.py
dhaselhan: Ah nvm, I see how this works. We have a generic default template for some notifications but others have specific content.

dhaselhan (changes requested):
Just want to make sure we are good with a new ENV var pattern.

backend/lcfs/web/api/email/services.py
prv-proton: Thanks for your review, I've made the necessary changes as per your suggestions.

prv-proton:
> Just want to make sure we are good with a new ENV var pattern.

Thanks @dhaselhan , as per your suggestion modified the code, please review it.

dhaselhan (changes requested):
Please fix the backend tests

dhaselhan (approved):
LGTM 👍

PR #1315, opened 2024-11-29
Fix Wrong Request

dhaselhan:
* Use request passed in instead of self.request which does not exist


areyeslo (approved):
Thanks for the fix!

PR #1326, opened 2024-11-30
fix: Default Summary Table inputs to 0 instead of empty string

dhaselhan:
* backend only takes int, string was failing pydantic validation
* Remove empty tag that wasn't needd


hamed-valiollahi (approved):
Looks good to me! 👍

PR #1333, opened 2024-12-02
feat: Use fuel code dates for filtering

dhaselhan:
* Use expired and effective dates to only filter active fuel codes
* Expired codes are valid for an extra year, so use start of previous year for its calculations


backend/lcfs/web/api/fuel_supply/repo.py
```diff
@@ -74,6 +77,10 @@ async def get_fuel_supply_table_options(self, compliancePeriod: str):
             .scalar_subquery()
         )
 
+        current_year = int(compliance_period)
```
AlexZorkin, line 80: a type check on this would be good

```python
try:
    current_year = int(compliance_period)
except ValueError as e:
    structlog.get_logger().error("Invalid compliance_period: not an integer", compliance_period=compliance_period, error=str(e))
    raise ValueError(f"Invalid compliance_period: '{compliance_period}' must be an integer.") from e

AlexZorkin (changes requested):

backend/lcfs/web/api/fuel_supply/repo.py
dhaselhan: Good call, nice to have some strict safety when doing conversions. Added.

AlexZorkin (approved):
👍

PR #1340, opened 2024-12-02
feat: Add sorting and filtering to Fuel Codes

dhaselhan:
* Hook up filtering and sorting methods
* Setup backend filtering lgic
* Change SetFilter to use custom so it avoids ag-grid filtering
* Don't convert filters to int so we can preserve float values


backend/lcfs/db/models/fuel/FuelCode.py
```diff
@@ -94,7 +94,7 @@ class FuelCode(BaseModel, Auditable, EffectiveDates):
     fuel_code_prefix = relationship(
         "FuelCodePrefix", back_populates="fuel_codes", lazy="joined"
     )
-    fuel_code_type = relationship(
+    fuel_type = relationship(
```
AlexZorkin, line 97: thank you for updating this, it was annoying having the discrepancy in var name

backend/lcfs/dependencies/dependencies.py
AlexZorkin, line 1: please confirm with @areyeslo that we're not stepping on each other here.

backend/lcfs/web/api/fuel_code/repo.py
```diff
@@ -189,12 +194,10 @@ async def get_transport_mode(self, transport_mode_id: int) -> TransportMode:
         )
 
     @repo_handler
-    async def get_transport_mode_by_name(self, mode_name: str) -> TransportMode:
-        query = select(TransportMode).where(TransportMode.transport_mode == mode_name)
-        result = await self.db.execute(query)
-        transport_mode = await result.scalar_one_or_none()
-
-        return transport_mode
+    async def get_transport_mode_by_name(self, transport_mode: str) -> TransportMode:
```
AlexZorkin, line 197: should add a check for when no matching TransportMode records are returned.

backend/lcfs/web/api/fuel_export/schema.py
```diff
@@ -9,12 +9,8 @@
 )
 from pydantic import Field, field_validator, validator
 
-
-class FuelTypeQuantityUnitsEnumSchema(str, Enum):
-    Litres = "L"
-    Kilograms = "kg"
-    Kilowatt_hour = "kWh"
-    Cubic_metres = "m³"
+from lcfs.web.api.fuel_code.schema import FuelCodeResponseSchema
```
AlexZorkin, line 12: 👍

frontend/src/views/FuelCodes/_schema.jsx
```diff
@@ -184,560 +205,3 @@ export const fuelCodeColDefs = (t) => [
     minWidth: 600
   }
 ]
-
-export const addEditSchema = {
```
AlexZorkin, line 188: was this just left hanging around?

AlexZorkin (changes requested):
Looking good. 

Is there no 'contains' for floating point filters? i found it a bit confusing that when i put in 37, the record with a CI of 37.21 didn't show up. Default is Equals which is pretty specific.

Sorting works great.

frontend/src/views/FuelCodes/_schema.jsx
dhaselhan: Yes, this was unused when I finished the last ticket to combine add/edit/view into a single view but missed deleting this.

backend/lcfs/dependencies/dependencies.py
dhaselhan: Arturos code is merged and I have rebased

backend/lcfs/web/api/fuel_code/repo.py
dhaselhan: Changes from scalar_one_or_none to scalar_one so it throws an exception.

dhaselhan:
> Looking good.
> 
> Is there no 'contains' for floating point filters? i found it a bit confusing that when i put in 37, the record with a CI of 37.21 didn't show up. Default is Equals which is pretty specific.
> 
> Sorting works great.

Currently there is no contains option for numbers. Lindsy wants a start with functionality so that if I enter 32, it would returns 32.1 but not 132. I will talk to Al tomorrow but I feel like we should make this its own ticket and apply it everywhere.

dhaselhan:
> > Looking good.
> > Is there no 'contains' for floating point filters? i found it a bit confusing that when i put in 37, the record with a CI of 37.21 didn't show up. Default is Equals which is pretty specific.
> > Sorting works great.
> 
> Currently there is no contains option for numbers. Lindsy wants a start with functionality so that if I enter 32, it would returns 32.1 but not 132. I will talk to Al tomorrow but I feel like we should make this its own ticket and apply it everywhere.

Created a new issue in TODO to add this new behaviour.

AlexZorkin (approved):
Thanks for creating the followup ticket. 👍

prv-proton (approved):
LGTM

PR #1351, opened 2024-12-03
Logging errors if refresh_jwk fails

areyeslo:
- Adding Logging to Redis on refresh jwk.
- Adding Logging to DocumentService on initialization.


areyeslo:
Closing this PR. The approach will be going away from Redis pool.

PR #1353, opened 2024-12-03
feat: Add schema validation to fuel code dates

dhaselhan:
* Fix existing rule for nameplate and nameplate capacity, use much simpler after validator
* Add tests


kevin-hashimoto (approved):
LGTM

PR #1361, opened 2024-12-04
Chore: Auth Middleware Tests

AlexZorkin:
- updated two main auth middleware tests after redis pool changes


areyeslo (approved):
Thank you!

PR #1366, opened 2024-12-04
Chore: Refactor Redis Client

AlexZorkin:
- Refactored redis app.state to use Redis instance instead of ConnectionPool


backend/lcfs/db/dependencies.py
```diff
@@ -49,15 +50,3 @@ async def get_async_db_session(request: Request) -> AsyncGenerator[AsyncSession,
...
-def create_redis():
-    return aioredis.ConnectionPool(
-        host=settings.redis_host,
-        port=settings.redis_port,
-        db=settings.redis_db,
-        decode_responses=True,
-    )
-
-
-pool = create_redis()
```
areyeslo, line 63: Removing redis pool

backend/lcfs/conftest.py
```diff
@@ -153,26 +154,24 @@ async def dbsession_factory(
...
+    application.state.redis_client = fake_redis_client
     application.state.settings = settings
 
     # Set up mock authentication backend with the specified roles
     set_mock_user(application, user_roles)
 
     # Initialize the cache with fake Redis backend
-    fake_redis = aioredis.FakeRedis(connection_pool=fake_redis_pool)
-    FastAPICache.init(RedisBackend(fake_redis), prefix="lcfs")
+    FastAPICache.init(RedisBackend(fake_redis_client), prefix="lcfs")
```
areyeslo, line 174: Initialize without connection pool.

backend/lcfs/services/redis/lifetime.py
```diff
@@ -10,22 +10,25 @@
...
+    Initializes the Redis client and tests the connection.
 
-    :param app: current fastapi application.
+    :param app: current FastAPI application.
     """
     try:
-        app.state.redis_pool = ConnectionPool.from_url(
-            str(settings.redis_url),
-            encoding="utf8",
+        # Initialize Redis client directly
```
areyeslo, line 18: Need to add the values for redis here because the pool configuration was removed.

backend/lcfs/web/api/redis/views.py
```diff
@@ -1,44 +0,0 @@
-from fastapi import APIRouter
```
areyeslo, line 1: Are we going to need these endpoints in the future?

areyeslo (approved):
LGTM

backend/lcfs/db/dependencies.py
AlexZorkin: We don't need this one.

backend/lcfs/conftest.py
AlexZorkin: the default param is a Redis instance not a connection pool.

backend/lcfs/services/redis/lifetime.py
AlexZorkin: Not sure what you mean, the Redis instance handles its own connections under the hood.

backend/lcfs/web/api/redis/views.py
AlexZorkin: no, these were just placeholders left over from the inital app template

PR #1367, opened 2024-12-04
feat: Add front end feature flags

dhaselhan:
* Update config with new feature flags
* Add new helper function for checking if one is enabled
* Follow the withRole pattern to add a withFeatureFlag HOC


frontend/src/utils/__tests__/withRole.test.jsx
AlexZorkin, line 1: thanks for adding this.

AlexZorkin (approved):
Looks great ty 👍

PR #1368, opened 2024-12-04
Fix: Default allocation agreement to 1 row on empty data

AlexZorkin:
- Added check for empty data and creates default row
- Added test for addeditallocationagreement component


dhaselhan (approved):
LGTM 👍

PR #1369, opened 2024-12-04
fix: Data and Label Fixes

dhaselhan:
* Change hydrogen to kg
* Relabel allocation agreement transaciton types
* Rename first column to Responsibility
* Add $ infront of CAD
* Change other Diesel to Fossil Derived - False


hamed-valiollahi (approved):
Looks good to me! 👍

PR #1376, opened 2024-12-05
fix: Update front end allocation agreement options

dhaselhan:
* Previously updated backend, but they are also hard coded in the front end


hamed-valiollahi (approved):
LGTM 👍

dhaselhan:
This is going to majorly conflict with the original feature to sort / filter columns.
https://github.com/bcgov/lcfs/pull/1340

I really didnt expect this to get reviewed first.

PR #1377, opened 2024-12-05
fix: Load EER ratio even when end use is not defined

dhaselhan:


AlexZorkin (changes requested):
Looks good but this seems like a good place to have some test coverage.

AlexZorkin (approved):
Looks good 👍

PR #1379, opened 2024-12-05
fix: Change Header padding and sizing

dhaselhan:
* Add max width to name and set it to truncate to ...
* Drop padding to make more room for name
* Delete old unused header
* Set menu items to not wrap text


hamed-valiollahi (approved):
LGTM 👍

Issue #1175, opened 2024-11-06
LCFS - Edit options to LNG "End use" in the compliance report Fuel supply and Export sections


Grulin:
**Describe the task**
Edit/replace the LNG fuel type "End Use" drop-down options of the Fuel Supply and Export sections of the Compliance Report AG grids, with options matching specific LNG use cases and pre-populate the corresponding EER and UCI values. The choices available in the "End Use" dropdown should align with the distinct scenarios outlined in the UCI values, ensuring consistency in both data entry and reporting.

**Purpose**
To standardize and clarify end-use categories for fuel reporting in the Compliance Report. This column will allow users to select from predefined end-use options, streamlining the reporting process and aligning with UCI-specific data requirements.

**Acceptance Criteria**

- [ ] The "End Use" dropdown includes the following options and associated pre-populated cells for the EER and UCI columns:

End Use | EER value | Additional CI attributable to use (UCI) value

- Compression-ignition engine- Marine, general | 1 | 27.3
- Compression-ignition engine- Marine, operated within 51 to 75% of load range | 1 | 17.8
- Compression-ignition engine- Marine, operated within 76 to 100% of load range | 1 | 12.2
- Compression-ignition engine- Marine, with methane slip reduction kit- General | 1 | 10.6
- Compression-ignition engine- Marine, with methane slip reduction kit- Operated within 51 to 75% of load range | 1 | 8.4
- Compression-ignition engine- Marine, with methane slip reduction kit- Operated within 76 to 100% of load range | 1 | 8
- Compression-ignition engine- Marine, unknown whether kit is installed or average operating load range | 1 | 27.3
- Unknown engine type | 0.9 | 27.3
- Other (i.e. road transportation) | 0.9 | 0

- [ ] The dropdown options align exactly with those listed for the "Additional CI Attributable to Use (UCI) - gCO₂e/MJ" column, supporting consistent data selection across related fields.
- [ ] The new "End Use" column is functional and visible for both IDIR and BCeID users.

**Additional information:**

Confirmation of end use choices and associated EERs and UCIs has been confirmed. 
Adjust column width to the values in the select list

PR #1383, opened 2024-12-06
feat: migration for end use types + migration fix

kevin-hashimoto:
- feat: migration for end use type table row updates, #1175


AlexZorkin (approved):
Looks good. Thanks for fixing the downgrade. 👍

PR #1390, opened 2024-12-06
feat: Add Cypress test for entering FS

dhaselhan:
* Expand existing cypress test to enter an FS and confirm the values show in the compliance report summary


dhaselhan:
Part of the messiness of this one is that we don't have a loading indicator for compliance summary, so it shows the old values when the page is first loaded, and we just have to wait for an estimated amount.

kevin-hashimoto (approved):
greeeen LGTM

Issue #1226, opened 2024-11-13
LCFS - Implement Email Notification Triggers in Backend for Subscribed Users


AlexZorkin:
**Description:**

Implement email notification triggers throughout the backend codebase. For example, when a compliance report is assessed, users from the organization that owns the report should receive an email notification if they are subscribed to that notification type. Each notification type should have its trigger implemented at the appropriate location in the backend.

**Purpose and benefit to user:**

- Keeps users informed about important actions related to them.
- Enhances user engagement and satisfaction.
- Ensures timely communication of relevant events.

**Acceptance Criteria:**

- [ ] Given I am a **user subscribed to a notification type**, when the corresponding event occurs (e.g., compliance report is assessed), then I receive an email notification.
- [ ] Given I am a **user not subscribed to a notification type**, when the event occurs, then I do **not** receive an email notification.
- [ ] Given the system triggers an event, notifications are sent only to users subscribed to that notification type.

**Development Checklist:**

- [ ] Identify all notification types and their triggering events.
- [ ] Implement notification triggers at the correct locations in the backend code.
- [ ] Ensure notifications respect user subscription preferences.
- [ ] Integrate triggers with the existing email templates.
- [ ] Test to confirm that notifications are sent correctly to subscribed users.

**Notes:**

- Coordinate with the business team to verify notification requirements.
- Handle user data securely when sending emails.
- Ensure that adding new notification types in the future is straightforward.

airinggov:
Email template

Subject: B.C. LCFS Notification
From: lcfs@gov.bc.ca
Body:
This email was generated by the Government of British Columbia, Low Carbon Fuel Standard portal. 
A <transaction type> update has occurred within the Low Carbon Fuel Standard portal. For more details, please go to: https://lowcarbonfuels.gov.bc.ca/<full-url>
You received this email because you subscribed at the site above, to stop receiving these email, log in to your account here: https://lowcarbonfuels.gov.bc.ca/notifications

ANALYST
Transfers
•	Submitted to government for analyst review
o	**SUBMITTED** (Signed and submitted by transfer partner 2)
•	Rescinded by either transfer partner
o	**RESCINDED** (this should only trigger if the report was previously Submitted but not Sent)
•	Director recorded/refused
o	**RECORDED** or **REFUSED**

Initiative agreements and other transactions
•	Director approved/returned to analyst
o	**APPROVED** or **RETURNED**  (initiative agreements, designated actions, admin adjustments)

[...]

PR #1402, opened 2024-12-09
LCFS - Implement Email Notification Triggers in Backend for Subscribed Users #1226

prv-proton:
#1226


backend/lcfs/web/api/base.py
AlexZorkin, line 1: is there a better place to keep this in the backend? Or is it used in enough places to justify having it in base?

AlexZorkin (approved):
Looks really good, just one minor question. 👍

backend/lcfs/web/api/base.py
prv-proton: Thanks Alex for the review. Since this schema will be used across various services, it causes greenlet spawn error if placed inside the notifications schema, hence it's being placed inside the base.

PR #1405, opened 2024-12-09
fix: Allow negative symbol in transaction amount input

dhaselhan:


areyeslo (approved):
LGTM. Just need to rebase.

PR #1409, opened 2024-12-09
feat: Bypass auth for health check

dhaselhan:
* Add code to bypass auth check for health endpoint

### DIFF
--- a/backend/lcfs/web/api/monitoring/views.py
+++ b/backend/lcfs/web/api/monitoring/views.py
@@ -4,9 +4,10 @@ router = APIRouter()
 
 
 @router.get("/health")
-def health_check() -> None:
+def health_check() -> str:
     """
     Checks the health of a project.
 
     It returns 200 if the project is healthy.
     """
+    return "healthy"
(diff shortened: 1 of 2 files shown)


AlexZorkin (approved):
👍

prv-proton (approved):
Looks good to me. But just a suggestion, can we have health check report something like below to get the status of connections?

`{
  "status": "DOWN",
  "components": {
    "database": {
      "status": "UP",
      "details": {
        "response_time_ms": 10
      }
    },
    "cache": {
      "status": "UP",
      "details": {
        "cache_hits": 150,
        "cache_misses": 10
      }
    },
  }
}
`

dhaselhan:
> Looks good to me. But just a suggestion, can we have health check report something like below to get the status of connections?
> 
> `{ "status": "DOWN", "components": { "database": { "status": "UP", "details": { "response_time_ms": 10 } }, "cache": { "status": "UP", "details": { "cache_hits": 150, "cache_misses": 10 } }, } } `

This endpoint is for Openshift where its just used to ping to check for uptime. If we want a more in depth health check that reports to some other system I am good with that but it should be its own endpoint.

PR #1269, opened 2024-11-21
LCFS-1207: Adding new columns to Other Use

areyeslo:
Add new columns to the Fuels for Other Use AG grid table in the compliance report and Report Details:
- Determining Carbon Intensity
- Fuel Code
- CI of Fuel
image

image

PR #1411, opened 2024-12-09
LCFS-1207: Update fuel type for other_uses

areyeslo:
Fuel type for other uses should be the following:

- Alternative Jet Fuel
- Biodiesel
- Ethanol
- Fossil-Derived Diesel
- Fossil-Derived Gasoline
- Fossil-Derived Jet Fuel
- HDRD (Hydrotreated Renewable Diesel)
- Other Diesel Fuel
- Renewable Gasoline

[Story](https://github.com/orgs/bcgov/projects/137/views/1?filterQuery=ARE&pane=issue&itemId=86585124&issue=bcgov%7Clcfs%7C1207)


backend/lcfs/db/migrations/versions/2024-12-09-22-33_cd8698fe40e6.py
```diff
@@ -11,7 +11,7 @@
 
 # revision identifiers, used by Alembic.
 revision = "cd8698fe40e6"
-down_revision = "9206124a098b"
+down_revision = "7ae38a8413ab"
```
AlexZorkin, line 14: please double check this migration id is in the correct order.

AlexZorkin (changes requested):
For this ticket it was required to add 3 new columns to the other_use table in the frontend. Was this completed in a separate ticket?

backend/lcfs/db/migrations/versions/2024-12-09-22-33_cd8698fe40e6.py
areyeslo: updating this another migration appeared before mine.

areyeslo:
The addition of the fields was already merged https://github.com/bcgov/lcfs/pull/1269

This is a follow up ticket.

PR #1416, opened 2024-12-09
Fix: Remove Notification Email and Use Profile Email

hamed-valiollahi:
This PR removes the notification email field and updates the logic to use the profile email for notifications.


prv-proton (approved):
Looks good

Issue #1302, opened 2024-11-27
LCFS - BUG AG Grid copy and paste function for Fuel codes table on IDIR side not functioning


Grulin:
**Describe the Bug**
The copy and paste function in the Fuel Code Input Form works correctly for most columns but fails for the Prefix and Fuel columns. When testing, these columns do not accept pasted values, while other columns in the form function as expected. The Prefix and Fuel columns should also support the copy and paste feature to improve data entry efficiency.

**Expected Behaviour**
Users should be able to copy and paste values into the Prefix and Fuel columns in the Fuel Code Input Form, just like the other columns.

**Actual Behaviour**

- The Prefix and Fuel columns do not accept pasted values, while other columns in the form work correctly with the copy and paste functionality.

- The prefix column remains empty and the fuel column continues to show the "Select" option in the field, rather than the fuel type that was on the Excel sheet. 

![Image](https://github.com/user-attachments/assets/8a3f9c1f-79fa-446e-b267-a7e98f482d23)
![Image](https://github.com/user-attachments/assets/eb973472-cf87-4dd7-854c-feab6bc021ec)

Used the following spreadsheet to test:
[Testing of Fuel Code form copy and paste.xls](https://github.com/user-attachments/files/17939942/Testing.of.Fuel.Code.form.copy.and.paste.xls)

**Implications**

- Users cannot use the copy and paste feature to quickly input data into the Prefix and Fuel columns, leading to inefficient workflows and increased manual data entry.

- Also prevents users from saving fuel codes.

**Steps To Reproduce**
Steps to reproduce the behaviour:
User/Role:

1. Navigate to the Fuel Code Input Form.
2. Attempt to copy a value from another source (e.g., spreadsheet, text file).
3. Paste the value into any column except Prefix and Fuel and observe that it works correctly.
4. Attempt to paste the same value into the Prefix or Fuel columns.
5. Observe that the pasted value does not populate the field.

Grulin:
Hi @prv-proton! The prefix column input is working as expected now, but the "Fuel" column still isn't populating with the copy and paste function in the AG grid for fuel codes on the IDIR side.

PR #1420, opened 2024-12-10
Fix: LCFS - BUG AG Grid copy and paste function for Fuel codes table …

prv-proton:
…on IDIR side not functioning #1302
- fix for create fuel code function
- for suffix increment version once the subversion reaches 9
- raise ValueErrors when invalid data is passed for save and propagate this error back to the user.
- In the action Tooltip renderer show validation failure message.
- fix delete endpoint for fuel codes.
- create separate paste handler for fuel code.
- Fuel field name is changed to FuelType

### DIFF
--- a/frontend/src/views/FuelCodes/AddFuelCode/AddEditFuelCode.jsx
+++ b/frontend/src/views/FuelCodes/AddFuelCode/AddEditFuelCode.jsx
@@ -25,6 +25,7 @@ import BCModal from '@/components/BCModal'
 import BCTypography from '@/components/BCTypography'
 import { FUEL_CODE_STATUSES } from '@/constants/statuses'
 import { useCurrentUser } from '@/hooks/useCurrentUser'
+import Papa from 'papaparse'
 
 const AddEditFuelCodeBase = () => {
   const { fuelCodeID } = useParams()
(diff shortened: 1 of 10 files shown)


frontend/src/views/FuelCodes/AddFuelCode/AddEditFuelCode.jsx
```diff
@@ -229,10 +231,12 @@ const AddEditFuelCodeBase = () => {
           const errMsg = `Error updating row: ${
             fieldLabels.length === 1 ? fieldLabels[0] : ''
           } ${message}`
-
+          updatedData.validationMsg = errMsg
```
areyeslo, line 234: Tested and received 422, when processing the date fields. They are not in correct format. Then, needed to give a proper format. Something like this:
```const dateFields = ['applicationDate', 'approvalDate', 'effectiveDate', 'expirationDate'];

      // Format date fields to ISO 8601
      const formattedData = { ...params.node.data };
      dateFields.forEach((field) => {
        if (formattedData[field]) {
          const date = new Date(formattedData[field]);
          if (!isNaN(date.getTime())) {
            formattedData[field] = date.toISOString().split('T')[0]; // Keep only the YYYY-MM-DD part
          }
        }
      });
      let updatedData = Object.entries(formattedData)
```
After that the request passes the schema validation, it can start executing create_fuel_code. However, an error is raised: ValueError: Invalid transport mode: Truck

prv-proton: Thanks for your review @areyeslo , 
- yes, from excel data needs to be copied in proper format otherwise it throws error.
- Truck is no longer among the valid transport mode in the database. hence value error.

PR #1428, opened 2024-12-11
Feat: Compliance Report ETL

AlexZorkin:
- new groovy file for compliance report etl
- local nifi script to run etl in local databases


etl/nifi_scripts/compliance_report.groovy
```diff
@@ -0,0 +1,647 @@
...
+                report.manager_status_id,
+                report.director_status_id,
+                referenceData
+            )
+
+            def supplementalInitiator = report.is_supplemental ? "SUPPLIER_SUPPLEMENTAL" : null
+            def reportingFrequency = "ANNUAL"
+
+            // create_user and update_user are always strings
+            def reportCreateUser = "imported_user"
```
prv-proton, line 185: updating the data as imported_user might cause issues w.r.t displaying user names in UI history log.

prv-proton (approved):
Looks good 👍. Just a suggestion, we could bring in the snapshot json data (compliance_report_snapshot) into a new column within compliance report table, so that it'll be easier to render legacy data in UI.

etl/nifi_scripts/compliance_report.groovy
AlexZorkin: yes good catch, i will update.

PR #1431, opened 2024-12-11
fix: ci fuel for other fuel type

### LINKED ISSUE #1400: LCFS - BUG -  fuel type Other, CI of fuel value
**Describe the Bug:**
When selecting the fuel type "Other," the 'CI of fuel' value does not populate or display properly. The CI value briefly flashes on the screen and then resets back to zero.

**Expected Behaviour:**
The 'CI of fuel' should populate correctly based on the input and remain visible without resetting to zero.

**Actual Behaviour:**
- The 'CI of fuel' value briefly flashes up but resets to zero immediately afterward.
- This behavior occurs specifically when the fuel type "Other" is selected.
- 
Image

**Implications:**
- Prevents accurate calculation for "Other" fuel types.
- May result in incomplete or incorrect data in compliance reports.

**Steps To Reproduce:**
1. User/Role: BCeID
2. Navigate to the add fuel supply.
3. Select the fuel type "Other."
4. Observe the CI value briefly flashing and then resetting to zero.

**Additional Context:**

kevin-hashimoto:
- fix: ci fuel appearing as 0 when fuel type is 'other'

closes #1400


AlexZorkin (approved):
👍

PR #1439, opened 2024-12-11
Chore: Parallel Test Workflow

AlexZorkin:
- updated docker-auto-test with separate jobs for frontend and backend tests


dhaselhan (approved):
LGTM 👍

PR #1442, opened 2024-12-12
hotfix: Replace Typography with BCTypography

areyeslo:
User can not access FinaL Supply Equipment due to error in Typography not found. This component was replaced for BCTypography.


hamed-valiollahi (approved):
LGTM! 👍

PR #1447, opened 2024-12-12
hotfix: Provide default value to organization_name field in FSE on creation

areyeslo:
Provide default value for existing records in FSE.


AlexZorkin (approved):

hamed-valiollahi (changes requested):
Existing migration files shouldn't be modified, as changes won't reflect in production. Please create a new migration file instead.

hamed-valiollahi (approved):

areyeslo:
because if a migration hasn't run in test and it is still in dev. So the migration will be applied in test and in Prod when we will go live.

hamed-valiollahi:
That makes sense! Thank you for the clarification, Arturo!

PR #1466, opened 2024-12-16
feat: Sync reports from TFRS -> LCFS

dhaselhan:
* Add logic to handle create, submit, and approve messages
* Add a function / test to handle logic for each one
* Add legacy_id to track the id from TFRS
* Change consumer to use app state directly since it has no request


AlexZorkin (approved):
Looks great, nice breakdown of the different action types. 👍

Issue #1426, opened 2024-12-11
LCFS - Enable IDIR analyst to upload file attachments to compliance reports


airinggov:
**Description:**
Enable the IDIR analyst to upload file attachments to compliance reports or supplemental reports that are in either Submitted or Assessed status. Provide a 'Supporting documents' accordion with an edit icon that opens the upload utility.

**Wireframe:**
https://preview.uxpin.com/da68049c0af15edd7ce4e47e4bee431fed40c30d#/pages/194028770

**Purpose and Benefit to User:**
This feature allows the IDIR analyst to add additional documents received after submission, ensuring that the compliance report or supplemental report contains a complete record of all relevant information. It streamlines the process of managing supplemental documentation and improves the accuracy and reliability of record-keeping.

**Acceptance Criteria:**
- [ ] Display the 'Supporting documents' accordion on compliance reports and supplemental reports even if no documents were attached previously.
- [ ] Provide an edit icon within the 'Supporting documents' accordion (when the report is in Submitted or Assessed status) to open the file upload utility.
- [ ] Allow the upload of file attachments to these reports.
- [ ] Ensure that uploaded documents are saved and associated with the appropriate report.

**Development Checklist:**
- [ ] Ensure the 'Supporting documents' accordion is displayed, even if no documents have been uploaded.
- [ ] Implement an edit icon that triggers the upload utility.

**Notes:**

airinggov:
@areyeslo looks great Arturo

PR #1471, opened 2024-12-17
LCFS-1426: Upload file attachements compliance reports for Analyst user

areyeslo:
Extend the logic of Supporting Documents to allow Analyst to attach Supporting Documents as long as the compliance report status is submitted or assessed. Supporting documents section is visible at all times.

**New Logic:**

- BCeID user can attach supporting documents to a compliance report under `draft`:

image

- Analyst user can check the `submitted` report and attach supporting documents:

image

- Under `submitted` report, BCeID user can see the supporting documents section, but not modify docs in the section:

image

### DIFF
--- a/backend/lcfs/web/api/document/views.py
+++ b/backend/lcfs/web/api/document/views.py
@@ -40,7 +40,7 @@ async def get_all_documents(
     response_model=FileResponseSchema,
     status_code=status.HTTP_201_CREATED,
 )
-@view_handler([RoleEnum.SUPPLIER])
+@view_handler([RoleEnum.SUPPLIER, RoleEnum.ANALYST])
 async def upload_file(
     request: Request,
     parent_id: int,
--- a/backend/lcfs/web/api/document/views.py
+++ b/backend/lcfs/web/api/document/views.py
@@ -1,3 +1,6 @@
+from http.client import HTTPException
+from lcfs.db.models.compliance.ComplianceReportStatus import ComplianceReportStatusEnum
+from lcfs.web.api.compliance_report.services import ComplianceReportServices
 import structlog
 from typing import List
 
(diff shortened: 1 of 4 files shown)


backend/lcfs/web/api/document/views.py
```diff
@@ -40,7 +40,7 @@ async def get_all_documents(
     response_model=FileResponseSchema,
     status_code=status.HTTP_201_CREATED,
 )
-@view_handler([RoleEnum.SUPPLIER])
+@view_handler([RoleEnum.SUPPLIER, RoleEnum.ANALYST])
```
prv-proton, line 43: Just a question, can the supplier upload the files, even after the compliance report is submitted using this endpoint? as this could pose a security threat.

prv-proton (changes requested):
can you please validate the endpoint for any security threats.

backend/lcfs/web/api/document/views.py
areyeslo: There is no possibility for the supplier to upload files when the report is submitted in the Front End. However, I can add a condition in the backend as well.

prv-proton: yes, please. Let's add some validations at the backend as well.

backend/lcfs/web/api/document/views.py
```diff
@@ -40,14 +43,27 @@ async def get_all_documents(
...
     document_service: DocumentService = Depends(),
+    compliance_report_service: ComplianceReportServices = Depends(),
 ) -> FileResponseSchema:
+    # Fetch the compliance report
+    compliance_report = await compliance_report_service.get_compliance_report_by_id(parent_id)
+    if not compliance_report:
+        raise HTTPException(status_code=404, detail="Compliance report not found")
+
+    # Check if the compliance report is submitted
+    if compliance_report.current_status.status == ComplianceReportStatusEnum.Submitted:
```
prv-proton, line 61: - Can you please move this logic to service.py or validation.py?
- Restriction to upload file when ComplianceReport is not in Draft status applies only to Supplier users. Analyst should be able to upload when it's in Submitted status as per LCFS #1426 , "Enable the IDIR analyst to upload file attachments to compliance reports or supplemental reports that are in either Submitted or Assessed status. "

prv-proton (changes requested):
It's not meeting all the acceptance criteria.

prv-proton (approved):
Looks good

PR #1475, opened 2024-12-17
bug fix for ag-grid checkbox

prv-proton:
updating code as per new migration: https://www.ag-grid.com/react-data-grid/row-selection/


areyeslo (approved):
LGTM

PR #1477, opened 2024-12-17
feat: Add validation rule for other_fuel_type

dhaselhan:
* Validate that if the fuel type is unrecognized we require other to be populated
* Add new tests
* Fix broken tests in develop


hamed-valiollahi (approved):
LGTM! 👍

PR #1478, opened 2024-12-17
feat: toast on 400+ errors

### LINKED ISSUE #1441: LCFS - Show red toast on network errors
As a tester of LCFS, I want to know as soon as a network failure happens so I can better report issues to developers.

Currently network errors fail silently, often leaving the loading spinner or updating message. We want to use react material toast handler to show failure toasts when a 400+ response is received.

kevin-hashimoto:
- feat: 400+ errors display an error toast globally using notistack(mui underneath) for imperative handling. closes #1441 

note: mui v6 has their own implementation via their toolpad package. would be good to look into later when/if upgrading mui to v6.

### DIFF
--- a/frontend/package-lock.json
+++ b/frontend/package-lock.json
@@ -40,6 +40,7 @@
         "material-ui-popup-state": "^5.0.10",
         "moment": "^2.30.1",
         "mui-daterange-picker-plus": "^1.0.4",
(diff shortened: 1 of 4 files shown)


dhaselhan (approved):
Definitely agree that Material V5 Snackbar is lacking an imperative interface. Nice work.

PR #1481, opened 2024-12-17
fix: Round all the things

dhaselhan:
* Convert all column in summary that store quantity or units to integer as they should never have decimals
* Update summary calculation to round unit penalties


AlexZorkin (changes requested):
We will have to review this with Al because there is some policy around *when* the rounding should occur. I believe they want to store the precision in floats in the db and round only at the last possible second...

PR #1482, opened 2024-12-17
feat: Add Legacy Report View

dhaselhan:
* Add new brancher to control legacy vs non-legacy view
* Split HistoryCard to its own component for re-usability and add tests
* Update legacy view with legacy assessment card
* Sent legacy_id to the front end


frontend/public/config/config.js
```diff
@@ -1,5 +1,6 @@
 export const config = {
   api_base: 'http://localhost:8000/api',
+  tfrs_base: 'http://localhost:3001',
```
AlexZorkin, line 3: in local we're using port 3000

frontend/src/App.jsx
```diff
@@ -196,7 +198,7 @@ const router = createBrowserRouter([
       },
       {
         path: ROUTES.REPORTS_VIEW,
-        element: <EditViewComplianceReport />,
+        element: <ViewComplianceReportBrancher />,
```
AlexZorkin, line 201: maybe a different name 'ComplianceReportViewSelector'?

AlexZorkin (changes requested):
Looks great, only 2 comments.

frontend/public/config/config.js
dhaselhan: I set this to 3001 as LCFS also runs on 3000 so if you were to run both locally at the same time TFRS would need to be on something else. Should I set them to the same?

frontend/src/App.jsx
dhaselhan: Done!

frontend/public/config/config.js
AlexZorkin: I do like the idea but there are quite a few places we would have to update this. vite config, cypress config, application settings, and a few more spots. Maybe leave it as 3000 and then adjust locally when testing this hybrid code. If you want to go through and update it everywhere im not opposed but we would have to test thoroughly.

dhaselhan: Updated to 3000.

AlexZorkin (approved):
👍

PR #1484, opened 2024-12-17
fix: Sorting & Display of FSE

dhaselhan:
* Sort both paginated and non paginated by create date asc to match other schedules
* Fix kWH usage column to use nice number formatter without decimals

### DIFF
--- a/backend/lcfs/web/api/final_supply_equipment/repo.py
+++ b/backend/lcfs/web/api/final_supply_equipment/repo.py
@@ -201,7 +201,7 @@ class FinalSupplyEquipmentRepository:
                 joinedload(FinalSupplyEquipment.level_of_equipment),
             )
             .where(FinalSupplyEquipment.compliance_report_id == report_id)
-            .order_by(FinalSupplyEquipment.final_supply_equipment_id)
+            .order_by(FinalSupplyEquipment.create_date.asc())
         )
         return result.unique().scalars().all()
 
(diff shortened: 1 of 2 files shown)


hamed-valiollahi (approved):
Looks good to me!

As a suggestion, would sorting by `id` be more efficient here since integer comparisons are typically faster than timestamp comparisons like `created_at`?

dhaselhan:
> Looks good to me!
> 
> As a suggestion, would sorting by `id` be more efficient here since integer comparisons are typically faster than timestamp comparisons like `created_date`?

Depending on how ETL works this may or may not be true. ETL may insert records with a higher ID but a copied Create Date from TFRS. Ill ask Alex.

hamed-valiollahi:
> > Looks good to me!
> > As a suggestion, would sorting by `id` be more efficient here since integer comparisons are typically faster than timestamp comparisons like `created_date`?
> 
> Depending on how ETL works this may or may not be true. ETL may insert records with a higher ID but a copied Create Date from TFRS. Ill ask Alex.

Thank you! In case there is significant improvement, we can also update the ETL to ensure the correct order of fetching data. Also, I noticed that there are a few places in the database where the tables don’t have proper indexing. It would be great if we could check and address that as well.

dhaselhan:
> > > Looks good to me!
> > > As a suggestion, would sorting by `id` be more efficient here since integer comparisons are typically faster than timestamp comparisons like `created_date`?
> > 
> > 
> > Depending on how ETL works this may or may not be true. ETL may insert records with a higher ID but a copied Create Date from TFRS. Ill ask Alex.
> 
> Thank you! In case there is significant improvement, we can also update the ETL to ensure the correct order of fetching data. Also, I noticed that there are a few places in the database where the tables don’t have proper indexing. It would be great if we could check and address that as well.

Talked to Alex, he mentioned we are transferring Create Date so we can use that.

hamed-valiollahi:
Nice, thanks, Daniel!

PR #1486, opened 2024-12-17
fix issue of updating gov users with "Government of B.C." in database.

prv-proton:


dhaselhan (approved):
LGTM 👍 
I wonder why this code was ever there, its so odd

PR #1490, opened 2024-12-18
feat: Add default CI to Fuel Category

dhaselhan:
* Default CI is actually associated to fuel category and used as a fallback in several scenarios
* The one covered here is Other, where if the user selects other it needs to use the CI of the Cateory, NOT the Type.


kevin-hashimoto (approved):
LGTM - correct ci values are displayed

PR #1491, opened 2024-12-18
feat: clear alert on action

### LINKED ISSUE #1456: BUG
**Describe the Bug:** 
Error messages on the Fuel Supply Input page do not clear properly when a row is deleted. 
Image

**Expected Behaviour:** 
When a row is deleted, all associated error messages should be removed from the page to ensure a clean and accurate user interface.

**Actual Behaviour:** 
The error message sometimes persists on the page even after the row that caused the error is deleted.

**Implications:** 
- Confusing user experience as it appears that the error is unresolved.
- Potential frustration for users who may believe they need to take additional action to clear the error.
- Possible interference with subsequent input validation and form submissions.

**Steps To Reproduce:** 
**User/Role:** Any user with access to the Fuel Supply Input page.
1. Go to the Fuel Supply Input page.
2. Add a new row to the input table.
3. Select a fuel type but leave the "Determining carbon intensity" field empty.
4. Observe the error message “Determining carbon intensity Field required.”
5. Delete the row you just added.
6. Observe that the error message persists even after the row is deleted.

kevin-hashimoto:
- feat: alerts now clear when actions are taken so that they do not persist when they dont need to, closes #1456


AlexZorkin (approved):
👍

PR #1492, opened 2024-12-18
fix: filter showing on other uses summary

### LINKED ISSUE #1458: LCFS - BUG - Other Use display table is incorrectly showing filtering options
**Describe the Bug:** 
The "Other Use" display table is incorrectly showing filtering options. Filtering was not intended for this area and is inconsistent with other schedules.

Image

**Expected Behaviour:** 
The "Other Use" display table should not show filtering options, maintaining consistency with the other schedules.

**Actual Behaviour:** 
Filtering options are visible in the "Other Use" display table, which is not intended and creates inconsistency in the user interface.

**Implications:** 
- Potential confusion for users due to inconsistent design.

**Steps To Reproduce:** 
**User/Role:** Any user with access to compliance reports.
1. Navigate to the "Other Use" display table.
2. Observe the presence of filtering options in the table header.
3. Compare with other schedules that do not display filtering options.

kevin-hashimoto:
- fix: filters showing in header of other uses summary in compliance reports page, closes #1458

### DIFF
--- a/frontend/src/views/OtherUses/OtherUsesSummary.jsx
+++ b/frontend/src/views/OtherUses/OtherUsesSummary.jsx
@@ -109,6 +109,7 @@ export const OtherUsesSummary = ({ data }) => {
           gridKey={'other-uses'}
           getRowId={getRowId}
           columnDefs={columns}
+          defaultColDef={{ filter: false, floatingFilter: false }}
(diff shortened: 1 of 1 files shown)


AlexZorkin (approved):
👍

PR #1499, opened 2024-12-18
fix: grid alert not returning correct alert

kevin-hashimoto:
- fix: fixes a bug where cell changes dont return the correct alert


areyeslo (approved):
LGTM

PR #1504, opened 2024-12-18
fix: Round non-currency numbers in Summary Table

dhaselhan:
* Update number formatter to take max decimals
* Set max decimals to 0 on Summary Page


AlexZorkin (approved):
Rounding is fun.

PR #1505, opened 2024-12-18
fix: Disable navigating unless report is draft

dhaselhan:
* Only execute navigate if report is in Draft status to prevent draft clicks from going to edit


kevin-hashimoto (approved):
👍🏻 clicking on rows outside of draft status no longer go to the edit page

PR #1521, opened 2024-12-19
LCFS-1515: Fuel supply, Other, Jet fuel category

areyeslo:
When selecting a `Fuel Type` that has multiple `Fuel Categories`, the first element of the category is selected. However, this can cause that user submits data that did not intend to include.

With the change the `Fuel Category` is cleared when there are multiple elements.

image

[Story](https://github.com/orgs/bcgov/projects/137/views/1?pane=issue&itemId=91529565&issue=bcgov%7Clcfs%7C1515)


frontend/src/views/FuelSupplies/_schema.jsx
dhaselhan, line 1: With these new changes, do we still need the logic in the valueGetter to default to the 0th option?

    valueGetter: (params) => {
      const options = optionsData?.fuelTypes
        ?.find((obj) => params.data.fuelType === obj.fuelType)
        ?.fuelCategories.map((item) => item.fuelCategory)
      if (options?.length === 1) {
        return options[0]
      } else {
        return params.data.fuelCategory
      }

frontend/src/views/FuelSupplies/_schema.jsx
```diff
@@ -145,9 +145,9 @@ export const fuelSupplyColDefs = (optionsData, errors, warnings) => [
     minWidth: 250
   },
   {
-    field: 'fuelCategory',
+    field: 'fuelCategoryId',
```
dhaselhan, line 148: The value getter is still using return params.data.fuelCategory, why was this changed?

frontend/src/views/FuelSupplies/_schema.jsx
areyeslo: Removed. This was change to get directly from params.data.fuelCategory because the logic is under onCellValueChanged.

frontend/src/views/FuelSupplies/_schema.jsx
areyeslo: Changed back.

frontend/src/views/FuelSupplies/_schema.jsx
```diff
@@ -147,7 +148,7 @@ export const fuelSupplyColDefs = (optionsData, errors, warnings) => [
   {
     field: 'fuelCategory',
     headerComponent: RequiredHeader,
-    headerName: i18n.t('fuelSupply:fuelSupplyColLabels.fuelCategory'),
+    headerName: i18n.t('fuelSupply:fuelSupplyColLabels.fuelCategoryId'),
```
dhaselhan, line 151: Why are you changing the label to have a different name than the column?
fuelCategory

dhaselhan (changes requested):

frontend/src/views/FuelSupplies/_schema.jsx
areyeslo: How we are showing errors in the AG Grids? We are saving the records using the `useSaveFuelSupply` hook which sends a POST request to backend. Then Pydantic validates the payload, if there is any missing field then this is returned as response:
```
        {
            "type": "missing",
            "loc": [
                "body",
                "fuelCategoryId"
            ],
            "msg": "Field required",
            "input": {
                "complianceReportId": "1",
                "id": "a9d30ac2-ba2d-47e2-a655-00e27edc71c2",
                "fuelType": "CNG",
                "fuelTypeId": 2,
                "units": "m³",
                "unrecognized": false,
                "modified": true,
                "validationStatus": "pending"
            }
        },
```
Then, frontend receives the response, and if any error, then this is catched and then we iterate over the errors returned by the backend.  We need fo have the same field in the front end. Otherwise, how are we going to map the errors?

Should we do in a different way?

dhaselhan: No, I was just confused because the error is working in your screenshot and the MR description doesn't mention fixing an error message so I wasn't sure why this was being changed.

dhaselhan (approved):
LGTM 👍

PR #1526, opened 2024-12-19
fix: Use compliance period from the Compliance Report

dhaselhan:
* Pass the compliance period from the report instead of reading from the front end data
* Add unique constraint on 1 TCI per Category per Period
* Change query to return a single TCI and fail otherwise
* Update tests


AlexZorkin (approved):
Looks good, nice adding compliance period safety. 👍

PR #1548, opened 2024-12-20
fix: Add conditional check to cellRenderer

dhaselhan:
* Move from onRowClick to LinkRenderer and update linkRendeder to only be used if its a draft report

### DIFF
--- a/frontend/src/views/FinalSupplyEquipments/FinalSupplyEquipmentSummary.jsx
+++ b/frontend/src/views/FinalSupplyEquipments/FinalSupplyEquipmentSummary.jsx
@@ -1,12 +1,12 @@
 import BCAlert from '@/components/BCAlert'
 import BCBox from '@/components/BCBox'
 import BCDataGridServer from '@/components/BCDataGrid/BCDataGridServer'
-import { apiRoutes, ROUTES } from '@/constants/routes'
-import { CommonArrayRenderer } from '@/utils/grid/cellRenderers'
+import { apiRoutes } from '@/constants/routes'
+import { CommonArrayRenderer, LinkRenderer } from '@/utils/grid/cellRenderers'
 import Grid2 from '@mui/material/Unstable_Grid2/Grid2'
 import { useEffect, useMemo, useRef, useState } from 'react'
 import { useTranslation } from 'react-i18next'
-import { useLocation, useParams, useNavigate } from 'react-router-dom'
+import { useLocation, useParams } from 'react-router-dom'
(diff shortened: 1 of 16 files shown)


frontend/src/views/FinalSupplyEquipments/FinalSupplyEquipmentSummary.jsx
```diff
@@ -29,26 +28,37 @@ export const FinalSupplyEquipmentSummary = ({ data, status }) => {
     }
   }, [location.state])
 
-  const gridOptions = useMemo(() => ({
-    overlayNoRowsTemplate: t(
-      'finalSupplyEquipment:noFinalSupplyEquipmentsFound'
-    ),
-    autoSizeStrategy: {
-      type: 'fitCellContents',
-      defaultMinWidth: 50,
```
areyeslo, line 38: should not need this for small devices?

frontend/src/views/FuelSupplies/FuelSupplySummary.jsx
```diff
@@ -35,7 +34,6 @@ export const FuelSupplySummary = ({ data, status }) => {
       overlayNoRowsTemplate: t('fuelSupply:noFuelSuppliesFound'),
       autoSizeStrategy: {
         type: 'fitCellContents',
-        defaultMinWidth: 50,
```
areyeslo, line 38: Why removing min width?

areyeslo:
Looks great! Thanks for cleaning up some code.

frontend/src/views/FuelSupplies/FuelSupplySummary.jsx
dhaselhan: I think we should either set the min width in the default column or in the autosize strategy but no both. I chose to remove it from here and leave it set on the defaultColumn.

frontend/src/views/FinalSupplyEquipments/FinalSupplyEquipmentSummary.jsx
dhaselhan: I moved the minWidth to the default column

areyeslo: Thanks for the clarification. Considering column properties such as minWidth, maxWidth, filter, sortable are better defined in defaultColdDef. Should we also move the defaultMaxWidth?

dhaselhan:
This is now just a smaller version of my other MR, no longer needed.

PR #1561, opened 2024-12-23
LCFS-1509:  Fuels for other use and Export fuels not saving

areyeslo:
The issue affecting the row to be saved was fixed with this [commit](https://github.com/bcgov/lcfs/commit/f6a18d6b67bde24f06790301ce21a0a70cd55a7c#diff-583c50cd62ebddce486e72458583d5abc2181be5c7acc5ec3fcc30423105e859). The root cause was the use of a non-existent function name,  `get_fuel_category_by_name`.

Howerver, other issues were affecting the functionality of Other Use and Export fuels:
- When selecting fuel type and there is only one fuel category, then preselect the value. Otherwise, show the values in the dropdown to allow selection.
- When updating the fuel type, the user needs to select the option for Determining Carbon intensity
- When modifying existing rows, quantity field should not be reset to 0


AlexZorkin (approved):
Looks good 👍

PR #1565, opened 2024-12-30
feat: move common seeders to migration file

### LINKED ISSUE #1503: Migrate Key Seed Data to SQL Migration
As the administrator for LCFS, I want to be confident that the core seed data (fuel types, fuel categories, ....) is consistent and maintainable across all environments.

**Purpose**
This task converts the existing JSON seed files to a SQL Migration. This means the data will be set in stone at that point in time and all future changes will also need migrations. This fixes the issue where once seed_data is ran on an environment, it cannot be easily changed.

The goal here is to move all the common seed data into a SQL migration so it can be reliably updated via SQL.

**Acceptance Criteria**
- common_seeder.py is deleted along with the json seed data
- All of this is converted to SQL statements and added to a new migration
- Database has the same seed data as before this change
- Currently end use other options is NOT in common but should be moved into the SQL and removed from the dev/test/prod seeders as it does not need to change per environment.

kevin-hashimoto:
- feat: move all common seeders to a new migration file, closes #1503 
- chore: delete common seeder files


dhaselhan (changes requested):
The common changes look good but I think this is missing expected_use_type.

dhaselhan (approved):
Nice work Kevin! LGTM

PR #1566, opened 2024-12-30
Fix: Create Fuel Supply Unit Calculation - 1549

AlexZorkin:
- Added missing compliancePeriod value to create new row method
- Added missing test files for fuel supply


dhaselhan (approved):
LGTM 👍

PR #1567, opened 2024-12-30
Fix: Transfer Counts - 1518

AlexZorkin:
- Added missing materialized view on organization transfer counts
- Added update query to update all existing transfer counts


dhaselhan (approved):
LGTM 👍

PR #1568, opened 2024-12-30
Fix: Marine End Use - 1498

AlexZorkin:
- Added back marine end use and eer


kevin-hashimoto (commented):
this will need to be added into the common seeders migration ticket: https://github.com/bcgov/lcfs/pull/1565

prv-proton (approved):
Looks good

AlexZorkin:
Need to rework this to work with our migration seeder now

PR #1570, opened 2024-12-31
Feat: Transaction Unregistered Org List - 1525

AlexZorkin:
- updated org names endpoint to accept only_registered param


prv-proton (approved):
Looks good

PR #1578, opened 2025-01-02
Fix: LCFS - BUG - year selection when creating a new compliance report #1577

### LINKED ISSUE #1577: LCFS - BUG - year selection when creating a new compliance report
**Describe the Bug**
When creating a new compliance report, the dropdown for year selection no longer includes 2024 now that we're in 2025. This creates an issue for users who need to create reports for 2024.

Image

**Expected Behaviour**
The dropdown for year selection should include 2024 as an option, allowing users to create compliance reports for 2024, regardless of calendar year. 2024 should only disappear if they have an existing 2024 report in any status.

**Actual Behaviour**
The year 2024 is missing from the dropdown, preventing users from selecting it and creating compliance reports for that year.

**Implications**
- Users are unable to create compliance reports for the year 2024.
- This could lead to delays in reporting and compliance issues.
- User frustration and system usability concerns.

**Steps To Reproduce**
User/Role:
1. Go to the compliance report creation page.
2. Open the dropdown for selecting the year of the report.
3. Observe that 2024 is not an available option.

**Additional Notes**
- Verify if the logic that populates the year dropdown dynamically removes previous years once the calendar year advances.
- Ensure that compliance rules or reporting policies do not restrict the creation of past-year reports unnecessarily.
- Update automated tests for year selection logic to prevent regression.

prv-proton:
Fix #1577


dhaselhan (approved):
LGTM 👍

PR #1587, opened 2025-01-02
fix: compliance report fuel supply end use logic

kevin-hashimoto:
- fix: end use validation not triggering when empty value set
- feat: auto select end use when it is the only option

note: will need to add the new seed values to migration in #1503  pr: #1565


frontend/src/views/FuelSupplies/AddEditFuelSupplies.jsx
```diff
@@ -161,11 +161,39 @@ export const AddEditFuelSupplies = () => {
...
             fuelCategoryOptions.length === 1 ? fuelCategoryOptions[0] : null
+          const endUseValue =
+            endUseTypes.length === 1 ? endUseTypes[0].type : null
 
           params.node.setDataValue('fuelCategory', categoryValue)
+          params.node.setDataValue('endUseType', endUseValue)
+        }
+      }
+
+      if (params.column.colId === 'fuelCategory') {
```
dhaselhan, line 179: Do you its cleaner for column specific logic to live in a value setter or value getter for that column rather than a big if else in handle cell changed? I believe we kinda have a mix of both.

dhaselhan (approved):
LGTM 👍

frontend/src/views/FuelSupplies/AddEditFuelSupplies.jsx
kevin-hashimoto: Yea I was debating this as well. The thing I don't like about ag grid is that they don't provide a clean validation schema that's separated from row/column/cell logic. I'm going to look to rework some of this logic in a refactor ticket

PR #1590, opened 2025-01-03
feat: Switch from handleRowClicked to defaultColumn

dhaselhan:
* Set linkCell Renderer and link props on default column and remove row click.
* Update cell renderers to all support links if present
* Apply to lots of tables
* Move test files to be closer to their components


frontend/src/views/ComplianceReports/ComplianceReports.jsx
```diff
@@ -66,42 +51,53 @@ export const ComplianceReports = () => {
...
 
   useEffect(() => {
     if (isCreating) {
       setIsButtonLoading(true)
     }
   }, [isCreating])
+
+  const defaultColDef = useMemo(
+    () => ({
+      floatingFilter: false,
```
prv-proton, line 90: should we be disabling floating filter here?

prv-proton (changes requested):
Everything looks good, just for compliance reporting floating filter is not visible.

frontend/src/views/ComplianceReports/ComplianceReports.jsx
dhaselhan: Good catch, definitely not. Removed and double checked the other changes to make sure I didnt accidentally turn them off elsewhere.

prv-proton (approved):
looks good.

PR #1603, opened 2025-01-03
fix: Fix Status Filters for Compliance Reports

dhaselhan:
* Remove conditional logic so we always join statuses, we cannot filter on it unless we join it
* Fix mapping to handle multiple values from Submitted status
* Rename ReAssessed to Reassesed to match the front end


AlexZorkin (approved):
👍

PR #1604, opened 2025-01-04
LCFS-1553: BUG - Other diesel

areyeslo:
Changes:

- Rename `other diesel` for `other diesel fuel` in fuel types for all the schedules.
- Allocation agreement is showing `Prescribed carbon intensity - section 19 (a)` which is related to fossil derived. Change to  `Default carbon intensity - section 19 (b) (ii)` as the other schedules.

image


hamed-valiollahi (approved):
Looks good to me! 👍

PR #1605, opened 2025-01-06
Feat: Refine Report History Display - 1535

### LINKED ISSUE #1535: LCFS - Updates to Report History
**Description:**  
Refine the Report History section to display only relevant and meaningful information. Avoid showing empty or unnecessary history entries.  

**Requirements:**  
1. **Conditional Display:** Only show the Report History section if there are entries to display.  
2. **Filter History Entries:** Only include reports in the history chain if they contain meaningful data.  
3. **Draft State Handling:** Do not display the history block for reports in a "Draft" state.  

**Acceptance Criteria:**  
- Report History section is hidden when there are no relevant entries.  
- Only reports with meaningful history are displayed in the history chain.  
- History block is not visible for reports in "Draft" state.  
- Automated tests cover the updated behavior.

hamed-valiollahi:
This PR refines the report history display logic to ensure only meaningful entries are shown and excludes drafts.

Closes #1535


frontend/src/views/ComplianceReports/components/AssessmentCard.jsx
```diff
@@ -129,19 +138,19 @@ export const AssessmentCard = ({
                   <StyledListItem>
                     <ListItemText primaryTypographyProps={{ variant: 'body4' }}>
                       <span
```
dhaselhan, line 140: This is an odd formatting change as its making the end tag not line up with the opening tag, was this from Prettier?

frontend/src/views/ComplianceReports/components/AssessmentCard.jsx
```diff
@@ -23,7 +24,7 @@ export const AssessmentCard = ({
   currentStatus,
   complianceReportId,
   alertRef,
-  chain
+  chain = []
```
dhaselhan, line 27: Why are we updating this parameter to be optional?

frontend/src/views/ComplianceReports/components/AssessmentCard.jsx
```diff
@@ -52,6 +53,14 @@ export const AssessmentCard = ({
       }
     })
 
+  const filteredChain = useMemo(() => {
```
dhaselhan, line 56: If we never want to show Draft report history, would it be more correct to delete it from ever being created?

dhaselhan (changes requested):

frontend/src/views/ComplianceReports/components/AssessmentCard.jsx
hamed-valiollahi: Yes, that's weird, as it's fine on my VS Code.

image

frontend/src/views/ComplianceReports/components/AssessmentCard.jsx
hamed-valiollahi: Thanks, Daniel! I applied the change. The code is now cleaner!

dhaselhan (approved):
LGTM 👍

PR #1608, opened 2025-01-06
feat: Add Legacy Provisions

dhaselhan:
* Add is_legacy to provisons and follow similar logic to fuel types
* Copy over legacy provisions from TFRS -> LCFS


kevin-hashimoto (approved):
LGTM

PR #1609, opened 2025-01-06
fix: Check dateFrom instead of filter

dhaselhan:
* Date filter is not settings its initial value on load
* Check dateFrom the field we use to see if its valid instead of one we don't


AlexZorkin (approved):
👍

PR #1614, opened 2025-01-07
fix: 403 error loading compliance report

kevin-hashimoto:
quick fix for 403 errors that are triggered by current user data not loaded in.

should eventually migrate the deprecated BCDataGridServer with the newer BCGridViewer component. Should also addres the TODOs in the BCGridViewer component


AlexZorkin (approved):
Nice 👍

PR #1622, opened 2025-01-07
Add Legacy Fuels

dhaselhan:
* Migrate over legacy Fuel Types from TFRS and set is_legacy flag.


AlexZorkin (changes requested):
This is missing all the other legacy fuel types. We need to add all fuel types from TFRS even if they are duplicates because we need to associate the legacy energy density, eer values. They will be differentiated by the is_legacy flag.

AlexZorkin (approved):
👍

PR #1627, opened 2025-01-08
Fix: LCFS BUG - Navigating directly to organization detail errors #1586

### LINKED ISSUE #1586: LCFS BUG - Navigating directly to organization detail errors
Log in as IDIR user
Navigate to organizations tab
Select an organization
Refresh the page.

Expected: Organization detail page is shown
Actual: Error!  prevDeps.join is not a function

This causes an issue if the user opens the organization detail in a new tab.

prv-proton:
- Fix #1586.
- Minor typo corrections and aria-labelling.
- Fix Accessible header component, as it is impacting the sorting functionality within ag-grid.


AlexZorkin (approved):
👍

PR #1628, opened 2025-01-08
Fix: Add Marine End Use - 1498

AlexZorkin:
- migration to add marine end use and eer


backend/lcfs/db/migrations/versions/2025-01-07-19-35_998929392c8b.py
```diff
@@ -0,0 +1,51 @@
...
+Create Date: 2025-01-07 19:35:00.064999
+
+"""
+
+import sqlalchemy as sa
+from alembic import op
+
+# revision identifiers, used by Alembic.
+revision = "998929392c8b"
+down_revision = "94306eca5261"
```
prv-proton, line 14: may need fix for this.

prv-proton (changes requested):
Revision head fix is required.

hamed-valiollahi (approved):
Looks good to me! 👍
Please update the `down_revision` ID before merging.

prv-proton (approved):
👍

PR #1632, opened 2025-01-08
feat: Dynamically change grid from autoHeight -> Normal

dhaselhan:
Add a new autoHeight feature to BCGridBase that auto changes the dom rendering methods when the number of rows reaches the size of the screen.


AlexZorkin (commented):
If the zero rows case is covered then this is good to go.

prv-proton (approved):
Looking very good, thanks for the implementation.

PR #1646, opened 2025-01-09
fix: Remove Active filter button and Remove User ID Column

dhaselhan:


AlexZorkin (approved):
feels good to delete this extra code 👍

Issue #1625, opened 2025-01-08
LCFS - BUG - Fuel supply label


airinggov:
**Describe the Bug**

![Image](https://github.com/user-attachments/assets/78191861-af9c-45ab-8b23-28cd33afeda7)

**Expected Behaviour**

**Actual Behaviour**
A clear and concise description of what actually happens.

**Implications**
A clear and concise description of any implications.

**Steps To Reproduce**
Steps to reproduce the behaviour:
User/Role:

Grulin:
Still showing the incorrect label:

![Image](https://github.com/user-attachments/assets/5f9bf579-0594-4cdb-9768-945a7faae166)

Issue #1593, opened 2025-01-03
LCFS - Update CR schedule text


airinggov:
**Task:**  
Update the page titles, breadcrumb and instruction text at the top of the following pages.

**Fuel supply page**
Retitle page and breadcrumb to:

Supply of fuel

Update text to:

Report fuel volumes you supplied for the various end uses.

**FSE page** 
Retitle page and breadcrumb to:

Final supply equipment

Update text to:

Report dates of supply for your final supply equipment (FSE).

Organization: Enter your organization name unless you are reporting FSE that you have received allocated responsibility. In this case, enter the utility account holder's organization name.

Enter the billing address in the Notes field if it is different than the equipment address. Also, enter an explanation in the Notes field if you select 'Other' from any of the select options.

**Allocation agreement page**
Retitle page and breadcrumb to:

Allocation agreements

Update text to:

Report fuel volumes for which you have an allocation agreement. Report both fuels that you have allocated responsibility to another organization, and fuels you have received responsibly from another organization.

Fuels that you have received responsibility for must also be reported in fuel supply.

**Notional transfer page**
Retitle page and breadcrumb to:

Notional transfer of eligible renewable fuels

Add text:

Report renewable fuel volumes that you received or transferred through a notional transfer.

**Fuels for other use page**
Retitle page and breadcrumb to:

Fuels for other use

Update text to:

Report fuel volumes that you supplied for heating or other prescribed uses other than transportation.

**Export fuels page**
Retitle page and breadcrumb to:

Export fuels

Update text to:

Report fuel volumes that were marketed or supplied in B.C. and then exported out of the province. Only negative compliance units (debits) can be issued, credits cannot be issued for exported fuel.

Grulin:
Additional reference: 

BC Gov writing for web content notes that headings and page titles should use 'Sentence case'

This means in headings and page titles you should only capitalize:

The first word, if it starts with a number such as '2020 School tax rates', capitalize the first word after the number
[Proper nouns](https://www2.gov.bc.ca/gov/content/governments/services-for-government/service-experience-digital-delivery/web-content-development-guides/web-style-guide/writing-guide/capitalization)
[Abbreviations and acronyms](https://www2.gov.bc.ca/gov/content/governments/services-for-government/service-experience-digital-delivery/web-content-development-guides/web-style-guide/writing-guide/abbreviations) if they have already been defined or are better known to the audience than their long-form equivalent, such as PDF or DNA
[...]

PR #1650, opened 2025-01-09
Feat: LCFS - Update CR schedule text

prv-proton:
#1593 update compliance report schedules text.
#1625


AlexZorkin (approved):
I'm assuming the text changes are all correct. Looks good 👍

PR #1653, opened 2025-01-09
fix: Lock Summary on Submit

dhaselhan:
* Copy over is_locked flag on update method to summary
* Remove old unused lock method
* Remove before_update trigger as we reports can be unlocked now


backend/lcfs/web/api/compliance_report/repo.py
```diff
@@ -601,6 +601,7 @@ async def save_compliance_report_summary(
                 f"No summary found with report ID {summary.compliance_report_id}"
             )
 
+        summary_obj.is_locked = summary.is_locked
```
prv-proton, line 604: Since we've removed in listeners, shouldn't we change the query in update_compliance_report to respect is_locked?

dhaselhan: Originally I updated the listener to only allow saving when not locked and to only allow locking with no other updates, but going forward we need the ability to move compliance reports in and out of draft / locked so Alex said to just delete it. Since we need to be moving reports in and out of locked, I don't think we want a check in update. Going forwards lock is just locking the calculation of the summary.

prv-proton (approved):
looks good 👍

PR #1670, opened 2025-01-11
Lcfs 1539: Add Clear Filters Button to all Tables

areyeslo:
Add `Clear filters` button to all tables

image

[Story](https://github.com/orgs/bcgov/projects/137/views/1?filterQuery=arey&pane=issue&itemId=91661203&issue=bcgov%7Clcfs%7C1539)


frontend/src/views/Admin/AdminMenu/components/AuditLog.jsx
```diff
@@ -33,12 +35,26 @@ export const AuditLog = () => {
     []
   )
 
+  const handleSetResetGrid = useCallback((fn) => {
+    setResetGridFn(() => fn)
+  }, [])
```
dhaselhan, line 40: Do you need the most inner function? I think this can just be setResetGridFn(fn)

dhaselhan (changes requested):

frontend/src/views/Admin/AdminMenu/components/AuditLog.jsx
areyeslo: Yes, I need the inner function, it allows to have an stable function reference:
image

If I update it to just include `fn` instead of the arrow function, then I am loosing the function reference:
image

When clicking the button, the function is not executed.

dhaselhan (approved):
LGTM 👍

PR #1673, opened 2025-01-12
feat: Enable Cypress Daily

dhaselhan:
* Remove duplicate script for running on PR
* Update script with rabbitmq and a cron schedule


.github/workflows/cypress.yml
```diff
@@ -0,0 +1,135 @@
...
+          --health-timeout=5s
+          --health-retries=5
+        ports:
+          - 5432:5432
+    steps:
+      - uses: actions/checkout@v2
+      - name: Set up Python
+        uses: actions/setup-python@v2
+        with:
+          python-version: '3.9'
```
prv-proton, line 73: have we not updated to use 3.11 version?

dhaselhan: Good catch, I uncommented the old file but should have done a more thorough look. Removed all the extra checks that are not cypress related.

prv-proton (approved):
Looks good

PR #1674, opened 2025-01-12
feat: Fix Bad Transaction Times

dhaselhan:
* TFRS has a bug where the timezone is not considered when setting the transaction effective date, meaning a good chunk of them are one day off.
* Write a clean up script to manually correct these records in LCFS after they are ETLd


areyeslo (approved):
LGTM

PR #1675, opened 2025-01-12
feat: Check and refresh token on each click

dhaselhan:
* Add a click listener that checks the current tokens validity time
* If less than 60 run a refresh


frontend/src/components/KeycloakProvider.jsx
```diff
@@ -20,6 +20,14 @@ export const KeycloakProvider = ({ authClient, children }) => {
     )
   }
 
+  useEffect(() => {
```
prv-proton, line 23: instead of having a click listener, isn't it a good idea to intercept the backend api calls to refresh the tokens or redirect them to login page? especially something like useNotificationsCount which gets called quite frequently?

dhaselhan: I don't think so, most of those are called far too often and are not a direct result of a user interaction. I had originally put this on API service but realized its all just more complex ways for finding a user interacting. I went with the simplest approach in the simplest place.

prv-proton (approved):
Thanks for the implementation and clarifications, looks good 👍. Please remove the commented lines if not required.

PR #1676, opened 2025-01-13
fix: Set max width and re-use plain style

dhaselhan:
* Math was setting max width above XLS allowed, cap at the maximum width
* You can only have 4096 styles, so re-use the plain style instead of re-creating it for each cell
* Upgrade Redis to latest as my local was crashing


hamed-valiollahi (approved):
Looks good to me! 👍

PR #1680, opened 2025-01-14
feat: More Cypress Fixes

dhaselhan:
Add missing installs, waiting, commands to install cypress etc

Runs correctly only thing missing is configs


AlexZorkin (approved):
👍

PR #1685, opened 2025-01-14
Fix: Separate where clauses for compliance report

dhaselhan:
* We need some filters to be applied to the inner select and some to the outer
* Create two filter functions for separating the filters and apply appropriately
* Always join org and compliance period so we can filter/sort by it


AlexZorkin (approved):
Looks good

PR #1686, opened 2025-01-14
LCFS-1654: Protect routes for compliance report

areyeslo:
If the user has Government role and the compliance report status is "Draft" restrict access to the compliance report and its schedules to ensure they remain protected. The user is returned to /compliance-reporting

/compliance-reporting/:compliancePeriod/:complianceReportId/
/compliance-reporting/:compliancePeriod/:complianceReportId/supply-of-fuel
/compliance-reporting/:compliancePeriod/:complianceReportId/final-supply-equipments
/compliance-reporting/:compliancePeriod/:complianceReportId/allocation-agreements
/compliance-reporting/:compliancePeriod/:complianceReportId/notional-transfers
/compliance-reporting/:compliancePeriod/:complianceReportId/fuels-other-use
/compliance-reporting/:compliancePeriod/:complianceReportId/fuel-exports

[Story](https://github.com/orgs/bcgov/projects/137/views/1?pane=issue&itemId=93208433&issue=bcgov%7Clcfs%7C1654)


AlexZorkin (changes requested):
Although your solution is workable, this adds an extra api call every time we open one of these views. A more efficient solution would be to use the existing validation files on the backend and check user permissions there. validation.py already uses compliance record so we have status available. Let me know what you think.

AlexZorkin (approved):
Looks good. Thank you for moving to a backend solution. 👍

PR #1688, opened 2025-01-14
fix: Rename Labels and Fix User Name Search

dhaselhan:
* Use concatenated full name column with a single ilike to fix issues and reduce complexity
* Rename CI of fuel to RCI


AlexZorkin (approved):
Much simpler query 👍

PR #1692, opened 2025-01-14
Fix: ETL Buy Transfer

AlexZorkin:
- etl logic for buy transfer type
- reverted display logic for org balances mv


prv-proton (approved):
Looks good 👍

Issue #1649, opened 2025-01-09
LCFS - BUG- Organization status in view table in PROD ETL mapped "Unregistered" in TFRS to "Suspended" in the LCFS Portal


Grulin:
**Describe the Bug**
The PROD ETL mapped "Unregistered" in TFRS on the organization view table to "Suspended" in the LCFS Portal. 

**Expected Behaviour**
The view table should be showing "Unregistered" organization in TFRS with a status of "Unrecognized" in the LCFS Portal.

**Actual Behaviour**
The selection of "registered" or "unregistered" on the Organization details page came across correct, but the label in the view table is incorrectly showing examples of "Suspended" statuses and "Unrecognized" statuses. 

**Implications**
The implication is that we may have some users within the system that are labelled "Suspended" but in fact only have not registered to transfer or have been moved to "Unregistered" for transfer after being inactive for too long. 

**Steps To Reproduce**
Steps to reproduce the behaviour:
User/Role: IDIR
Navigate to the organization view table
Observe "Registered" status organizations and "Suspended" Organizations in the view table

PR #1704, opened 2025-01-15
Organiation status fix for archived orgs

prv-proton:
#1649 
After discussing with Lindsy and Al, we'll be classifying existing organizations to either registered and unregistered statuses in LCFS portal.


AlexZorkin (approved):
👍

PR #1708, opened 2025-01-15
fix: Update Cypress Tests

dhaselhan:
* Update some selectors and IDs to fix some bugs
* Remove contact-us test as we deleted that page
* Remove disableHover as it wasnt a real prop


hamed-valiollahi (approved):
Looks good to me! 👍

Issue #1585, opened 2025-01-02
LCFS - Implement Accessibility & Functional Enhancements for AG Grid Filtering Components


prv-proton:
**Describe the task**
Enhance the AG Grid filtering components across the LCFS application to ensure compliance with accessibility standards, improved functionality, and user-friendly date filtering.

**Scope**
1. Compliance Reports:

- Dropdowns: Enable navigation between dropdown elements using arrow keys.
- Date Picker: Ensure arrow keys function correctly within the date picker and restrict date input to four characters.
- Date Range: Add functionality to set maximum and minimum date ranges to limit user selection.

2. Fuel Codes:

- Date Picker: Resolve issues where keyboard keys become non-functional within the date picker.
- Date Range: Enforce a four-character date input limit and set maximum and minimum date ranges.

3. Organizations:
- Dropdowns: Implement proper keyboard navigation through dropdowns using arrow keys.

4. Cross-Grid Filter Check:
- Add functionality to verify and fix filters across all grids in the LCFS application.

5.  Accessibility Standards (WCAG):
- Implement screen reader support for dropdowns and date pickers to meet WCAG standards.
- Ensure proper ARIA roles, labels, and live region announcements for grid components.

**Acceptance Criteria**
- [ ] All dropdown menus across grids are navigable using keyboard arrow keys.
- [ ] Date pickers allow navigation and selection using arrow keys, enforce a four-character date input limit, and have defined max/min date ranges.
- [ ] Filters across all grids are checked and corrected for consistent functionality.
- [ ] Screen readers accurately interpret and announce dropdown and date picker elements per WCAG standards.
- [ ] Accessibility improvements are tested and verified in the Compliance Reports, Fuel Codes, and Organizations sections.

**Additional context**
-

Grulin:
Not all arrows are working to navigate through the grids.

StuartGALL:
Please add functionality for the enter key.

Grulin:
Enter key functionality is working now, but still having issues with the down arrow. For example, I can navigate to "Fuel type" in the compliance report AG grid for fuel supply, press enter on this drop down and successful choose an option only by using the up arrow. I am unable to move through the list by pressing the down arrow.

PR #1715, opened 2025-01-16
fix: Adding enter key functionality to select editor in ag-grid.

prv-proton:
Fix for selector component in ag-grid:
This fix adds the functionality to open/close the drop down using Enter key
https://github.com/bcgov/lcfs/issues/1585#issuecomment-2588101469


dhaselhan (approved):
LGTM 👍

PR #1718, opened 2025-01-16
fix: Include approved and assessed transactions for Orgs

dhaselhan:
* Compliance Report Transactions are assessed status when finalized, not approved
* Add them to the where clause used for Orgs


kevin-hashimoto (approved):
LGTM

PR #1731, opened 2025-01-17
fix: Change Transaction history to full timestamp

dhaselhan:
* Use custom fn as dayjs has no way to display timezones
* Try and remove moment


AlexZorkin (approved):
yay for dayjs!

PR #1742, opened 2025-01-18
Fix: Org Pagination - 1721

AlexZorkin:
- fixed re-rendering issue on organization list view 
- small etl validation fix


frontend/src/views/Organizations/Organizations.jsx
```diff
@@ -21,98 +21,110 @@ import { LinkRenderer } from '@/utils/grid/cellRenderers.jsx'
...
-    setResetGridFn(() => fn)
-  }, [])
-
-  const handleClearFilters = useCallback(() => {
-    if (resetGridFn) {
-      resetGridFn()
-    }
-  }, [resetGridFn])
+  // Row ID
+  const getRowId = useCallback((params) => params.data.organizationId, [])
```
dhaselhan, line 114: getRowId should return a string to prevent warnings from ag-grid. Is this throwing warnings?

dhaselhan (changes requested):

frontend/src/views/Organizations/Organizations.jsx
AlexZorkin: no warnings shown but I have hard casted this just in case.

dhaselhan (approved):
LGTM 👍

PR #1658, opened 2025-01-10
LCFS-1659: Incorrect fuel type list in Fuels for other use

areyeslo:
This fix is focused on getting the correct list in fuel_type. The data might be changed in Dev and affected the list shown in the dropdown.

In addition, added clearing provisionOfTheAct when changing fuel_type.

image

_Note: Renewable naphtha is not included because its fuel code is expired._

[Story](https://github.com/orgs/bcgov/projects/137/views/1?pane=issue&itemId=93214072&issue=bcgov%7Clcfs%7C1659)

PR #1743, opened 2025-01-18
LCFS-1659: Incorrect fuel type list in Fuels for other use

areyeslo:
A previous [PR](https://github.com/bcgov/lcfs/pull/1658) was created to show the requested fuel types. However, _Renewable naphtha_ is not included because its fuel code is expired.

This PR includes fuel types such as  _Renewable naphtha_ with its `Default carbon intensity`:
image


AlexZorkin (changes requested):
Looks good, can we add a test or two to the test_other_uses_repo.py file just to cover this new logic.

AlexZorkin (approved):
Thanks for the pytests 👍

PR #1748, opened 2025-01-20
feat: Add Organization Address Snapshot

dhaselhan:
* Snapshot org details when creating compliance report and store in new table
* Allow editing of the snapshots
* New UI for editing


AlexZorkin (approved):
This is very clean, nice work 👍

PR #1750, opened 2025-01-21
feat: Add Env to Front End Config

dhaselhan:
* Use to control red toasts


AlexZorkin (approved):
Looks good, just need to make sure we update the config map in prod. 👍

PR #1759, opened 2025-01-21
LCFS-1385: Only ETL Users with Valid Emails

areyeslo:
Modified user.groovy to ensure that the email field is always assigned a valid value.

[Story](https://github.com/orgs/bcgov/projects/137/views/1?filterQuery=arey&pane=issue&itemId=90026159&issue=bcgov%7Clcfs%7C1385)


AlexZorkin (approved):
👍

PR #1761, opened 2025-01-21
feat: New Shiny Yellow Errors!

dhaselhan:
* Standardize row saving and error handling model across all schedules
* Add new state for storing warnings
* Update schema to use new handler that supports warnings
* Cleanup and other fixes


prv-proton (approved):
looks good 👍

PR #1764, opened 2025-01-21
LCFS-1746: Pagination through Administration > Users

areyeslo:
Fixed an issue where the `Users` component was re-rendering multiple times, negatively impacting pagination performance. The root cause was that `defaultSortModel` variable was updating its reference in memory on each render.

Applied memoization to prevent unnecessary updates and re-renders, ensuring stable pagination behavior:
image

[Story](https://github.com/orgs/bcgov/projects/137/views/1?filterQuery=arey&pane=issue&itemId=94477691&issue=bcgov%7Clcfs%7C1746)


dhaselhan (approved):
LGTM 👍

PR #1767, opened 2025-01-22
fix: Move renewable flag to DB

dhaselhan:
* Fuel Type name changed breaking current code
* Move to DB so its less fragile


AlexZorkin (approved):
Much simpler this way. 👍

PR #1768, opened 2025-01-22
Fix: Fix Transaction From and Sorting by Report Type

dhaselhan:
* Report type needs to be mapped to nickname
* Move the current user check above the callbacks to prevent them from being conditional
* Add missing .data


hamed-valiollahi (approved):
Looks good to me! 👍

PR #1778, opened 2025-01-22
feat: Hide Compliance Reporting without Required Roles

dhaselhan:
* Previously supplier role was the only one required, now we explicitly check for ComplianceReporting or SigningAuthority.
* Add new ConditionalLinkRenderer to handle tables that have some links
* Update roles on back end endpoints
* Hive NavBar item
* Add trigger to refresh Transaction Aggregate for CRs


areyeslo (approved):
LGTM

PR #1779, opened 2025-01-22
fix: Return transaction instead of applying

dhaselhan:
* BCGridBase now takes a Transaction returned from onAction and will apply it if it contains rows.
* This was changed so we can focus on the newly added cell, and it looks like this table was not updated


AlexZorkin (approved):
👍

PR #1783, opened 2025-01-23
Fix: LCFS - ETL BUG- PROD Edited label being attached to all analyst comments #1755

### LINKED ISSUE #1755: LCFS - BUG- PROD Edited label being attached to all analyst comments
**Describe the Bug**
In the LCFS Portal PROD environment, the "Edited" label is being incorrectly stamped on all transaction comments, regardless of whether the comment was actually edited. Comments that were never edited still display the "Edited" label, often with identical date and time stamps as the original comment creation.

**Expected Behaviour**
- The "Edited" label should only appear on comments that have been modified after their initial submission.
- Comments that have not been edited should not display the "Edited" label.
- If a comment is edited, the "Edited" label should reflect the updated timestamp accurately.

**Actual Behaviour**
- The "Edited" label is stamped on all comments, including those that were never edited.
- Comments have identical date and time stamps for creation and the supposed "edit."

Image

**Implications**
- This creates confusion for users, as it falsely suggests comments were edited when they were not.
- Users may lose confidence in the accuracy of comment histories.

**Steps To Reproduce**
User/Role: IDIR

1. Navigate to a transaction with comments in the LCFS Portal PROD environment.
2. Observe that the "Edited" label is applied to all comments from analysts

prv-proton:
Fix #1755


AlexZorkin (commented):
👍

AlexZorkin (approved):

PR #1785, opened 2025-01-23
fix: Update Summary Calculations to not exclude 11 and 21

dhaselhan:
* Refactor function into smaller pieces for easier readability / understandability
* Add new logic for skipping the duplicated 11 columns
* Convert line from str to int *


AlexZorkin (approved):
Nce job breaking this down into parts. 👍

PR #1787, opened 2025-01-24
fix: More Cypress Updates

dhaselhan:
* Use uppercase consistently across env vars
* Remove unused env vars
* Fix some tests
* Map in all env vars from github actions


AlexZorkin (approved):
👍

PR #1807, opened 2025-01-27
fix: default filter on filter reset

### LINKED ISSUE #1775: LCFS - Clearing filters results in BCeID users
Currently when navigating to the Users page as an IDIR user you will only see IDIR users.

* Click the Clear filter button
* Expected: Nothing changes
* Actual: BCeID users being to show

kevin-hashimoto:
- fix: filter reset clearing defaults, closes #1775

### DIFF
--- a/frontend/src/components/BCDataGrid/BCDataGridServer.jsx
+++ b/frontend/src/components/BCDataGrid/BCDataGridServer.jsx
@@ -189,7 +189,7 @@ const BCDataGridServer = ({
     setPage(1)
     setSize(paginationPageSize)
     setSortModel(defaultSortModel || [])
-    setFilterModel(defaultFilterModel || [])
+    setFilterModel([])
 
     // Clear UI filters
     if (gridRef.current?.api) {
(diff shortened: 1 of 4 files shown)


frontend/src/components/BCDataGrid/BCDataGridServer.jsx
```diff
@@ -189,7 +189,7 @@ const BCDataGridServer = ({
     setPage(1)
     setSize(paginationPageSize)
     setSortModel(defaultSortModel || [])
-    setFilterModel([])
+    setFilterModel(defaultFilterModel || [])
```
prv-proton, line 192: Would it be a good idea to move this logic to backend instead of frontend?

prv-proton (commented):
Wouldn't it be good idea to move this logic to backend instead of frontend?

frontend/src/components/BCDataGrid/BCDataGridServer.jsx
kevin-hashimoto: i suppose you're right as this user endpoint only serves this one purpose. ill move the defaults to the backend

AlexZorkin (approved):
👍

PR #1810, opened 2025-01-27
fix: Hide AG Grid Menus outside Viewport

dhaselhan:
* When scroll ends check to see if selected cell is outside viewport, if it is clear its selection / close the select.


kevin-hashimoto (approved):
working great! LGTM

PR #1815, opened 2025-01-27
fix: cypress accessibility issues

### LINKED ISSUE #1709: LCFS - Fix A11y issues
Part of our Cypress tests is running an a11y check to make sure we are following usability standards. 
We currently have 6 violations that need to be investigated / fixed:

Image

Requirements:
* Aria issues are fixed
* Accessibility cypress test passes

kevin-hashimoto:
closes #1709


dhaselhan (approved):
Nice work 👍

PR #1817, opened 2025-01-27
fix: Separate hasMet into two variables

dhaselhan:
* They can be met/not met independently, so use two variables
* Fix logic to check specific lines instead of the totals, as reserve credits do not count towards met / not met


AlexZorkin (approved):
Nice 👍

PR #1818, opened 2025-01-28
fix: Add custom CSS for Notification Button

dhaselhan:
* Since its a button not just a link, it has extra padding causing it to move on hover / active
* Use custom CSS override for this button to fix the issues


AlexZorkin (approved):
👍

PR #1820, opened 2025-01-28
Fix: Updated Materialized Views, Functions, Triggers - 1813

AlexZorkin:
- Updated migration that handles materialized views, functions, and trigger creation to better match what our live environment contains.
- Manually updated all environments to match migration to avoid re-running this older migration. It keeps local developments environments in line with deployed environments.
- Added validation queries for all of our db level mv, functions, and triggers to make sure all custom sql is deployed.


prv-proton (approved):
Looks good 👍

PR #1830, opened 2025-01-28
LCFS-1801: 500 error in Compliance report summary in TEST

areyeslo:
When an organization has compliance reports from previous periods that were copied from TRFS, these reports do not have a corresponding record in compliance_report_summary. To calculate the compliance report summary, the logic searches for previously retained and obligation values. For compliance reports originating from TRFS, these values are derived from the current compliance report using line 7 and 9.

With this change, compliance reports that have previous reports copied from TRFS are now loaded.


areyeslo:
Closing PR, compliance_summary.groovy was executed.

PR #1831, opened 2025-01-28
feat: Standardize Deletion Logic

dhaselhan:
* Move all schedule logic to a standard function
* Add logic to standard function to re-create a default row when the last row is deleted


areyeslo (approved):
LGTM

PR #1838, opened 2025-01-29
fix: Various Bug Fixes

dhaselhan:
* Skip ratio column in summary so we only have a single line 21
* Remove dead help links from Cards
* Give buttons a bit more room on transactions page to prevent them from squishing


AlexZorkin (approved):
👍

PR #1840, opened 2025-01-29
LCFS-1827: Recorded transfer in BCeID transaction history is showing director user name

areyeslo:
User login as BCeID Organization does not show name of Director:
image

IDIR users can see the name of Director:
image


AlexZorkin (approved):
👍

Issue #1800, opened 2025-01-27
LCFS - BUG - Inconsistent Dropdown List for Approved Fuel Codes in Compliance Reporting Schedules


prv-proton:
**Describe the Bug**
The dropdown for approved fuel codes list in compliance reporting schedules is inconsistent across "Fuel Supply," "Export of Fuels," and "Allocation Agreements." In certain cases, it lists "Draft" fuel codes, and in other cases, it shows unrelated fuel codes for the selected fuel type.

**Expected Behaviour**
The dropdown should consistently display only the approved fuel codes relevant to the provided fuel type in all compliance reporting schedules, including "Fuel Supply," "Export of Fuels," and "Allocation Agreements."

**Actual Behaviour**
The dropdown displays:
- Draft fuel codes in some instances.
- Unrelated fuel codes for the selected fuel type in other instances.
This inconsistency occurs across different sections of compliance reporting schedules.

**Implications**

- Users might select incorrect or irrelevant fuel codes, leading to errors in compliance reporting.
- It can cause delays in submission and review processes due to inaccurate or invalid data.
- Potential regulatory non-compliance if the incorrect codes are reported.

**Steps To Reproduce**
Steps to reproduce the behaviour:
User/Role: BCeID/IDIR Compliance reporting user roles

1. Go to the "Compliance Reporting" section.
2. Navigate to "Fuel Supply," "Export of Fuels," or "Allocation Agreements."
3. Open the dropdown for fuel codes in the reporting schedule.
4. Observe that the dropdown inconsistently shows draft or unrelated fuel codes for the provided fuel type.

airinggov:
@prv-proton nice work

PR #1841, opened 2025-01-29
Fix: LCFS - BUG - Inconsistent Dropdown List for Approved Fuel Codes in Compliance Reporting Schedules

prv-proton:
Missed out fix for Other uses.
Fixes: #1800


AlexZorkin (approved):
👍

PR #1844, opened 2025-01-29
fix: Update met/not met logic

dhaselhan:
* It should be checking the summary penalty amount not the amounts in the report


prv-proton (approved):
Looks good 👍

PR #1857, opened 2025-01-29
fix: Move Fuel Code validation to the back end

dhaselhan:
* Fuel code was using custom logic meaning it didnt get any of the yellow upgrades done previously
* Remove the FE logic to Backend validation and let the errors go through the normal validation route


backend/lcfs/web/api/other_uses/views.py
```diff
@@ -123,7 +123,6 @@ async def get_other_uses_paginated(
 
 @router.post(
     "/save",
-    response_model=Union[OtherUsesSchema, DeleteOtherUsesResponseSchema],
```
dhaselhan, line 126: Union types have some funky behaviour in Pydantic, so in this case it was applying the new validator to the DeleteOtherUsesResponseSchema even though it wasnt defined on it.

AlexZorkin:
Will have to repoint this to the new release-1.0.0 branch

PR #1859, opened 2025-01-29
fix: Allow Submissions without enough credits

dhaselhan:
* Previously a supplier would be blocked if they didnt have enough credits
* Now we take as many as we can and allow it


AlexZorkin (changes requested):
Looks good, we should probably add some pytest coverage for this change.

dhaselhan:
> Looks good, we should probably add some pytest coverage for this change.

Added 3 new unit tests to cover enough credits, no credits, and not enough credits.

AlexZorkin (approved):
👍

PR #1862, opened 2025-01-30
fix: Check from organization balance

dhaselhan:
* Method currently checks whichever ID was passed in, always check the from_organization_id for the balance.
* Remove txn:description text as per business request


prv-proton (approved):
Looks good 👍

PR #1875, opened 2025-01-30
Filter instance highlight issue

prv-proton:
Fix: After moving back to grid page, the filter instance is not being highlighted.


AlexZorkin (approved):
👍

dhaselhan (changes requested):
Needs to be rebased to merge into release-1.0.0

PR #1879, opened 2025-01-31
fix: Move Fuel Code validation to the back end

dhaselhan:
* Fuel code was using custom logic meaning it didnt get any of the yellow upgrades done previously
* Remove the FE logic to Backend validation and let the errors go through the normal validation route


AlexZorkin (approved):
👍

PR #1881, opened 2025-01-31
fix: Allow Submissions without enough credits

dhaselhan:
* Previously a supplier would be blocked if they didnt have enough credits
* Now we take as many as we can and allow it


AlexZorkin (approved):
👍

PR #1884, opened 2025-01-31
LCFS-1556: Migrate Schedule C Records from TFRS to Other Uses in LCFS with Version Chaining

areyeslo:
- Migrate Schedule C to other_uses in LCFS
- Allow null values for provision_of_the_act_id and ci_of_fuel in other_uses table


AlexZorkin (approved):
Looks good ty 👍

PR #1897, opened 2025-01-31
fix: Don't fire refresh queries when changing compliance report status

dhaselhan:
* New code navigates to the main listing, so we dont need to refresh all the sub-data
* Firing the refresh in the case of return to supplier was resulting in a 403


AlexZorkin (approved):
Nice fix 👍

PR #1898, opened 2025-01-31
fix: 2 Bug Fixes

dhaselhan:
* Pass in request.user instead of trying to read it from self
* The FE for Compliance Report was updated to show nickname in the Type column but the filter was not updated


AlexZorkin (approved):
👍

PR #1900, opened 2025-01-31
fix: Cypress Env Vars

dhaselhan:
In order to be parsed by use in Cypress.env they need to start with CYPRESS_ which will then be trimmed off

Add flag to disable CHES
Disable CHES during tests


AlexZorkin (approved):
Looks good 👍

PR #1904, opened 2025-02-03
LCFS-1854: Rescinded transfers

areyeslo:
- For IDIR users, `Rescinded` transfers are shown as long as they were submitted before. This change in logic is included in get_transactions_paginated to avoid modifying the view.
- The logic for BCeID users is kept the same

**For IDIR users:**
image
In this scenario CT4 was `Submitted` before being `Rescinded`. IDIR users cannot see the other record because it was never submitted.

**For BCeID users:**
image
In this scenario CT3 was in `Draft` and then `Rescinded`. Then only BCeID users can see the record

[Story](https://github.com/orgs/bcgov/projects/137/views/1?pane=issue&itemId=95736380&issue=bcgov%7Clcfs%7C1854)


AlexZorkin (approved):
👍

PR #1905, opened 2025-02-03
feat: Activity List new UI

dhaselhan:
* Remove list bullets and replace with chips
* Separate into two lists with their own header


AlexZorkin (approved):
👍

PR #1906, opened 2025-02-04
feat: Redirect on Login

dhaselhan:
* When user gets redirect to login, store their current path
* When use logs in, check if they have a stored path and take them there


AlexZorkin (approved):
Nice and simple 👍

PR #1918, opened 2025-02-05
Fix: ETL for Compliance Report History

AlexZorkin:
- Created new groovy script to specifically handle compliance report history
- updated data-transfer.sh script to allow for table specific pgdump commands


prv-proton (approved):
That's a good way to reload specific table 👍.

PR #1921, opened 2025-02-05
Fix: LCFS - Filter bug in report pull over from TFRS to the Portal #1892

### LINKED ISSUE #1892: LCFS - Filter bug in report pull over from TFRS to the Portal
**Describe the Bug**
The filter in the Portal is pulling draft reports and showing nothing as returned on the front end to the IDIR side. 

**Expected Behaviour**
An IDIR user should only see a most recent report on the front end, not any drafts. 

**Actual Behaviour**
The system is trying to show the draft report an organization has made, but drafts do not show up on the front end, therefore it just appears that the report is missing

**Implications**
An IDIR user cannot see a report that is in progress in TFRS, but also the organization has started a draft supplemental for. 

**Steps To Reproduce**
User/Role: IDIR 
1.Go to the compliance report view table
2. See that a compliance report that has a draft supplemental in progress is not visible on the front end.

prv-proton:
Fix #1892

### DIFF
--- a/backend/lcfs/web/api/compliance_report/repo.py
+++ b/backend/lcfs/web/api/compliance_report/repo.py
@@ -377,6 +377,12 @@ class ComplianceReportRepository:
             latest_version_subquery = latest_version_subquery.where(
                 ComplianceReport.organization_id == organization_id
             )
+        else:
+            latest_version_subquery = latest_version_subquery.join(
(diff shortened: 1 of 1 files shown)


AlexZorkin (approved):
quick fix nice 👍

Issue #1909, opened 2025-02-04
LCFS - Make "Job Title" Field Optional in Add/Edit BCeID User Page


Grulin:
**Describe the task**
Update the "Job Title" field on the Add/Edit BCeID User page to be optional instead of required. This change is necessary because job title information is not required under the legislation and should not be mandatory for user creation or editing.

**Purpose**
To align the system with legislative requirements by ensuring that unnecessary fields are not mandatory, improving user experience and reducing barriers to adding or editing users.

**Acceptance Criteria**
- [ ]  The "Job Title" field is no longer a required field when adding or editing a user.
- [ ]  Users can leave the "Job Title" field blank and still save changes successfully.
- [ ]  Ensure existing users with no job title do not experience issues after the update.
- [ ]  Validate that the change applies correctly to both new user creation and user edits.
- [ ]  No other required fields are affected by this update.

**Additional context**

PR #1923, opened 2025-02-05
Feat: LCFS - Make "Job Title" Field Optional in Add/Edit BCeID User Page #1909

prv-proton:
#1909


AlexZorkin (approved):
easy that its not a data model fix 👍

PR #1932, opened 2025-02-06
LCFS-1863: Fix calculation for available balance

areyeslo:
When the user exhausts their available balance, they cannot create another transfer.

The balance is set to 50,000 and the user creates a transfer for the same amount:
image

Then, the user cannot create another transfer:
image


AlexZorkin (approved):
👍

PR #1933, opened 2025-02-06
LCFS-1925: Supporting Documents Accordion Displays Even When No Documents Are Uploaded

areyeslo:
When there is no documents, `Supporting Documents` section should not appear. If there is at least one doc, then show the accordion opened:

image

image


AlexZorkin (approved):
👍

PR #1936, opened 2025-02-06
Fix: Clean up transaction history by removing draft entries - 1855

### LINKED ISSUE #1855: LCFS - Created draft transfer history
**Description:**
Remove the "Created draft" line from the Transaction history of a transfer, as it is inconsistent and not an intended record. Additionally, remove "Draft" from the progress bar.

**Wireframe:**  
Image

**Purpose and Benefit to User:**  
The "Created draft" entry in the transaction history is inconsistent and not meant to be displayed. Its presence can cause confusion, as drafts are not considered official transaction records. Similarly, the progress bar should not display "Draft," as it does not represent a meaningful stage in the transfer process. Removing both will ensure a cleaner and more accurate representation of transaction progress.

**Acceptance Criteria:**
- [ ] The "Created draft" entry is removed from the Transaction history of all transfers.
- [ ] The "Draft" stage is removed from the progress bar.
- [ ] No unintended side effects on other transaction history records.
- [ ] No impact on back-end data storage (drafts should still exist, just not be displayed).
- [ ] Ensure history and progress bar updates correctly across all affected pages.

**Development Checklist:**
- [ ] Identify where the "Created draft" entry is generated in the transaction history.
- [ ] Update logic to prevent "Created draft" from being displayed.
- [ ] Remove "Draft" from the progress bar.
- [ ] Verify that transaction history still correctly tracks other relevant statuses.
- [ ] Perform UI testing to confirm that the transaction history and progress bar appear as expected.
- [ ] Validate that no unintended changes impact reporting, auditing, or data integrity.

**Notes:**
- This change simplifies the transaction display by removing unnecessary draft-related records.
- Ensure that drafts can still be saved and

hamed-valiollahi:
This PR removes the "Created draft" entry from transaction history and the "Draft" stage from the progress bar for a clearer, more accurate user experience.

Closes #1855

### DIFF
--- a/frontend/src/views/Transfers/components/TransferHistory.jsx
+++ b/frontend/src/views/Transfers/components/TransferHistory.jsx
@@ -66,6 +66,11 @@ function TransferHistory({ transferHistory }) {
     category = 'C'
   }
 
+  // Filter out any DRAFT records, so “Created draft” never shows up
+  const filteredHistory = transferHistory?.filter(
+    (item) => item.transferStatus?.status !== TRANSFER_STATUSES.DRAFT
+  )
+
   return (
     <BCBox mt={2} data-test="transfer-history">
       <BCTypography variant="h6" color="primary">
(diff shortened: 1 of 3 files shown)


frontend/src/views/Transfers/components/TransferHistory.jsx
```diff
@@ -66,6 +66,11 @@ function TransferHistory({ transferHistory }) {
     category = 'C'
   }
 
+  // Filter out any DRAFT records, so “Created draft” never shows up
```
AlexZorkin, line 69: Let's do this on the backend instead so the frontend just renders what it receives.

AlexZorkin (changes requested):
I think a backend solution might be better just so we're not sending unneeded data to the frontend.

hamed-valiollahi:
> I think a backend solution might be better just so we're not sending unneeded data to the frontend.

Thanks for your feedback, Alex. Please let me know if the PR is ready to merge.

AlexZorkin (approved):
Thanks for frontend solution.

PR #1941, opened 2025-02-06
LCFS-1925: Allow Expand/Collapse for Supporting Documents

areyeslo:
Allow Expand/Collapse for Supporting Documents:

image

image


prv-proton (changes requested):
Supporting docs accordian appearing based on data is working great, but for some reason after I've uploaded the file I'm not able to see any edit icon in front of the Header, as it appears in "Supply of fuel"
image

dhaselhan:
> Supporting docs accordian appearing based on data is working great, but for some reason after I've uploaded the file I'm not able to see any edit icon in front of the Header, as it appears in "Supply of fuel"
That is correct, you manage files through the link/dialog at the top, the supporting documents list is view only.

prv-proton (approved):
looks good 👍

PR #1947, opened 2025-02-06
Fix: UCI Summary Calculation - 1943

AlexZorkin:
- Added missing uci value into summary compliance units calculation


hamed-valiollahi (approved):
LGTM 👍

PR #1948, opened 2025-02-07
feat: Add Export and Template for FSE

dhaselhan:
* Add new Export for FSE to produce templates and download data
* Add data_validators to the spreadsheet builder as well as the ability to set column type
* Add new routes for the new API endpoints and add download button to the front end
* Add tests to FSE


AlexZorkin (changes requested):
The power of openpyxl on full display. Dropdown options in the template excel file are great. Only addition that would be nice is to add a loading spinner to the download button after you click it. The template creation takes a number of seconds and will leave the user confused as to whether the download is happening.

dhaselhan:
> The power of openpyxl on full display. Dropdown options in the template excel file are great. Only addition that would be nice is to add a loading spinner to the download button after you click it. The template creation takes a number of seconds and will leave the user confused as to whether the download is happening.

The worst part is the slowdown is applying the formatting to all the empty cells in case they don't copy/paste. It sets formulas for the first 10,000 rows which made the download go from instant -> few seconds.

AlexZorkin:
> > The power of openpyxl on full display. Dropdown options in the template excel file are great. Only addition that would be nice is to add a loading spinner to the download button after you click it. The template creation takes a number of seconds and will leave the user confused as to whether the download is happening.
> 
> The worst part is the slowdown is applying the formatting to all the empty cells in case they don't copy/paste. It sets formulas for the first 10,000 rows which made the download go from instant -> few seconds.

Gotcha, maybe we ask what the 'average' amount of rows an org would be inputting and we use that. If we can get away with 1000 that might be a faster download?

dhaselhan:
> > > The power of openpyxl on full display. Dropdown options in the template excel file are great. Only addition that would be nice is to add a loading spinner to the download button after you click it. The template creation takes a number of seconds and will leave the user confused as to whether the download is happening.
> > 
> > 
> > The worst part is the slowdown is applying the formatting to all the empty cells in case they don't copy/paste. It sets formulas for the first 10,000 rows which made the download go from instant -> few seconds.
> 
> Gotcha, maybe we ask what the 'average' amount of rows an org would be inputting and we use that. If we can get away with 1000 that might be a faster download?

I asked Lindsy and the hard theoretical limit would be 10,000. I've added a loading spinner but Ill bring this up tomorrow at standup and see if we can drop it in my next MR.

AlexZorkin (approved):
Looks great ty 👍

PR #1954, opened 2025-02-07
LCFS-1863: Credit transfer credit amount maximum

areyeslo:
The user adds digits after quantity amount was adjusted. We need to add isAllowed property in NumericFormat to validate the quantity amount does not go beyond the available balance.

[Story](https://github.com/orgs/bcgov/projects/137/views/1?pane=issue&itemId=95751541&issue=bcgov%7Clcfs%7C1863)


dhaselhan (approved):
LGTM 👍

PR #1976, opened 2025-02-12
Fix: LCFS - Fuel Code error shows when inputting Fuel Production columns #1961

### LINKED ISSUE #1961: LCFS - Fuel Code error shows when inputting Fuel Production columns
**Describe the Bug**
When an analyst enters data in the first of the three "Fuel Production" columns of the new fuel code input grid, a fuel code error appears, preventing further data entry. The error message does not provide useful guidance and seems to be triggered incorrectly.

Image

**Expected Behaviour**
- The fuel code should be validated only when all required fields are entered.
- No error should appear unless the fuel code is actually invalid.
- Analysts should be able to proceed with data entry without being blocked by an erroneous validation message.

**Actual Behaviour**
- A fuel code error is displayed as soon as data is entered in the first "Fuel Production" column.
- The analyst is unable to proceed past this step.
- The error message does not clearly explain what is causing the issue.

**Implications**
- Prevents analysts from completing fuel production data entry.
- Causes confusion due to the unclear error message.
- May delay compliance reporting if not resolved.

**Steps To Reproduce**
User/Role: IDIR

1. Log into the LCFS Portal as an IDIR user.
2. Navigate to the new fuel code 
3. Enter data in the first of the three "Fuel Production" columns.
4. Observe the fuel code error appearing prematurely.

prv-proton:
Fix #1961


frontend/src/views/FuelCodes/AddFuelCode/_schema.jsx
```diff
@@ -452,36 +446,33 @@ export const fuelCodeColDefs = (optionsData, errors, isCreate, canEdit) => [
...
-      multiple: false,
-      disableCloseOnSelect: false,
-      freeSolo: true,
-      openOnFocus: true
-    },
+    cellEditorParams: (params) => ({
+      queryKey: 'fuel-production-province-search',
+      queryFn: async ({ queryKey, client }) => {
+        let path = apiRoutes.fuelCodeSearch
+        path += 'fpProvince=' + queryKey[1]
```
dhaselhan, line 457: Should you URIEncode the queryKey in case it contains ? or spaces?

frontend/src/views/FuelCodes/AddFuelCode/_schema.jsx
```diff
@@ -452,36 +446,33 @@ export const fuelCodeColDefs = (optionsData, errors, isCreate, canEdit) => [
...
+        const response = await client.get(path)
+        return response.data
+      },
+      title: 'fuelProductionFacilityProvinceState',
+      enabled: true
+    }),
     minWidth: 325,
     valueSetter: (params) => {
-      params.data.fuelProductionFacilityProvinceState = params.newValue
+      if (!params.newValue) return false // Handle empty values safely
```
dhaselhan, line 466: Do we still need to set the empty value on params.data? Is it leaving the old value around?

dhaselhan (approved):
LGTM 👍

frontend/src/views/FuelCodes/AddFuelCode/_schema.jsx
prv-proton: fixed.

PR #1998, opened 2025-02-13
Feat: Draft Schedule Records Exclusions - 1931

AlexZorkin:
- updated schedules to exclude draft records for gov users
- updated tests


backend/lcfs/tests/fuel_export/test_fuel_exports_services.py
```diff
@@ -44,6 +54,14 @@ async def test_get_fuel_export_options_success(fuel_export_service, mock_repo):
 
 @pytest.mark.anyio
 async def test_get_fuel_export_list_success(fuel_export_service, mock_repo):
+    # Set up a dummy request with a valid user
+    from types import SimpleNamespace
```
dhaselhan, line 58: Why is this import inside the test instead of at the top of the file? Is this a Python thing I'm not aware of?

dhaselhan (approved):
LGTM 👍 , just some nitpicks and questions

PR #1999, opened 2025-02-13
fix: Add Decimal6 to XLSX

dhaselhan:
* Lat/Long can have 6 decimals, so add a custom formatter just for that
* Update front end to allow 6 decimals and add a max of 1000


areyeslo (approved):
LGTM

PR #2003, opened 2025-02-13
Fix: Adjust Notification Filter for Analysts - 1916

### LINKED ISSUE #1916: LCFS - Adjust Notification Filter for IDIR Transfer Notifications
**Describe the task**
Modify the IDIR notification system so that analysts receive only one notification per transfer transaction, instead of separate notifications for both the Transferer and Transferee. This will help prevent confusion when recording multiple transfers and ensure analysts can clearly identify which organization has received the credits.

**Purpose**
To reduce notification overload and confusion for analysts, making it easier to track transfers accurately. By consolidating notifications by Transaction ID and displaying only the receiving organization in the view table, analysts can better confirm that the correct organization has received the credits.

**Acceptance Criteria**
- [ ] Analysts receive only one notification per transfer transaction ID, rather than separate notifications for both transfer parties.
- [ ]  Notifications clearly indicate the receiving organization rather than both parties.
- [ ]  Update the transfer view table to only display the receiving organization for clarity.
- [ ]  Test to ensure no unintended notification disruptions occur for other user roles.
- [ ]  Validate the change across email and in-app notifications.

**Additional context**
- Investigate whether the current duplicate notifications are due to separate triggers for sender and receiver.
- Confirmed this is only happening on the IDIR side

hamed-valiollahi:
This PR updates the analyst notification system to prevent duplicate notifications for recorded and refused transfer transactions.

Closes #1916


AlexZorkin (approved):
👍

PR #2026, opened 2025-02-14
feat: Add Import of FSE's

dhaselhan:
* Add new importer.py to handle imports
* Add new dialog with states to go through the import process
* Use a newly spawned thread to process the import async


AlexZorkin (approved):
This is a powerful feature, and code looks great. nice work! 👍

PR #2039, opened 2025-02-15
feat: Add Line 15/16 Calculations

dhaselhan:
* Look up line 18/19 of previous report and add them all together
* Add new unit test to cover scenario


backend/lcfs/web/api/compliance_report/summary_service.py
```diff
@@ -778,9 +782,18 @@ async def calculate_low_carbon_fuel_target_summary(
         compliance_units_issued = await self.repo.get_issued_compliance_units(
             compliance_period_start, compliance_period_end, organization_id
         )  # line 14
-        # TODO - add the logic as required
-        compliance_units_prev_issued_for_fuel_supply = 0  # line 15
-        compliance_units_prev_issued_for_fuel_export = 0  # line 16 - always 0
+        compliance_units_prev_issued_for_fuel_supply = (
+            previous_summary.line_18_units_to_be_banked
```
AlexZorkin, line 786: So in the scenario of 2 or more supplementals, this value will be the sum of all prev credits issued in the whole chain of reports without double adding them? just want to confirm that with n supplementals this previous value is the previous total of all reports and not an ever increasing/decreasing value.

frontend/src/views/CompareReports/CompareReports.jsx
AlexZorkin, line 1: nice cleanup.

AlexZorkin (approved):
Looks good, just one question to confirm the changes. 👍

PR #2043, opened 2025-02-18
feat: Add FSE Tests

dhaselhan:
* Move FSE Schema into FSE schema.py instead of ComplianceReport schema


AlexZorkin (approved):
Looks good 👍

PR #2044, opened 2025-02-18
feat: Feature Flag Legacy Details

dhaselhan:
* Remove no longer used notification flag
* Default to true on localhost


AlexZorkin (approved):
Looks good 👍  
please followup with our devops lead to implement this in the deployment config for production env. test can just leave this feature enabled as well.

PR #2045, opened 2025-02-18
fix: New Excel Validation Approach

dhaselhan:
* Write a second sheet of valid values instead of using formula values directly
* List formulas do not allow special characters nor do they support escaping
* This new approach does double the XLSX size but allows any values in the selects (such as comma)


AlexZorkin (approved):
nice update 👍 
I'd like to hear more about the second sheet idea.

PR #2047, opened 2025-02-19
feat: Add DRAFT back to IDIR filter options

dhaselhan:
* Was filtered out before, we now want to allow it
* Backend is still correctly filtering who can see what


AlexZorkin (approved):
👍

PR #2059, opened 2025-02-20
Fix: LCFS - BUG Government user able to use the URL transfer ID to view a transfer in "sent" status #2038

### LINKED ISSUE #2038: LCFS - BUG Government user able to use the URL transfer ID to view a transfer in "sent" status
**Describe the Bug**
We would like to prevent government users from being able to view transfers in a "sent" status by inputting the transfer ID # in the URL

**Expected Behaviour**
Government IDIR users should not be able to view transfers in a "sent" status

**Actual Behaviour**
Government IDIR users can view transfers in a "sent" status by inputting the transfer ID # in the URL 

**Implications**
Sent transfers should not be visible to Government users. Only should be visible once they are submitted

**Steps To Reproduce**
User/Role: IDIR
1. Go to any transfer
2. Use the URL to input a number for a transfer that is in a "sent" status
3. Observe that as an IDIR user, you can view the "sent" transfer

prv-proton:
Fix #2038

### DIFF
--- a/backend/lcfs/web/api/transfer/services.py
+++ b/backend/lcfs/web/api/transfer/services.py
@@ -91,7 +91,17 @@ class TransferServices:
     async def get_transfer(self, transfer_id: int) -> TransferSchema:
         """Fetches a single transfer by its ID and converts it to a Pydantic model."""
         transfer = await self.repo.get_transfer_by_id(transfer_id)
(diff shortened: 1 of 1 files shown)


AlexZorkin (approved):
👍

PR #2063, opened 2025-02-21
fix: Remove return types from Save

dhaselhan:
* This was causing the CreateAllocationAgreementSchema validation to run despite the types being defined as a union. AFAIK Flask does not like union types on views.


AlexZorkin (approved):
👍

PR #2066, opened 2025-02-21
Fix: Restrict Pencil Icon Visibility - 1995

### LINKED ISSUE #1995: LCFS - BUG - Pencil icon appearing when it shouldn't
**Describe the Bug**  
The pencil icon appears for users who do not have the "Manage User" role. This creates the impression that they can edit user details when they should not have access to this functionality.  

Image

**Expected Behaviour**  
The pencil icon should only be displayed for users with the "Manage User" role. Users without this role should not see the icon.  

**Actual Behaviour**  
The pencil icon is visible for all users, regardless of whether they have the "Manage User" role.  

**Implications**  
- Unauthorized users may believe they can edit user details, leading to confusion.  
- The UI does not accurately reflect role-based permissions.  

**Steps To Reproduce**  
1. Log in as a user without the "Manage User" role.  
2. Navigate to the user management page.  
3. Observe that the pencil icon is still visible.

hamed-valiollahi:
### Changes in this PR:
- Ensures the pencil icon is only visible to users with the "Manage User" and "Administrator" roles.

Closes #1995

### DIFF
--- a/frontend/src/views/Admin/AdminMenu/components/ViewUser.jsx
+++ b/frontend/src/views/Admin/AdminMenu/components/ViewUser.jsx
@@ -140,11 +140,15 @@ export const ViewUser = () => {
               id="user-card"
               title={t('admin:userDetails')}
               color="nav"
-              editButton={{
-                text: canEdit ? t('admin:editBtn') : null,
-                route: canEdit ? editButtonRoute : null,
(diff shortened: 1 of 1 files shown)


AlexZorkin (approved):
👍

PR #2071, opened 2025-02-21
fix: Fix Org Filter

dhaselhan:
* Prop is queryParams not queryOptions
* Add selectedOrgId into the filter key so changing it does trigger a re-fetch


AlexZorkin (approved):
👍

PR #2089, opened 2025-02-24
Fix: LCFS - BUG - Comment Clears When Second Supplier Clicks Signature Checkbox in Transfers #1977

### LINKED ISSUE #1977: LCFS - BUG - Comment Clears When Second Supplier Clicks Signature Checkbox in Transfers
**Describe the Bug:**  
When the second supplier types a comment in a transfer and then clicks the signature checkbox, the comment field gets cleared unexpectedly.  

**Expected Behaviour:**  
The comment should remain intact when the signature checkbox is clicked, allowing the second supplier to submit both the comment and their signature without data loss.  

**Actual Behaviour:**  
- The second supplier types a comment.  
- Upon clicking the signature checkbox, the entered comment disappears.  

**Implications:**  
- Loss of user-entered comments, potentially leading to missing information in transfer records.  
- User frustration due to unintended data clearing.  
- Increased need for re-entry, reducing efficiency.  

**Steps To Reproduce:**  
1. As the second supplier, navigate to a transfer requiring a signature, SENT status
2. Enter a comment in the comment field.  
3. Click the signature checkbox.  
4. Observe that the comment disappears.  

**Notes:**  
- Verify if this issue occurs across different browsers.  
- Check if the comment clears only on the second supplier’s side or for all users.  
- Ensure changes are saved correctly before toggling the signature checkbox.

prv-proton:
Fix #1977

### DIFF
--- a/frontend/src/views/Transfers/AddEditViewTransfer.jsx
+++ b/frontend/src/views/Transfers/AddEditViewTransfer.jsx
@@ -121,41 +121,21 @@ export const AddEditViewTransfer = () => {
   useEffect(() => {
     if (!transferId) return
     if (isFetched && transferData) {
-      // Fetch the comments from the transfer data
-      const fromOrgCommentObj = transferData.comments?.find(
-        (c) => c.commentSource === 'FROM_ORG'
-      )
-      const fromOrgCommentText = fromOrgCommentObj?.comment || ''
-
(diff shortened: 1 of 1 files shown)


dhaselhan (approved):
LGTM 👍

PR #2095, opened 2025-02-25
Fix: Notional Transfers Method Param - 2073

AlexZorkin:
- missing request.user param added to notional transfers paginated method


dhaselhan (approved):
LGTM 👍

PR #2105, opened 2025-02-26
Fix: Masked Report Logic, 2024 CR Dropdown - 2076 2103

AlexZorkin:
- fixed masked report logic on backend cr
- temporary hard code of 2024 new reports in dropdown
- dependancy fix on addedituser


hamed-valiollahi (approved):
Looks good to me! 👍

PR #2108, opened 2025-02-26
Add 5 Credits to Fuel Supply, Fuel Export, and Credit Calculations

dhaselhan:


AlexZorkin (approved):
Looks good to me. Let's get this merged and have a couple business area folks to confirm that the numbers match what they're expecting before we go to prod.

PR #2123, opened 2025-02-27
fix: Filter Recommended Status if Transfer is already approved

dhaselhan:
* If transfer is approved and recommended is NOT in history, do not add it
* Remove and clean up dead code from AddEditViewTransfer.


prv-proton (approved):
Looks good 👍

PR #2131, opened 2025-02-28
HotFix: limit added to materialized view subqueries

AlexZorkin:
- subqueries in materialized view were returning more than one result


hamed-valiollahi (approved):
Looks good to me! 👍

PR #2136, opened 2025-02-28
feat: Add Not Recommended status

dhaselhan:
* We need legacy status from TFRS so we can map it in LCFS


prv-proton (approved):
Looks good 👍

PR #2140, opened 2025-03-03
Feat: Delete Actions for Schedules

hamed-valiollahi:
- Implemented and verified delete actions for Notional Transfer, Allocation Agreement, Fuel Export, Fuel Supply, Other Use, and Final Supply Equipment schedules.

### DIFF
--- a/backend/lcfs/tests/fuel_export/test_fuel_exports_repo.py
+++ b/backend/lcfs/tests/fuel_export/test_fuel_exports_repo.py
@@ -232,17 +232,3 @@ async def test_update_fuel_export_success(fuel_export_repo, mock_db):
             "end_use_type",
         ],
     )
-
-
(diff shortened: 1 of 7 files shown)


AlexZorkin (changes requested):
As discussed we would like to revert the FSE versioning changes, and just test that all other schedule delete actions work in a supplemental change log. ty

hamed-valiollahi:
> As discussed we would like to revert the FSE versioning changes, and just test that all other schedule delete actions work in a supplemental change log. ty

I reverted the changes related to FSE versioning and tested all other schedules. Just wanted to confirm whether we need versioning for the allocation agreement since there are no versioning columns in the database.

AlexZorkin (approved):
looks good 👍

PR #2142, opened 2025-03-03
fix: BCGridViewer refactor spike

### LINKED ISSUE #1926: LCFS - Improve AG grid viewer component
We are currently passing the api query function directly to the ag grid viewer component and letting it do the fetching. Instead we should be fetching the data outside of the component and then passing it in. 

The way that it is currently set up, we have to modify the backend to conform to the frontend. it should be the other way around. have a standard output from the backend and modify the data where needed on the frontend. We cant modify the data based on side effects since the data is basically locked.

This way, we can leverage the caching that react query does as well, reducing the number of calls to the backend.

kevin-hashimoto:
Note: This is just a refactor of the `BCGridViewer` component. It is not the implementation so I've created the `BCGridViewer2` component so pages that use the legacy component still work. Once the refactored component is implemented into the pages, we can rename it back to `BCGridViewer`

implementation example in `FuelCodes.jsx` file

- refactor BCGridViewer component to be dumb and stateless for better flexibility, scalability and ease of use, closes #1926 
- forward ref and imperative handling to bring the `resetGrid` function to the parent
- query fetching is removed. This will now be done in the parent component, passing the data in. We can now manipulate this data in the parent component if we need to, whereas we could not do this before.
- pagination options logic simplification. the pagination options are moved to the parent.


prv-proton (approved):
Looks very good, simplifying and more control over pagination is handled very well. 👍

PR #2143, opened 2025-03-03
fix: Remove special fuel category logic

dhaselhan:
* Use the regular field instead of ID, like most other schedules
* Delete git conflict


prv-proton (approved):
Looks good 👍

PR #2144, opened 2025-03-03
Update 2024 Schedules with 5 digit TCI

dhaselhan:


backend/lcfs/db/migrations/versions/2025-03-01-22-42_fd8ee994668c.py
AlexZorkin, line 1: Well done on this migration. Testing in a production scenario before merge would help verify the effects of this.

backend/lcfs/web/api/compliance_report/repo.py
```diff
@@ -3,11 +3,14 @@
 from typing import List, Optional, Dict, Union
 
 import structlog
+from datetime import datetime
 from fastapi import Depends
 from sqlalchemy import func, select, and_, asc, desc, update, String, cast
 from sqlalchemy.ext.asyncio import AsyncSession
 from sqlalchemy.orm import joinedload, aliased
 from sqlalchemy.inspection import inspect
+from sqlalchemy.orm import joinedload, aliased
```
AlexZorkin, line 12: duplicate imports?

backend/lcfs/web/api/compliance_report/views.py
```diff
@@ -119,9 +118,7 @@ async def get_compliance_report_summary(
     Retrieve the comprehensive compliance report summary for a specific report by ID.
     """
     await validate.validate_organization_access(report_id)
-    return await summary_service.calculate_compliance_report_summary(
-        report_id, request.user
-    )
+    return await summary_service.calculate_compliance_report_summary(report_id)
```
AlexZorkin, line 121: Without passing in the user, how do we exclude draft summary calculations for government users? Gov users should not see summary calculations from a supplemental draft report changes. Are you handling this at the schedule level now?

AlexZorkin (changes requested):
I like the include_drafts change but just want to confirm that the get_summary endpoint for gov_users will hide the draft changes. 👍

backend/lcfs/web/api/compliance_report/views.py
dhaselhan: And once we go into the world of re-assessment, that will no longer be true as Government users will be the ones creating the draft records

backend/lcfs/web/api/compliance_report/repo.py
dhaselhan: Fixed!

AlexZorkin (approved):
Changes look good 👍

PR #2146, opened 2025-03-03
feat: Add Compliance Report Export

dhaselhan:
* Add exporter for all compliance report schedules and the summary
* Since summary is non-rectangle, add raw_sheet functionality to builder to handle new formats


backend/lcfs/web/api/compliance_report/export.py
```diff
@@ -0,0 +1,515 @@
...
+            rows=data,
+            styles={"bold_headers": True},
+        )
+        return builder
+
+    async def load_fuel_supply_data(
+        self, compliance_report_group_uuid: str, is_draft: bool
+    ) -> List[List]:
+        results = await self.fs_repo.get_effective_fuel_supplies(
+            compliance_report_group_uuid, not is_draft
```
kevin-hashimoto, line 246: It may have been discussed somewhere else but whats the reason for not showing some of these sections for draft reports? I see that FSE is the only one that gets exported for draft reports

kevin-hashimoto (approved):
LGTM! button showing on the necessary pages and download is working with the correct template

backend/lcfs/web/api/compliance_report/export.py
dhaselhan: These do show for draft reports, I need to pass the draft=True if the report is draft so it shows the draft records in the download.

PR #2147, opened 2025-03-04
fix: Feature Fixes

dhaselhan:
* Update status enum migration to only add values if they dont exist so I can manually insert in prod
* Add new migration to swap the Target CI of Diesel and Gasoline, it was entered backwards
* Fix dropdowns on FuelExports


backend/lcfs/db/migrations/versions/2025-03-04-19-29_ba5a7f5c6a7c.py
AlexZorkin, line 1: It would be better to just fix the existing migration and then update these manually to avoid having an incorrect state in the migration history. Let me know what you think.

AlexZorkin (changes requested):
Looks good, just one comment to address.

backend/lcfs/db/migrations/versions/2025-03-04-19-29_ba5a7f5c6a7c.py
dhaselhan: How can I fix the values on develop in that case?

AlexZorkin (approved):
Looks good ty 👍

PR #2157, opened 2025-03-05
hotfix: Fix selection of organization in Allocation Agreement

areyeslo:
On Dev, the user cannot select  a value in the column Legal name of transaction partner. The schema is expecting the label and not all the object, so the setter is not triggered. As a result, we are not able to show organization details in the grid:

image

image


areyeslo:
Closing PR , instead this https://github.com/bcgov/lcfs/pull/2156

PR #2174, opened 2025-03-06
Fix/prashanth-addr-null-bug

prv-proton:
- Fix construct address to work even with undefined address being passed.
- Fix Editor to work with null address fields.


AlexZorkin (approved):
👍

PR #2178, opened 2025-03-06
Fix: Hide Draft Status from IDIR Users in Compliance Report History - 2074

### LINKED ISSUE #2074: LCFS - BUG - Draft supplemental report status appearing to IDIR
**Describe the Bug**
This bug may only occur after a second supplemental report is created. 
  
IDIR users are seeing "Draft" statuses in the Compliance Report history when supplemental reports are created. The "Draft" status should not be visible to IDIR users.

Image

Image

**Expected Behaviour**  
- IDIR users should only see relevant statuses in the Compliance Report history.  
- "Draft" status should not appear to IDIR users when supplemental reports are created.  

**Actual Behaviour**  
- "Draft" statuses are appearing in the Compliance Report history for IDIR users.  
- This occurs specifically when supplemental reports are created.  

**Implications**  
- Causes confusion for IDIR users as draft reports should not be visible to them.  
- Could lead to incorrect assumptions about report status.  
- May indicate an issue with permissions or filtering logic in the Compliance Report history.  

**Steps To Reproduce**  
User/Role: IDIR User  
1. Log in as an IDIR user.  
2. Navigate to the Compliance Report history.  
3. Create a supplemental report.  
4. Observe that "Draft" statuses appear in the history, which should not be visible.  

**Notes:**  
- Review permission settings for Compliance Report history visibility.  
- Confirm whether this issue is limited to supplemental reports or affects other report types.  
- See attached screen captures for reference.

hamed-valiollahi:
This PR includes the following fixes:

- Hides "Draft" status from IDIR users in Compliance Report history.
- Updates filtering logic to exclude draft statuses.
- Adds tests for components.

Closes #2074

### DIFF
--- a/backend/lcfs/tests/compliance_report/test_compliance_report_views.py
+++ b/backend/lcfs/tests/compliance_report/test_compliance_report_views.py
@@ -253,7 +253,7 @@ async def test_get_compliance_report_by_id_success(
 
         assert response.json() == expected_response
         mock_get_compliance_report_by_id.assert_called_once_with(
-            1, False, get_chain=True
+            1, False, True, get_chain=True
         )
         mock_validate_organization_access.assert_called_once_with(1)
 
--- a/backend/lcfs/web/api/compliance_report/services.py
+++ b/backend/lcfs/web/api/compliance_report/services.py
@@ -236,14 +236,25 @@ class ComplianceReportServices:
 
     @service_handler
     async def get_compliance_report_by_id(
-        self, report_id: int, apply_masking: bool = False, get_chain: bool = False
+        self,
+        report_id: int,
+        apply_masking: bool = False,
+        is_gov: bool = False,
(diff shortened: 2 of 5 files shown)


AlexZorkin (changes requested):
This needs to be handled on the backend because the draft history record is still visible in the response payload, which technically allows gov users to still see the record.

hamed-valiollahi:
> This needs to be handled on the backend because the draft history record is still visible in the response payload, which technically allows gov users to still see the record.

You're right, thanks for your feedback! I'll apply the requested change today.

hamed-valiollahi:
I've applied your requested changes. I've also refactored the code to handle removing draft report history entries on the back-end side. Please let me know if any other changes are required. Thank you!

AlexZorkin (approved):
Looks good ty 👍

PR #2180, opened 2025-03-07
fix: fuel export date accessibility

### LINKED ISSUE #2138: LCFS BUG - Date picker bugs on FSE and Export fuel in AG Grid
**Describe the Bug**  

**Export Fuels** - Users cannot select the date picker using the keyboard's enter key. 

Image

**FSE** - Users must tab through each day when selecting a date. The arrow keys do not work while in the date picker view.  

Image

**Expected Behaviour**  

- Export fuel - Users should be able to select a date in the date picker by pressing the enter key. 
- FSE - Users should be able to navigate in the date picker using arrow keys to select dates.  

**Actual Behaviour**  

- The date picker does not respond to the enter key, forcing users to use a mouse.  
- Users must tab through each day because the arrow keys are non-functional.  

**Implications**  

- Decreases accessibility for users, particularly those relying on assistive technologies.  
- Significantly reduces usability and efficiency, making date selection cumbersome.  
- Impacts compliance with accessibility standards.

**Steps To Reproduce**  

User/Role: Any user interacting with the date picker.  

1. Navigate to the Export Fuels or FSE section where the date picker is available.  
2. Attempt to select a date using only the keyboard (press enter on a selected date).  
3. Observe that the enter key does not trigger date selection.  
4. Attempt to use arrow keys to navigate through dates.  
5. Observe that the arrow keys do not function, forcing the user to tab through each date.  

**Notes**  
- Ensure that the date picker is accessible via keyboard navigation and selection.  
- Please validate compliance with WCAG 2.1 accessibility guidelines.  
- Test across different browsers and devices to confirm consistent behaviour.

kevin-hashimoto:
- fix: not being able to tab open the date picker, closes #2138 

the date range picker in FSE uses a package that does not look like we'll be able to customize to add the ability to use arrow keys to select the date range. may need to explore other alternative packages that offer this feature or at least be able to customize to add this feature. spike ticket created: https://github.com/bcgov/lcfs/issues/2184

### DIFF
--- a/frontend/src/views/FuelExports/AddEditFuelExports.jsx
+++ b/frontend/src/views/FuelExports/AddEditFuelExports.jsx
@@ -307,7 +307,6 @@ export const AddEditFuelExports = () => {
             showAddRowsButton={true}
             context={{ errors }}
             onAction={onAction}
-            stopEditingWhenCellsLoseFocus
             saveButtonProps={{
               enabled: true,
               text: t('report:saveReturn'),
--- a/frontend/src/views/FuelExports/_schema.jsx
+++ b/frontend/src/views/FuelExports/_schema.jsx
@@ -102,7 +102,6 @@ export const fuelExportColDefs = (
 
     suppressKeyboardEvent,
     cellEditor: DateEditor,
-    cellEditorPopup: true,
     cellEditorParams: {
       autoOpenLastRow: !gridReady
     }


frontend/src/views/FuelExports/AddEditFuelExports.jsx
```diff
@@ -307,7 +307,6 @@ export const AddEditFuelExports = () => {
             showAddRowsButton={true}
             context={{ errors }}
             onAction={onAction}
-            stopEditingWhenCellsLoseFocus
```
dhaselhan, line 310: Does this change impact when the grid fires its save call? If I press tab does it still save?

dhaselhan (approved):
LGTM 👍

frontend/src/views/FuelExports/AddEditFuelExports.jsx
kevin-hashimoto: confirmed that it still functions properly

Issue #2152, opened 2025-03-04
LCFS - Refactor BCGridViewer to BCGridViewer2 in Notional Transfer Summary, Other Uses Summary, and Transactions Components


AlexZorkin:
#### **Describe the task**
Refactor the `BCGridViewer` component to `BCGridViewer2` in the following components:
1. `NotionalTransferSummary`
2. `OtherUsesSummary`
3. ~~`Transactions`~~

Each instance of `BCGridViewer` should be replaced with `BCGridViewer2`, ensuring all necessary props are correctly mapped. Once all instances are replaced, the old `BCGridViewer` component will be deprecated.

#### **Purpose**
The `BCGridViewer2` component provides improved functionality, better performance optimizations, and a consistent API across the application. This refactor aligns these components with the new standard for grid rendering, improving maintainability and future-proofing the system.

#### **Acceptance Criteria**
- [ ] Replace all instances of `BCGridViewer` with `BCGridViewer2` in `NotionalTransferSummary`, `OtherUsesSummary`, and `Transactions` components.
- [ ] Ensure correct prop mappings from the old component to the new one.
- [ ] Verify that filters, pagination, and sorting continue to work as expected.
- [ ] Confirm that no breaking changes occur in the affected components.
- [ ] Ensure visual consistency and correct rendering of data in the UI.

#### **Additional Context**
- Refer to the `FuelCodes` component for a correct implementation of `BCGridViewer2`.
- Ensure that the correct `queryData` prop is passed instead of `query` and `queryParams`.
- Pagination handling should be updated to follow the `BCGridViewer2` standard.
- The reset filter functionality should work correctly after refactoring.
- Any issues with the transition should be documented and addressed.

This is the fourth batch of refactors, and subsequent tickets will cover additional components.

PR #2186, opened 2025-03-07
fix: transaction id sorting

### LINKED ISSUE #2069: LCFS - BUG - Transaction ID Column Sorting Incorrect on BCeID and IDIR Transaction Pages
**Describe the Bug:**  
The **Transaction ID** column does not sort correctly on both the **BCeID** and **IDIR** transaction pages. Instead of sorting numerically, the IDs are sorted as text, leading to incorrect sequences such as **CR1, CR100, CR11** instead of **CR1, CR11, CR100**.  

**Expected Behaviour:**  
- The **Transaction ID** column should be sorted **numerically**, ensuring IDs appear in proper sequential order.  
- Sorting should maintain grouping by prefix.

**Actual Behaviour:**  
- Sorting treats transaction IDs as **text** rather than numbers.  
- Incorrect sequences occur, such as **CR1, CR100, CR11** instead of **CR1, CR11, CR100**.  

**Implications:**  
- Users cannot efficiently locate or track transactions.  
- Sorting inconsistencies may lead to reporting errors or confusion.  
- The issue affects **both BCeID and IDIR** users, impacting multiple roles.  

**Steps To Reproduce:**  
User/Role: **BCeID or IDIR**  
1. Navigate to the **Transactions** page.  
2. Click on the **Transaction ID** column to apply sorting.  
3. Observe that IDs are not sorted numerically but as text (e.g., **CR1, CR100, CR11**).  
4. Verify that sorting fails to follow expected numeric order while maintaining grouping by prefix.  

**Notes:**  
- Update sorting logic to recognize and **sort numeric values correctly**, even when prefixed (e.g., "CR").  
- Ensure sorting behaves correctly across different transaction ID formats.  
- Test sorting behavior across both **ascending and descending orders**.  
- Validate fix for **both BCeID and IDIR** transaction pages.

kevin-hashimoto:
- fix: transaction id not sorting correctly by numerical value, closes #2069 
- refactor: bc grid viewer to new bc grid viewer 2, touches on #2152

### DIFF
--- a/frontend/src/views/Transactions/_schema.js
+++ b/frontend/src/views/Transactions/_schema.js
@@ -31,6 +31,12 @@ export const transactionsColDefs = (t) => [
     },
     filterParams: {
       buttons: ['clear']
+    },
+    comparator: (valueA, valueB) => {
+      const numberA = parseInt(valueA.slice(1), 10)
+      const numberB = parseInt(valueB.slice(1), 10)
+
+      return numberB - numberA
     }
   },
   {
--- a/frontend/src/views/Transactions/Transactions.jsx
+++ b/frontend/src/views/Transactions/Transactions.jsx
@@ -24,8 +24,15 @@ import { govRoles, roles } from '@/constants/roles'
 import OrganizationList from './components/OrganizationList'
 import Loading from '@/components/Loading'
 import { ConditionalLinkRenderer } from '@/utils/grid/cellRenderers.jsx'
-import { BCGridViewer } from '@/components/BCDataGrid/BCGridViewer'
+import { BCGridViewer2 } from '@/components/BCDataGrid/BCGridViewer2'
 import { useGetTransactionList } from '@/hooks/useTransactions'
+
+const initialPaginationOptions = {
+  page: 1,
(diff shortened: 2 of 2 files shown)


frontend/src/views/Transactions/_schema.js
```diff
@@ -31,6 +31,12 @@ export const transactionsColDefs = (t) => [
     },
     filterParams: {
       buttons: ['clear']
+    },
+    comparator: (valueA, valueB) => {
+      const numberA = parseInt(valueA.slice(1), 10)
+      const numberB = parseInt(valueB.slice(1), 10)
+
+      return numberB - numberA
```
areyeslo, line 39: is it possible to sort by the transactions at the repo level for each of the endpoints?

kevin-hashimoto: I've implemented BCGridViewer2 so that we can have control of the sorting on the backend but even though we sort in the backend, ag grid will run a sort on the front end. so it looks like we will need this on the front end as well.

areyeslo (approved):
LGTM

PR #2187, opened 2025-03-07
Feat: Updated Txn Aggregate MV

AlexZorkin:
- Update the mv_transaction_aggregate materialized view to properly calculated the correct transfer compliance period
- Updated this manually in each environment and add the patch to a previously run migration


areyeslo (commented):
LGTM

areyeslo (approved):

PR #2194, opened 2025-03-10
fix: Properly populate endUseType in Fuel Exports

dhaselhan:
* Remove any default, Any is not a valid option for all fuel types (electricity)
* Update logic to look at unique fuel types, with the new code for EER ratios there can be many Anys per fuel type.


areyeslo (approved):
LGTM

PR #2197, opened 2025-03-10
feat: Change documents to download

dhaselhan:
* Change from inline -> download so we can specify the file name
* Add access checks for each of the current file types
* Add tests

### DIFF
--- a/backend/lcfs/tests/document/test_document_view.py
+++ b/backend/lcfs/tests/document/test_document_view.py
@@ -1,0 +1,213 @@
+from http.client import HTTPException
+
+import io
+import pytest
+from fastapi import UploadFile
+from starlette import status
+
+from lcfs.db.models.user.Role import RoleEnum
+from lcfs.services.s3.client import DocumentService
+from lcfs.web.api.compliance_report.validation import ComplianceReportValidation
+from lcfs.web.api.initiative_agreement.validation import InitiativeAgreementValidation
+from lcfs.web.api.admin_adjustment.validation import AdminAdjustmentValidation
+
+
+class DummyDocument:
+    def __init__(self, file_name: str, document_id: int, file_size: int):
+        self.file_name = file_name
+        self.document_id = document_id
+        self.file_size = file_size
+
+
+class FakeDocumentService:
+    async def get_by_id_and_type(self, parent_id: int, parent_type: str):
+        return [
+            {
+                "document_id": 1,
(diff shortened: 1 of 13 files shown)


backend/lcfs/tests/document/test_document_view.py
AlexZorkin, line 1: Very thorough ty

frontend/src/hooks/useDocuments.js
```diff
@@ -83,11 +83,21 @@ export const useViewDocument = (parentType, parentID, options) => {
...
-    window.open(fileURL, '_blank')
+    const link = document.createElement('a')
+    link.href = fileURL
 
-    // Revoke the object URL after opening
-    setTimeout(() => URL.revokeObjectURL(fileURL), 100)
+    // Extract filename from headers if available, otherwise set a default
+    const contentDisposition = res.headers['content-disposition']
+    const match =
+      contentDisposition && contentDisposition.match(/filename="(.+)"/)
```
AlexZorkin, line 92: This is a great way to pass the filename. In the case of the default name it would be better to have the filename be slightly less generic. we could do something like the following.
```
const mimeToExtension = {
  'application/pdf': 'pdf',
  'image/png': 'png',
  'image/jpeg': 'jpg',
  'application/msword': 'doc',
  'application/vnd.openxmlformats-officedocument.wordprocessingml.document': 'docx',
  'text/plain': 'txt',
  // Add other MIME types as needed
};

const contentType = res.headers['content-type'];
const extension = mimeToExtension[contentType] || 'bin';

const contentDisposition = res.headers['content-disposition'];
const match = contentDisposition && contentDisposition.match(/filename="(.+)"/);
const defaultFileName = `document-${new Date().toISOString().slice(0,10)}.${extension}`;
link.download = match ? match[1] : defaultFileName;
```
might be overkill if filename is always available but let me know if you think this is warranted.

AlexZorkin (changes requested):
Looks great, just one idea to review. ty

frontend/src/hooks/useDocuments.js
dhaselhan: I'll remove the default logic to make it more clear

dhaselhan: Removed unnecessary default

AlexZorkin (approved):
Looks good 👍

Issue #2014, opened 2025-02-13
LCFS - Fix Cypress Testing Pipeline for Login Flow Authentication


AlexZorkin:
**Description**  
Investigate and resolve failures in the daily Cypress testing job executed via GitHub Actions. Initial analysis suggests that login credentials may not be properly retrieved from GitHub secrets. The job currently exhibits multiple failures during the login flow, including missing elements and syntax errors.

**Task**  
- Analyze the GitHub Actions workflow to verify proper injection of login credentials from GitHub secrets.  
- Review the test scripts, particularly user_flow.cy.js, for syntax errors and potential misconfigurations in login steps.  
- Debug the authentication failures and ensure that help links and error messages appear as expected.  
- Update the testing pipeline and credentials management to ensure a stable daily run.

**Acceptance Criteria**  
- GitHub secrets are correctly accessed and injected into the Cypress environment.  
- All tests, especially those related to user login, pass without syntax or authentication errors.  
- The Cypress testing pipeline runs successfully with reliable login functionality and accurate reporting.

This ticket aims to stabilize the Cypress testing pipeline by addressing credential injection issues and ensuring that login flows function correctly in all environments.

Issue #2015, opened 2025-02-13
LCFS - Add Cypress Test Scenario for Multi-Schedule Compliance Report


AlexZorkin:
**Description**  
Create a new Cypress test scenario for the compliance reporting feature. This scenario will simulate a supplier creating a new compliance report and adding one record for each schedule type. The schedules to be tested include: Fuel Supply, Fuel Export, Other Uses, Notional Transfer, Allocation Agreement, and Fuel Supply Equipment. Once all schedules are added, the test should navigate back to the summary page and verify that each schedule section is present and expandable.

**Task**  
- Develop a new test scenario using the existing Cypress+Cucumber framework.  
- Simulate supplier login and creation of a new compliance report.  
- For each schedule type (Fuel Supply, Fuel Export, Other Uses, Notional Transfer, Allocation Agreement, Fuel Supply Equipment):
  - Navigate to the corresponding page.
  - Enter a valid record.
  - Return to the compliance report summary.
- Validate that the summary page displays expandable sections for all schedule types added.

**Acceptance Criteria**  
- A new scenario is added to the compliance reporting test suite using Gherkin syntax.  
- The scenario covers creating a new compliance report and adding one record for each schedule type.  
- After adding all records, the summary page must display all schedule sections (Fuel Supply, Fuel Export, Other Uses, Notional Transfer, Allocation Agreement, Fuel Supply Equipment) correctly.  
- The test passes without errors and confirms that the UI reflects all schedule entries as expected.

**Example Scenario (Gherkin)**

```gherkin
Feature: Compliance Report Management

  Scenario: Supplier creates a compliance report with multiple schedule types
    Given the user is on the login page
    And the supplier logs in with valid credentials
    And they navigate to the compliance reports page
    When the supplier creates a new compliance report
    Then the compliance report introduction is shown

    # Add Fuel Supply record
    When the supplier navigates to the Fuel Supply page
    And the supplier enters a valid Fuel Supply record
    And the supplier returns to the compliance report summary

    # Add Fuel Export record
    When the supplier navigates to the Fuel Export page
    And the supplier enters a valid Fuel Export record
    And the supplier returns to the compliance report summary

    # Add Other Uses record
    When the supplier navigates to the Other Uses page
    And the supplier enters a valid Other Uses record
    And the supplier returns to the compliance report summary

    # Add Notional Transfer record
    When the supplier navigates to the Notional Transfer page
    And the supplier enters a valid Notional Transfer record
[...]

PR #2198, opened 2025-03-11
Feat: cypress tests fix and compliance reports coverage.

prv-proton:
Fixes: #2014, #2015

- Fix cypress failing test cases
- Correct credentials for lcfstest user.
- handle null email during authentication.


AlexZorkin (changes requested):
This looks like a good start, any reason to keep it in draft still?

backend/lcfs/web/api/email/services.py
AlexZorkin, line 1: Was this still breaking submissions if the ches config was misconfigured?

AlexZorkin (approved):
Well done on this cypress ticket, quite a lot of work. Hopefully we can have this up and running now. 👍

backend/lcfs/web/api/email/services.py
AlexZorkin: just one conflict to resolve.

PR #2213, opened 2025-03-12
Feat: Updated CR Address Requirements

AlexZorkin:
- Updated address logic so that serviceAddress is now the only required address in the compliance report view
- Updated test


prv-proton (approved):
Looks good 👍

PR #2214, opened 2025-03-13
Fix: Apply 'Select All' to Bulk Notification Actions - 2028

### LINKED ISSUE #2028: LCFS - Notification bug when "selecting all" and "deleting selected"
**Describe the Bug**
When a user selects "select all" and then "mark as read' this function works for **all pages** of notifications. When a user selects "select all" and the "delete selected" it only deletes the notifications on the page (so 10 lines at a time if the page size is 10, 25 at a time if it is a page size of 25, etc.). This happens for the IDIR and the BCeID side.

Image

**New Wireframe**  
[Notifications](https://preview.uxpin.com/e0e2f15fc1b52987fd0287ad8c8103452dd16d60#/pages/199531705?mode=dm)

**Expected Behaviour**
Once all "select all" is selected, this should apply to all notification lines, not just the ones on the page
- The red alert circle should disappear
- The notification count should reset to zero.

**Actual Behaviour**
- Only the notifications on the current page are being deleted, rather than all 
- The red alert circle remains visible on the bell icon, showing the notifications left on other pages

**Implications**
Users may think they still have new notifications, causing confusion.

**Steps To Reproduce**
User/Role: IDIR

1. Open the notifications panel.
2. Click "Select all" and then "Delete selected"
3. Observe that only the notifications on the current page are deleted the bell icon still displays a red alert with a number.
4. Refresh the page and check if the issue persists.

hamed-valiollahi:
This PR ensures consistency in bulk notification actions:

- Fixes "Select All" to apply across all pages when deleting notifications.
- Fixes "Select All" to mark all notifications as read, not just those on the current page.

Closes #2028


AlexZorkin (changes requested):
Looks good. we could use a test file for this notifications component though so please add one.

hamed-valiollahi:
> Looks good. we could use a test file for this notifications component though so please add one.

Thanks, Alex for the suggestion! I’ve added a test file for the Notifications component.

hamed-valiollahi:
> looks good ty, just the one merge conflict to resolve.

Sure, Thank you, Alex!

PR #2222, opened 2025-03-13
Fix: Role Dropdown Filter - 2188

AlexZorkin:
- Created wrapper component for role filter to fix react query caching issue with role dropdown options.


dhaselhan (approved):
LGTM 👍

PR #2230, opened 2025-03-14
feat: Add Analyst Adjustments

dhaselhan:
* Remove userType
* Update list reports to only return the current visible report in the chain


backend/lcfs/web/api/compliance_report/repo.py
```diff
@@ -354,44 +359,28 @@ async def add_compliance_report_history(
...
-                    == latest_versions.c.compliance_report_group_uuid
-                )
-                & (ComplianceReportListView.version == latest_versions.c.max_version),
-            )
-        else:
-            query = select(ComplianceReportListView).where(
-                cast(ComplianceReportListView.report_status, String)
-                != ComplianceReportStatusEnum.Draft.value
-            )
+        is_analyst = user_has_roles(user, [RoleEnum.ANALYST])
```
AlexZorkin, line 373: make sure we have some updated tests for this updated status exclusion logic

backend/lcfs/web/api/compliance_report/repo.py
```diff
@@ -965,3 +956,58 @@ async def get_previous_summary(
...
+                ComplianceReportStatusEnum.Assessed.value
+            )
+            max_version_subq = max_version_subq.where(
+                or_(
+                    ComplianceReport.current_status_id
+                    == assessed.compliance_report_status_id,
+                    or_(
+                        ComplianceReport.supplemental_initiator
+                        != SupplementalInitiatorType.GOVERNMENT_REASSESSMENT,
+                        ComplianceReport.supplemental_initiator.is_(None),
```
AlexZorkin, line 995: I don't think we would ever have a null supplemental_initiator? or has that logic changed in this work?

backend/lcfs/web/api/compliance_report/repo.py
```diff
@@ -965,3 +956,58 @@ async def get_previous_summary(
...
+                    ),
+                )
+            )
+
+        # Build the main query filtering out excluded statuses and ensuring that
+        # the report version equals the maximum visible version for that group.
+        query = select(ComplianceReportListView).where(
```
AlexZorkin, line 1002: I see now that this is where you're filtering on statuses based on the user. Nice refactor 👍

backend/lcfs/web/api/compliance_report/services.py
```diff
@@ -96,6 +99,72 @@ async def create_compliance_report(
 
         return ComplianceReportBaseSchema.model_validate(report)
 
+    @service_handler
+    async def created_analyst_adjustment_report(
```
AlexZorkin, line 103: created not create?

backend/lcfs/web/api/compliance_report/services.py
```diff
@@ -96,6 +99,72 @@ async def create_compliance_report(
...
+        self, existing_report_id: int, user: UserProfile
+    ) -> ComplianceReportBaseSchema:
+        """
+        Creates a new supplemental compliance report.
```
AlexZorkin, line 107: pls update comment to explain what an analyst adjustment supplemental is all about.

backend/lcfs/web/api/compliance_report/services.py
```diff
@@ -207,39 +275,72 @@ async def get_compliance_reports_paginated(
...
-                }
-                if report.current_status.status in statuses:
-                    (
-                        report.current_status.status,
-                        report.current_status.compliance_report_status_id,
-                    ) = (
-                        ComplianceReportStatusEnum.Submitted.value,
-                        None,
-                    )
+    def _mask_report_status(self, reports: List, user: UserProfile) -> List:
```
AlexZorkin, line 278: I like the role refactoring you've done in this pr. We just need to make sure we cover our bases with tests for this updated logic.

backend/lcfs/web/api/final_supply_equipment/importer.py
```diff
@@ -235,7 +235,7 @@ async def import_async(
                         try:
                             fse_data = _parse_row(row, compliance_report_id)
                             await fse_service.create_final_supply_equipment(
-                                fse_data, user
+                                fse_data, "FIXME"
```
AlexZorkin, line 238: Fix now or later?

frontend/src/utils/grid/cellRenderers.jsx
```diff
@@ -312,17 +312,18 @@ export const ReportsStatusRenderer = (props) => {
...
     'Draft',
     'Submitted',
+    'Analyst adjustment',
     'Recommended by analyst',
     'Recommended by manager',
     'Assessed',
-    'Reassessed',
     'Rejected'
   ]
   const statusColorArr = [
```
AlexZorkin, line 321: I'm not a fan of this index to index matching between these arrays. Having a direct mapping would make more sense.

AlexZorkin (changes requested):
You really went to town on this pr, nice work. Let’s hold off on merging to develop until after our next release branch snapshot, just so we don’t over complicate our next deploy to prod.

Some thoughts:
- Good call on removing user_type, simplifies our get_effective logic and lot of other lookups.
- Updating the user exclusion filtering for statuses is great, I just want to be sure we have every scenario covered in our tests. Please double check we have sufficient testing for this.
- In the future breaking up a massive pr like this into separate domain related prs would make it easier to review. Just for an example, cleaning up enums could be one, removing user_type could’ve been a second, followed up by the rest in a third.

Take a look through my comments and we’ll circle back on this one.

backend/lcfs/web/api/compliance_report/repo.py
dhaselhan: Supplemental Initiator is only set on supplementals currently. It is null for original reports.

backend/lcfs/web/api/allocation_agreement/repo.py
```diff
@@ -76,7 +69,7 @@ async def get_table_options(self, compliance_period: str) -> dict:
 
     @repo_handler
     async def get_allocation_agreements(
-        self, compliance_report_id: int, exclude_draft_reports: bool = False
```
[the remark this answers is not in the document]
dhaselhan: We hide the entire report, gov users are unable to see / view reports in Draft status. In the case of draft supplemental, the view should not have the draft only the ID of the previous report which they open. It uses the ID/version of that report to only fetch records from that version and previous versions. We don't handle Draft records separately, we only look at if they are attached to a draft report.

frontend/src/utils/grid/cellRenderers.jsx
dhaselhan: Converted both this and the one for transfers to use enums and a map.

backend/lcfs/web/api/final_supply_equipment/importer.py
dhaselhan: Good catch, fixed now

backend/lcfs/db/migrations/versions/2025-03-11-21-13_bd4da1540e2e.py
```diff
@@ -0,0 +1,157 @@
...
+        "compliance_report_status"("compliance_report_status_id", "status",
+                                   "effective_status")
+        VALUES(10, 'Analyst_adjustment', 't');
+        """
+    )
+
+    op.execute(
+        """
+        DELETE FROM "public".
+        "compliance_report_status" WHERE "compliance_report_status_id" = 6
```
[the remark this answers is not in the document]
dhaselhan: All of the statuses were inserted with hard-coded IDs, they do not use the sequence / default of the table.

backend/lcfs/web/api/compliance_report/services.py
dhaselhan: Fixed

backend/lcfs/web/api/compliance_report/services.py
dhaselhan: Update the comments

frontend/src/utils/grid/cellRenderers.jsx
AlexZorkin: awesome thank you.

AlexZorkin (approved):
Thanks for the updates. Once conflicts are resolved this is good to go.

PR #2232, opened 2025-03-14
LCFS-2224: End Use is blank in Export Fuels

areyeslo:
Removing duplicates from backend.


AlexZorkin (changes requested):
Looks good but just want to confirm we have tests that cover this updated logic. Please double check we have a few tests that ensure the correct end use is returned.

areyeslo:
This is not a new logic, this was the existing logic before this [commit](https://github.com/bcgov/lcfs/commit/7ee1ee3ba40bb424cdc216c00318f669c1ecc668#diff-b0b09f63f27dca31c560399c27c8cb8cce46b95a0c2159562736bf9ac0cb3ca7)

The mentioned commit de-duplicated the records in the front end, but it was still showing issues. I retrieved the original logic and worked on the backend to avoid duplicates. Duplicates should be solved better in the backend instead of the frontend. This changes is just rolling back the de-duplication in the front end and  adding the fix in backend.

AlexZorkin (approved):
Looks good ty 👍

PR #2234, opened 2025-03-14
LCFS-2226 Export fuel schedule rci

areyeslo:
- Not cleaning correctly the fuel code for Fuel exports
- Migration to remove fuel code in existing records having  Default Carbon Intensity


AlexZorkin (changes requested):
I appreciate the backup table but I don't think we need it in this case. Depending on how many records have this issue it's probably best to just update this manually so we can avoid having a backup table persisting forever in our prod db. How many records did you find with this issue?

areyeslo:
Not many records, I am removing the creation of the backup table. 
Thinking we can avoid having data lost in migrations through creating a backup table that can be deleted once the change is confirmed in Production.

AlexZorkin (approved):
Looks good 👍

PR #2236, opened 2025-03-17
LCFS-2169: Metabase - Renewable fuel requirement queries/dashboards

areyeslo:
This update introduces latitude and longitude for fuel codes associated with fuel production facilities.
In Metabase, Where is this fuel coming from? (will need to use fuel code production facility information)

Added a location table to store unique records for production facilities.
Ensured unique entries to prevent duplicate location data.
Allow fuel production facility locations to be visualized on Metabase maps.

### DIFF
--- a/etl/geocode_locations.sh
+++ b/etl/geocode_locations.sh
@@ -1,0 +1,136 @@
+#!/bin/bash
+set -e
+
+# Usage:
+#   For local: ./geocode_locations.sh local <local_container>
+#   For remote (dev, test, prod): ./geocode_locations.sh <environment>
+#       where <environment> is one of 'dev', 'test', or 'prod'
+
+if [ "$#" -lt 1 ]; then
+    echo "Usage:"
+    echo "  For local: $0 local <local_container>"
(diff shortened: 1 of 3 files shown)


AlexZorkin (commented):
This is a very useful script! Are we assuming that location will become a table in our database?

areyeslo:
Yes, it is the creation of Location table. This script will take the city, state/province, country from fuel_code and provide latitute and longitude for better visualization in Metabase. Instead of connecting to openstreetmap, we would have a Location table storing use unique locations with its latitude and longitude and only if the place is new then we send a request to retrieve latitude and longitude from openstreetmap. We can apply this to other parts of the code to avoid connecting to openstreetmap looking for the same coordinates.

etl/geocode_locations.sh
```diff
@@ -0,0 +1,136 @@
+#!/bin/bash
+set -e
```
prv-proton, line 2: how frequently are we planning to run this script?

areyeslo: If a new location is created in fuel_code table.

AlexZorkin (approved):
Thank you for the table. We will create followup tickets to create a geo code service for locations.

PR #2241, opened 2025-03-17
Fix: Transfer Available Balance - 2227

AlexZorkin:
- updated in reserve balance queries to sum negative transactions only
- updated transfer view logic around available balance
- tests


prv-proton (approved):
Looks good 👍

PR #2243, opened 2025-03-17
LCFS-2219: PROD Page not loading for viewing a rescinded transaction

areyeslo:
When retriving an individual transaction check if it was `Submitted` before being `Rescinded`

### DIFF
--- a/backend/lcfs/web/api/transfer/services.py
+++ b/backend/lcfs/web/api/transfer/services.py
@@ -99,8 +99,16 @@ class TransferServices:
             in [
                 TransferStatusEnum.Draft,
                 TransferStatusEnum.Sent,
-                TransferStatusEnum.Rescinded,
             ]
+            or
+            (
+                transfer.current_status.status == TransferStatusEnum.Rescinded and
+                not any(
+                    history.transfer_status.status == TransferStatusEnum.Submitted
+                    and history.create_date < transfer.update_date
+                    for history in transfer.transfer_history
+                )
+            )
         ):
             raise DataNotFoundException(f"Transfer with ID {transfer_id} not found")
 


backend/lcfs/web/api/transfer/services.py
```diff
@@ -99,8 +99,16 @@ async def get_transfer(self, transfer_id: int) -> TransferSchema:
             in [
                 TransferStatusEnum.Draft,
                 TransferStatusEnum.Sent,
-                TransferStatusEnum.Rescinded,
             ]
+            or
```
AlexZorkin, line 103: What was the actual issue with the page not loading here? I'm not sure I follow what this query change is supposed to do.

AlexZorkin (changes requested):
just one question to address.

backend/lcfs/web/api/transfer/services.py
areyeslo: In the current code, when accessing a transfer with GET request to the endpoint pi/transfers/:id we throw `404` if the transfer is under the status `Draft`, `Sent` and `Rescinded`. The IDIR users are not able to access Transfers under `Rescinded` even though they appear in the list of transfers.

When loading /transactions page for IDIR users and calling POST request to endpoint api/transactions/, we have a logic in place that we are showing `Rescinded` transactions as long as they were `Sent` to the Government. Otherwise, we do not show `Rescinded` transactions. The issue is when the IDIR user wants to see these `Rescinded` transactions individually, we are sending `404`  because we do not have the logic.

Here we have two options to consider:
- Just remove `Rescinded` transfers from the filter  and then when the IDIR user selects the `Rescinded` transfer from the AG Grid, the invidual transfer will be displayed.
- Add the logic included in the AG Grid, to check if the transfer has been `Sent` before being `Rescinded`. If this is the case display the transfer instead of throwing `404`.

If we decide to just remove `Rescinded`, the IDIR user can access through manually modifying the URL to access individual `Rescinded` transfers that were not `Sent`. The second option provides protection for this case. The normal flow is that the IDIR user selects from the available transfers from the AG Grid and clicks the record to access individually, but the implementation in this PR is thought to offer this protection.

AlexZorkin (approved):
👍

PR #2250, opened 2025-03-18
fix: Switch all exports to use XLSX

dhaselhan:
* Most recent file format and it fixes the issues


backend/lcfs/tests/test_users.py
```diff
@@ -19,10 +19,14 @@ async def test_export_success(
     response = await client.get(url)
 
     assert response.status_code == status.HTTP_200_OK
-    assert response.headers["Content-Type"] == "application/vnd.ms-excel"
+    print(response.headers["Content-Type"])
```
AlexZorkin, line 22: remove print

AlexZorkin (approved):
just one print statement to remove 👍

Issue #2251, opened 2025-03-18
LCFS - BUG - Export fuel schedule is not updating the RCI


airinggov:
**Describe the Bug:**  
The Export fuel schedule is not updating the RCI between selecting the Default CI or a fuel code. 

**Expected Behaviour:**  
- When selecting a Fuel Code, the Export fuel schedule should update the RCI based on the selected fuel code.  
- The behavior should be consistent with the Fuel Supply schedule which works properly  
- The RCI should correctly reflect the value assigned to the selected fuel code.  

**Actual Behaviour:**  
- Switching between Default and Fuel Code does not update the RCI in the Export Fuel Schedule.  
- The RCI remains unchanged even when a different fuel code is selected.  
- The Fuel Supply Schedule does not have this issue, implying an inconsistency between the two schedules.  

**Implications:**  
- Incorrect carbon intensity values will lead to compliance miscalculations.  

**Steps To Reproduce:**  
1. Navigate to the Export Fuel Schedule.  
2. Select **Alternative Jet Fuel** as the fuel type.  
3. Toggle between **Default** and **Fuel Code** options.  
4. Observe that the RCI value remains unchanged.  
5. Compare with the Fuel Supply Schedule, where the RCI updates correctly.  

**Notes:**  
- Ensure the RCI updates dynamically when selecting a fuel code.  
- Verify consistency between Export Fuel Schedule and Fuel Supply Schedule.  
- Test this fix across all fuel types that have RCI dependencies.

PR #2252, opened 2025-03-18
fix: Rely on backend

dhaselhan:
* Don't try and use complex front end logic, rely on backend value to be correct

Fixes: #2251


AlexZorkin (changes requested):
Looks good but please point this directly to our release branch.

AlexZorkin (approved):
👍

PR #2261, opened 2025-03-19
feat: Add Togglable Changelogs

dhaselhan:
* Add new toggle component
* Use toggle to switch between summary and change log
* Add hover back to edit icons


kevin-hashimoto (approved):
Looks great! Liking how we simplify and reusing a good chunk of our previous code!

Issue #2205, opened 2025-03-12
LCFS - Allow a BCeID user to delete a draft supplemental report


airinggov:
**Description:**  
Allow a BCeID user to delete a draft supplemental report. When deleted, any changes made in the draft should be removed, and the report should revert to its original state. All data associated with the supplemental report should be permanently deleted from the database.

**Wireframe:**  
N/A

**Purpose and benefit to user:**  
This feature allows BCeID users to manage their supplemental reports efficiently by removing unnecessary drafts, ensuring data integrity, and preventing clutter in the system.

**Acceptance Criteria:**  
- [ ] A BCeID user can delete a supplemental report if it is in "Draft" status.
- [ ] Upon deletion, all changes made in the draft are removed.
- [ ] The base report remains unchanged.
- [ ] Any data specifically created for the supplemental report is fully removed from the database.
- [ ] The user receives a modal before deletion: "Do you want to delete this supplemental report?" Cancel | Delete supplemental report
- [ ] A success message is displayed once the report is deleted.

**Development Checklist:**  
- [ ] Implement a "Delete supplemental report" button that is only available when the supplemental report is in "Draft" status.
- [ ] Ensure all associated data is removed from the database upon deletion.
- [ ] Implement a confirmation modal before deletion.
- [ ] Ensure that deleting the draft does not impact the original report.

airinggov:
It appears to work very well. Only one issue found so far. The original report didn't have any FSE data. I added one in the draft supplemental then deleted the report. The FSE is still appearing and showing a duplicate row. If I click to create a new supplemental report the FSE schedule disappears (as it should be really).

Grulin:
Can't test as I can't seem to make a compliance report in DEV without getting an "internal server" message

PR #2263, opened 2025-03-20
Fix: To clear caching from react queries upon supplemental report delete

prv-proton:
Bug fix: #2205
https://github.com/bcgov/lcfs/issues/2205#issuecomment-2734823888


dhaselhan (approved):
LGTM 👍

PR #2266, opened 2025-03-20
feat: Delete transaction when returning to supplier

dhaselhan:
* Add new delete function + test for it


backend/lcfs/web/api/transaction/repo.py
```diff
@@ -489,3 +502,15 @@ async def get_transaction_start_year(self):
         )
 
         return oldest_year
+
+    @repo_handler
+    async def delete_transaction(self, transaction_id, attached_report_id):
```
prv-proton, line 507: Don't we need to delete this transaction from Transaction table? as it could still be in RESERVE'd transaction for organization balance calculation.

AlexZorkin (approved):
Looks good 👍

PR #2267, opened 2025-03-20
LCFS-2220: Retain data from lines 6, 7, 8, and 9

areyeslo:
Copy over summary lines from `Assessed` report in the same period of supplemental report
Ensure updated value in summary lines is loaded instead of cache value from React Query.


backend/lcfs/web/api/compliance_report/services.py
```diff
@@ -142,6 +142,22 @@ async def create_supplemental_report(
...
+            current_report.organization_id,
+            int(current_report.compliance_period.description)
+        )
+        if not assessed_report or not assessed_report.summary:
+            raise DataNotFoundException("Assessed report summary not found for the same period")
+
+        # Copy over the summary lines from the assessed report.
+        summary_data = {
+            column: getattr(assessed_report.summary, column)
+            for column in assessed_report.summary.__table__.columns.keys()
```
AlexZorkin, line 156: I don't think we want to bring across all line values, just 6-9.

AlexZorkin (changes requested):
Just double checking the requirements. ty

AlexZorkin (approved):
👍

PR #2274, opened 2025-03-21
feat: Move met/not met statements into history

dhaselhan:
* Show for each report instead of just the final result
* Remove formatting inside the text
* Return summary lines needed as part of the report chain


hamed-valiollahi (approved):
Looks great! 👍  Thanks, Daniel!

PR #2296, opened 2025-03-25
LCFS-2217: Deleted schedule display

areyeslo:
On Supplemental report when the user removes the records in a schedule,  the changelog toggle should be visible to view deleted rows.

image


prv-proton (approved):
Looks good 👍

PR #2301, opened 2025-03-26
fix: DateEditor Calendar Icon Clickability

### LINKED ISSUE #2240: LCFS - BUG - Fix DateEditor Calendar Icon Clickability in AG Grid
## Description
The calendar icon in AG Grid's custom DateEditor component remains unresponsive under normal conditions. Activation occurs only when the date field displays an invalid date, such as one outside the permitted min/max range. This inconsistent behavior diminishes the user experience by not aligning with expectations for immediate interaction.

This issue is related to the 'stopEditingWhenCellsLoseFocus' flag on BCGridEditor

## Expected Behaviour
Users must access the date picker by clicking the calendar icon at all times, irrespective of the validity of the date currently displayed.

## Actual Behaviour
The calendar icon only responds when the date field contains an invalid date. When the date is valid, clicking the icon produces no effect. Notably, the X (clear) button within the text field functions properly.

## Implications
- Users face frustration due to the visible yet non-functional calendar icon.
- The inconsistent behavior breeds confusion.
- Users may be forced to manually enter dates or intentionally input an invalid date to trigger the calendar.
- The issue might lead to the avoidance of date fields or the entry of incorrect data.

## Steps To Reproduce
1. Navigate to the Final Supply Equipments or Fuel Exports page.
2. Click 'Add' to create a new entry.
3. Select any date field to begin editing.
4. Attempt to click the calendar icon to open the date picker.
5. Observe the unresponsiveness of the calendar icon.
6. Enter an invalid date (e.g., one that violates the min/max range).
7. Click the calendar icon again, and observe that it now functions.

## Technical Details
- The issue appears linked to conflicts between AG Grid's cell editing lifecycle and Material UI DatePicker's open/close mechanisms.
- Event propagation within the DateEditor component disrupts proper click handling.
- The custom DateEditor implementation in `/frontend/src/components/BCDataGrid/components/Editors/DateEditor.jsx` requires redesign.
- Consider replacing Material UI's DatePicker with a custom solution using TextField, IconButton, Popover, and StaticDatePicker.

kevin-hashimoto:
- fix: DateEditor Calendar Icon Clickability, closes #2240


dhaselhan:
With these fixes are we able to add the stopEditingOnCellLosesFocus flag back on for Fuel Export?

kevin-hashimoto:
> With these fixes are we able to add the stopEditingOnCellLosesFocus flag back on for Fuel Export?

I couldnt get the datepicker working nicely with `stopEditingOnCellLosesFocus` so I ended up exposing the date picker events to manually trigger the stop editing event

dhaselhan:
> > With these fixes are we able to add the stopEditingOnCellLosesFocus flag back on for Fuel Export?
> 
> I couldnt get the datepicker working nicely with `stopEditingOnCellLosesFocus` so I ended up exposing the date picker events to manually trigger the stop editing event

Do we need to update the other editors to do this so the user experience is consistent?

kevin-hashimoto:
implemented the same changes to Fuel codes grid schema

kevin-hashimoto:
@AlexZorkin what is the 'click to save' feature, im not seeing that in the ticket? are we moving away from auto saving rows after losing focus?

For this ticket, the expected behaviour should be that clicking on the calendar icon opens the calendar. using the `stopEditingWhenCellsLoseFocus` causes some weird focus issues that prevents the calendar icon from opening the calendar. Not using that flag and manually triggering the lost focus event on the calendar retains he auto focus behaviour while fixing the calendar issue.

AlexZorkin:
> @AlexZorkin what is the 'click to save' feature, im not seeing that in the ticket? are we moving away from auto saving rows after losing focus?
> 
> For this ticket, the expected behaviour should be that clicking on the calendar icon opens the calendar. using the `stopEditingWhenCellsLoseFocus` causes some weird focus issues that prevents the calendar icon from opening the calendar. Not using that flag and manually triggering the lost focus event on the calendar retains he auto focus behaviour while fixing the calendar issue.

The Date Editors work but without the stopEditingWhenCellsLoseFocus prop, the other field types no longer have the save on focus lost feature. Ideally we make each field type work consistently and save when the user clicks outside of the row. For example on the FSE schedule input view, the date fields save on click of a date, but entering a kWh usage value and then clicking outside of the row does not.

AlexZorkin:
After chatting about this feature I think we're going to look at rebuilding a date picker that uses the ag-grid one, so we can avoid this clashing of events. we can follow up in a separate ticket.

PR #2307, opened 2025-03-26
feat: Add Fuel Supply Cypress Tests

dhaselhan:
* upgrade cypress


frontend/cypress.config.js
```diff
@@ -61,10 +62,10 @@ export default defineConfig({
         clearComplianceReports() {
           return new Promise((resolve, reject) => {
             const client = new Client({
-              user: process.env.DB_CYPRESS_USER,
+              user: process.env.DB_CYPRESS_USER || 'lcfs',
```
prv-proton, line 65: Just to confirm, won't this raise security comment from Code scan in future PR's? This was causing issues to me when I raised the PR so.

frontend/cypress/e2e/Pages/ComplianceReport/ComplianceReport.test.js
```diff
@@ -261,9 +204,6 @@ When('the supplier starts entering data into each schedules', (dataTable) => {
         // Save before refreshing
         cy.waitAndClick(SELECTORS.saveButton)
         cy.wait(1000)
```
prv-proton, line 206: sometimes when entering too many rows this was causing to refresh the page in the Cypress Daily run workflow. So introduced manual refresh before entering the data, we need to handle this case if not handled already.

prv-proton (approved):
Looks good 👍

frontend/cypress/e2e/Pages/ComplianceReport/ComplianceReport.test.js
dhaselhan: Thanks for the info, I've added the refresh back

frontend/cypress.config.js
dhaselhan: I'm not sure for future PRs, I did have to flag it here as ignore.

PR #2310, opened 2025-03-27
Fix: Add custom validation for fuelSuffix format in fuel code - 2102

### LINKED ISSUE #2102: Convert fuel_suffix to Numeric(4,1)
Currently fuel_suffix is text, however it will always be 3 digits followed by a single decimal. Lets update our Postgres types to match so we don't get any future bugs.

* Update fuel_suffix on fuel_code table to Numeric(4,1) and update existing values as necessary
* Update front end to receive and format number correctly
* Update schema to use float

hamed-valiollahi:
This PR introduces a new backend validator to enforce fuelSuffix format (must match 'XXX.X', e.g. '102.5')

Closes #2102


frontend/src/assets/locales/en/fuelCode.json
```diff
@@ -26,7 +26,7 @@
     "fuelCodeId": "Fuel Code ID",
     "status": "Status",
     "prefix": "Prefix",
-    "fuelSuffix": "Iteration",
+    "fuelSuffix": "Fuel code",
```
dhaselhan, line 29: Why are we changing the column label?

dhaselhan (changes requested):

frontend/src/assets/locales/en/fuelCode.json
hamed-valiollahi: Good Question!

I changed the label because the validation error was confusing. We had "Fuel Code" as the column label, but the validation message said:
> Unable to save row: **Iteration** format is invalid. Must be like '102.5'.

That didn’t match the column label.

Another reason is that its only usage is in the "Fuel codes" table, where the "Fuel code" column is displayed as "Iteration".

image

But we do display the "Fuel code" column correctly in Add/Edit, as shown in the screenshot below:

image

I wasn't sure if I should remove that and use "fuelCode" from the locale file instead. Should I?

hamed-valiollahi:
> Not sure which label is correct but I trust ya LGTM 👍

Thanks, Daniel, for pointing that out! I confirmed with Al that "Iteration" should be used for both cases. The Iteration refers to the number, and the Fuel code is the combination of the prefix and number. I've updated the code. Appreciate the review!

PR #2322, opened 2025-03-27
feat: Reduce API calls on saving fuel supply

dhaselhan:
* Flatten schema to better mirror row data to require less front end modification
* Use returned updated data from save call instead of re-fetching


AlexZorkin (approved):
Nice optimization 👍

PR #2324, opened 2025-03-27
Feat: GHGenius support for pre-2024 supplementals

AlexZorkin:
# GHGenius Provision Logic Update

## Description
Updated the fuel supply table options logic to handle GHGenius provision (ID: 8) based on compliance period and fuel type. This change ensures that:

1. For non-fossil-derived fuels:
   - Pre-2024: Add GHGenius provision (ID: 8)
   - 2024 and later: Show all provisions except the original provision (ID: 1) and GHGenius

2. For fossil-derived fuels (e.g., Petroleum Gasoline, Diesel):
   - Always show only the original provision (ID: 1), regardless of year


dhaselhan (approved):
LGTM 👍

PR #2326, opened 2025-03-28
Upgrade Front End to MUI 6

dhaselhan:
* Change Icons to no longer have Icon at the end
* Unstable Grid2 was changed to be stable


AlexZorkin (approved):
Thanks for going through this one. 👍

PR #2330, opened 2025-03-28
feat: Add missing column to AA export

dhaselhan:
* Cap number of formatted rows to length of data unless is_template is set to True


AlexZorkin (approved):
👍

dhaselhan:
Too many conflicts, going to make a new MR instead

PR #2337, opened 2025-03-28
Fix: Update Summary ETL

AlexZorkin:
- updated compliance report summary etl script to add additional fields including penalty information


etl/nifi_scripts/compliance_summary_update.groovy
```diff
@@ -0,0 +1,284 @@
...
+            def line11Diesel  = new BigDecimal(summaryJson.summary.lines."22")
+            // ------------------------------
+            // Jet Fuel: No TFRS data, set to 0
+            // ------------------------------
+            def jetFuelDefault = 0.0
+            // ------------------------------
+            // Low Carbon Fuel Requirement Summary
+            // ------------------------------
+            def lowCarbonSurplus      = new BigDecimal(summaryJson.summary.lines."25")
+            def bankedUsed            = new BigDecimal(summaryJson.summary.lines."26")
```
prv-proton, line 170: there is a mapping for legacy lines like 26a, 26b, 26c already placed in summary_service component. Aren't we going to follow this mapping?

AlexZorkin: You're right, i will review those specific fields we added in LCFS. I think we may need to a few additional fields here as well so i'll take another look.

dhaselhan (approved):
LGTM 👍

PR #2342, opened 2025-03-31
Fix: Disable Export Date & Handle Unknown Save Error - 2215

### LINKED ISSUE #2215: LCFS - Export fuel schedule updates
**Description:**  
Enhance the Export fuel schedule by modifying the placement and behavior of the ‘Export date’ field, updating the ‘Determining CI’ options, and implementing an alternate RCI lookup when necessary.

**Wireframe:**  
https://preview.uxpin.com/da68049c0af15edd7ce4e47e4bee431fed40c30d#/pages/201433209

These changes enhance the Export fuel schedule by improving more accurate carbon intensity (CI) calculations.

**Acceptance Criteria:**  
- [x] Move the ‘Export date’ field to appear after the ‘Fuel code’ select list.
- [x] Disable the ‘Export date’ field by default.
- [x] Make Export date optional unless Unknown is selected from Determining CI, then it is required.
- [x] Add ‘Unknown’ as an option in the ‘Determining CI’ dropdown.
- [x] If ‘Unknown’ is selected:
  - [x] The ‘Fuel code’ field remains disabled.
  - [x] The ‘Export date’ field is enabled.
  - [x] Entering an ‘Export date’ triggers an alternate RCI lookup.
- [x] The RCI lookup queries all active fuel codes for the selected fuel from the 12 months prior to the Export date.
- [ ] The system selects the lowest CI value from the queried results and applies it.
- [ ] Validate that all changes integrate seamlessly into the existing Export Fuel Schedule workflow.

**Development Checklist:**  
- [ ] Adjust UI placement of the ‘Export date’ field.
- [ ] Adjust database so the Export date is optional
- [ ] Implement logic to disable/enable fields based on ‘Determining CI’ selection.
- [ ] Configure the alternate RCI lookup query.
- [ ] Ensure the lowest CI value is correctly applied when ‘Unknown’ is selected.

**Notes:**  
- Ensure error handling is in place for scenarios where no CI values are found within the 12-month range.

hamed-valiollahi:
- Keeps Export Date field disabled unless "Unknown" is selected in the provision.
- Fixes 500 error caused by missing CI when no fuel codes are found in the past 12 months.

Closes #2215


frontend/src/views/FuelExports/_schema.jsx
AlexZorkin, line 1: Would it be possible to handle the getDefaultCi on the backend?

AlexZorkin (changes requested):
Looking good, just one question, and please add frontend updated tests that cover the different cases of getDefaultCi.

hamed-valiollahi:
> Looking good, just one question, and please add frontend updated tests that cover the different cases of getDefaultCi.

For sure! I’ll add the tests I missed. Thanks for pointing that out!

frontend/src/views/FuelExports/_schema.jsx
hamed-valiollahi: For sure! I just wanted to stick with what we currently have. We can definitely do it.

The only concern is that if we add another API, we’d need to call it every time the user selects an option, which could affect performance. Another one is having redundant code.

Please let me know what you think.

AlexZorkin: Ok np keeping the same pattern on the frontend is fine.

AlexZorkin (approved):
looks good 👍

PR #2352, opened 2025-03-31
fix: Update clearFilter to use clearFilters

dhaselhan:
* Previously was a mix of clearFilters and resetGrid, lets use clearFilters and remove resetGrid
* Move AuditLog into BCGridViewer
* Update BCGridViewer to only use gridRef instead of ref and gridRef


frontend/src/components/BCDataGrid/BCGridViewer.jsx
```diff
@@ -135,17 +107,17 @@ export const BCGridViewer = forwardRef(
...
         onPaginationChange({ filters: filterArr })
         sessionStorage.setItem(`${gridKey}-filter`, JSON.stringify(gridFilters))
       },
-      [gridKey, onPaginationChange, initialPaginationOptions.filters]
+      [gridKey, onPaginationChange, paginationOptions.filters]
     )
 
     const handleSortChanged = useCallback(() => {
-      const sortTemp = ref.current?.api
+      const sortTemp = gridRef.current?.api
```
AlexZorkin, line 120: nice cleanup to remove the double refs

AlexZorkin (approved):
Nice cleanup ty 👍

PR #2360, opened 2025-03-31
fix: Only show adjustment banner if total was adjusted

dhaselhan:
* Use on change events to set true/false
* Spent an hour trying to get the input events to fire in the tests but it didnt work, so comment out the broken parts


hamed-valiollahi (approved):
Looks good to me! Thanks

PR #2368, opened 2025-03-31
fix: Set cache rather than invalidate it

dhaselhan:
* Invalidating triggers a re-query, which will 403 once its in DRAFT mode


AlexZorkin (approved):
👍

PR #2378, opened 2025-04-02
feat: Clear filters before redirecting to Compliance Reports

dhaselhan:
* Clear saved filters when submitting a report


AlexZorkin (approved):
👍

PR #2385, opened 2025-04-03
feat: Expose all Transfer Comments

dhaselhan:
* Update the MV to expose all 3 types of comments
* Add save call to update_category so the changes are persisted


AlexZorkin (approved):
Looks good, we should confirm there are no more request.user use cases left in our services. 👍

PR #2390, opened 2025-04-04
fix: Remove custom Grid2 styling

dhaselhan:
* Grid2 only supports row layout now so we cant use column
* Update grid values to always add to 12


AlexZorkin (approved):
Looks good 👍

PR #2393, opened 2025-04-04
fix: Don't continually add line 15/16

dhaselhan:
* Supplementals should copy over line 18 into 15 and line 19 into 16 and shouldnt be adding the previous numbers


AlexZorkin (approved):
test needed or all good? 👍

dhaselhan:
> test needed or all good? 👍

I added more tests in the quarterly MR but definitely.

Issue #2351, opened 2025-03-31
LCFS - Disable the Overwrite data option


airinggov:
**Description:**
Disable the 'Overwrite data' option in the Excel upload feature for Allocation Agreements and Final Supply Equipment (FSE) schedules within Supplemental Reports and Early Issuance Reports (Quarter 2 and beyond), when those schedules already contain existing data.

Overwrite option is only available on version 0 of initial report.

**Wireframe:**
N/A

**Purpose and benefit to user:**
Overwrite data would break the changelog feature

**Acceptance Criteria:**
- [ ] In Supplemental Reports, the 'Overwrite data' select option is removed for Allocation Agreements and FSE schedules if the schedule already contains data.
- [ ] In Early Issuance Reports (Quarter 2 and later), the same disabling logic applies to Allocation Agreements and FSE schedules.
- [ ] If the schedule is empty, the 'Overwrite data' option remains enabled and usable.

**Development Checklist:**
- [ ] Implement conditional logic to detect existing data in the relevant schedules.
- [ ] Remove the 'Overwrite data' select option when conditions are met.
- [ ] Ensure logic is scoped only to Supplemental and Early Issuance (Q2+) reports.

**Notes:**
- Consider if we need to prevent duplicate data, FSE no duplicate Serial number?

PR #2394, opened 2025-04-07
Feat: Excel upload for Allocation Agreements, FSE fixes, and overwrite logic - 2273, 2351

### LINKED ISSUE #2273: LCFS - Add upload Excel feature to the "Allocation agreement" section of the compliance report
**Description:**
Similar to the FSE feature already implemented, develop an upload utility that allows users to parse Excel template data into the "Allocation agreement" table of a compliance report. Users should have the option to either replace the entire record set or update only specific changes. Any data issues that prevent a line from being inserted should generate error messages, while valid data should be successfully saved.

**Wireframe:**
[Upload Excel function for Allocation agreements](https://preview.uxpin.com/e0e2f15fc1b52987fd0287ad8c8103452dd16d60#/pages/202456432?mode=dm)

**Purpose and Benefit to User:**
Enables efficient bulk data management by allowing users to update or replace allocation records via Excel.
Provides clear feedback on any errors encountered during the upload process.
Streamlines workflows for organizations managing large volumes of allocation row data.

**Acceptance Criteria:**
- Users can upload an Excel file to update Allocation agreement data.
- Users can choose between replacing the entire record set or adding only changes.
- Validation errors prevent incorrect lines from being inserted and display clear error messages (similar to the FSE download Excel function).
- Valid records successfully pass validation and are saved to the database.
- Upload results provide a summary of successful entries and errors within the upload modal.

**Development Checklist:**

Implement an API endpoint for Excel file uploads for the Allocation agreement table.
Parse and validate data against system requirements.
Implement logic for replacing or updating records based on user selection.
Validate and filter out incorrect data, logging errors for user feedback.
Provide an upload summary indicating success and failure reasons.
Perform extensive testing with various data scenarios.

**Notes:**
Ensure the process is optimized for large file uploads without performance issues

hamed-valiollahi:
This PR includes the following updates:

**Allocation Agreement**
- Excel import/export feature
- Conditional overwrite logic

**FSE**
- Fixed import/export issues specially in Government Adjustment
- Conditional overwrite logic

**Other fixes**
- Fixed "Select All" button issue in Notifications
- Fixed user deletion functionality

Closes #2273 & #2351


backend/lcfs/web/api/allocation_agreement/importer.py
```diff
@@ -0,0 +1,472 @@
...
+        fuel_type=fuel_type,
+        fuel_type_other=fuel_type_other or None,
+        fuel_category=fuel_category,
+        provision_of_the_act=provision,
+        fuel_code=fuel_code or None,
+        quantity=quantity,
+    )
+
+
+async def _update_progress(
```
dhaselhan, line 451: Future thinking: This looks like a copy from the FSE importer, we should think about adding an JobService that handles the common redis functionality.

backend/lcfs/web/api/allocation_agreement/repo.py
```diff
@@ -317,3 +323,80 @@ async def get_latest_allocation_agreement_by_group_uuid(
 
         result = await self.db.execute(query)
         return result.unique().scalars().first()
+
+    async def delete_all_for_report(
```
dhaselhan, line 327: issue: It looks like this is only inserting DELETE records, but it needs to handle the case where we are in the same version it should hard delete the rows.

backend/lcfs/web/api/allocation_agreement/services.py
```diff
@@ -98,6 +106,18 @@ async def convert_to_model(
             fuel_code_id=fuel_code_id
         )
 
+        # Now recalc and assign ci_of_fuel and units
```
dhaselhan, line 109: question: I'm a bit concerned with a mapper/conversion function performing a re-calculation. Should we look into why the units are incorrect and need to be re-calculated?

backend/lcfs/web/api/allocation_agreement/services.py
```diff
@@ -411,6 +431,23 @@ async def delete_allocation_agreement(
...
+
+        Args:
+            compliance_report_id: The ID of the compliance report whose allocation agreements will be deleted.
+            current_user: The username of the user performing the deletion.
+
+        Returns:
+            None
+        """
+
+        # todo: check if version 0
```
dhaselhan, line 447: This TODO needs to be addressed

backend/lcfs/web/api/allocation_agreement/services.py
```diff
@@ -427,3 +464,49 @@ async def get_compliance_report_by_id(self, compliance_report_id: int):
             )
 
         return compliance_report
+
+    async def calculate_ci_of_fuel(
```
dhaselhan, line 468: question: Should this also be used in update_allocation_agreement where we set the ci_of_fuel when saving the rows? I feel like they should use the same logic.

backend/lcfs/web/api/allocation_agreement/views.py
```diff
@@ -189,3 +193,153 @@ async def search_table_options_strings(
...
+        )
+
+    await validate.validate_organization_access(compliance_report_id)
+
+    # Check if overwrite is allowed
+    compliance_report = await compliance_report_services.get_compliance_report_by_id(
+        report_id=compliance_report_id,
+        user=request.user,
+    )
+    version = compliance_report.version
```
dhaselhan, line 280: nitpick: version is only used once so I would just write:
```is_original = compliance_report.version == 0```

File: frontend/src/assets/locales/en/common.json [hunk 7]
```diff
@@ -53,5 +53,44 @@
   },
   "ClearFilters": "Clear filters",
   "changelogCurrentState": "Current changes",
-  "exitChangeLog": "Exit change log"
-}
\ No newline at end of file
+  "exitChangeLog": "Exit change log",
+  "importExport": {
```
dhaselhan, line 57: Nice clean up / re-use!

dhaselhan (changes requested):

backend/lcfs/web/api/allocation_agreement/services.py
hamed-valiollahi: Good question! In the AG Grid, users don’t select the unit, it’s auto-selected based on the fuel type. Since the importer uses this logic to determine the unit during import, I needed to recalculate the units here as well.

hamed-valiollahi:
@AlexZorkin Hi Alex, could you please review this PR and let me know if Dan’s expectations are met? Also, let me know if any changes are needed.

AlexZorkin (approved):
Nice clean up 👍

hamed-valiollahi:
Thanks for the review!

PR #2396, opened 2025-04-07
Feat: changelog view

### LINKED ISSUE #2216: LCFS - Changelog view should display every version of a schedule
**Description:**  
Enhance the Changelog view to display every version of a schedule from the original report through all supplemental reports, reassessments, and government adjustments up to the most current version, following the provided wireframe.

**Wireframe:**  
https://preview.uxpin.com/da68049c0af15edd7ce4e47e4bee431fed40c30d#/pages/188730848

This update improves transparency and traceability, allowing users to track all modifications made to a report over time. By displaying all versions, users can compare historical changes and ensure accountability in compliance reporting.

**Acceptance Criteria:**  
- [ ] The Changelog displays all versions of a schedule, from the original report through all subsequent updates.
- [ ] Each version is labeled according to its type (Original, Supplemental, Reassessment, Government Adjustment).
- [ ] Users can view different versions within the Changelog view.
- [ ] The most current version is at top for easy identification.
- [ ] Ensure the UI aligns with the wireframe specifications.
- [ ] The Changelog remains persistent in all report statuses where applicable.

**Development Checklist:**  
- [ ] Update the Changelog data structure to store and retrieve all schedule versions.
- [ ] Ensure the version history is displayed in reverse chronological order, newest at top, oldest at bottom.
- [ ] Only display paging element if required (more than 10 rows).
- [ ] Validate performance to ensure the Changelog loads efficiently for large datasets.

**Notes:**  
- Ensure access control is maintained so users only see versions relevant to their role.

kevin-hashimoto:
- feat: changelog view on the summary page, closes #2216 

experimenting with new DTO schema pattern. Separates the internal schemas from the view schemas. a few reasons why this can be beneficial:
- when using the orm with eager loading (joinedload etc), we have to load the full data, this often results in a lot of data we dont need in the business logic. Validating it through the DTO enables us to only pass the data we need.
- Serializing it before using it in our business logic makes it easier to work with.
- Data from the backend that is exclusively used only in the backend does not always get passed to the front end
- using DTOs for each repo return value to keep consistency
- gives more flexibility for the view schemas (hide sensitive data, only pass what we need etc.)
- only passing in data that we need to the service functions
- DTOs can be used as a a view schema if the data does not change.
- validates and serializes data coming from the db before being used in business logic. validates data going out from the view to the frontend
- helps keep service layer independent of external API structure.
- could be a good precursor to organizing our view schemas
- debating if validation should happen in the service layer instead of the repo layer.


AlexZorkin (approved):
Looking good ty 👍

PR #2397, opened 2025-04-07
feat: Add Quarterly Fuel Supply

dhaselhan:
* Set reporting frequenecy when creating a CR based on Orgs flag
* Add new columns for each quarterly amount
* Update front end to switch between layouts
* Add logic to backend + Summary service to add them together for calculations


prv-proton (approved):
Great work on forming the baseline, looks good 👍

Issue #2308, opened 2025-03-26
LCFS - BCeID Credit calculator


StuartGALL:
**Description:**
Use the logic from the Fuel Supply Schedule to reformat a BCeID Credit Calculator based on the provided mockup. The updated calculator should incorporate the correct rules and calculations from the fuel supply logic, including the renewable fuel requirements and low carbon fuel targets. Additionally, introduce toggle functionality that dynamically adjusts logic and inputs depending on whether the user selects the “Low Carbon Fuel Requirement” or “Renewable Fuel Requirement”, or both, using radio buttons.

**Wireframe:**
Mockup of the BCeID Credit Calculator. (https://preview.uxpin.com/e0e2f15fc1b52987fd0287ad8c8103452dd16d60#/pages/201358356/simulate/no-panels)

Mockup of the Dashboard.
(https://preview.uxpin.com/e0e2f15fc1b52987fd0287ad8c8103452dd16d60#/pages/203295973?mode=dm)

Mockup of "credit calculator button" for BCeID users on the Compliance reporting page.
(https://preview.uxpin.com/e0e2f15fc1b52987fd0287ad8c8103452dd16d60#/pages/203299272?mode=dm)

Mockup of "credit calculator button" for IDIR users on the Compliance reporting page.
(https://preview.uxpin.com/e0e2f15fc1b52987fd0287ad8c8103452dd16d60#/pages/203301529?mode=dm)

**Purpose and benefit to user:**
This feature allows users to calculate their compliance obligations more accurately and conveniently by embedding the fuel supply logic into the calculator tool.

**Acceptance Criteria:**
- [ ] The calculator accurately reflects fuel volumes, renewable content, and targets based on the Fuel Supply Schedule logic.
- [ ] Radio buttons switch between “Renewable Fuel Requirement” and “Low Carbon Fuel Requirement” or both.
- [ ] Selecting the radio buttons or links dynamically updates calculation logic and field labels as needed.
- [ ] The calculator validates inputs according to the selected requirement type.
- [ ] The output clearly displays any retained credits or net credits based on user input.
- [ ] Add a credit calculator button on both the BCeID and IDIR sides.
- [ ] Add a link from the dashboard to the credit calculator as per wireframe.

**Development Checklist:**
- [ ] Integrate calculation logic from the Fuel Supply Schedule.
- [ ] Add Radio buttons for “Renewable Fuel Requirement” vs. “Low Carbon Fuel Requirement” and include one for both.
[...]

PR #2398, opened 2025-04-08
Feat: LCFS - BCeID Credit calculator - 2308

prv-proton:
Feat: #2308


AlexZorkin (changes requested):
This all looks great but I have one thought about the /public endpoint. I think we should use /calculator as the url, and just add that to the auth exception. /public is an extra layer that I don't think we need. If we ever add more public features the PublicRepository/Service would get crowded so having CalculatorRepository/Service is more defined. let me know what you think.

prv-proton:
> This all looks great but I have one thought about the /public endpoint. I think we should use /calculator as the url, and just add that to the auth exception. /public is an extra layer that I don't think we need. If we ever add more public features the PublicRepository/Service would get crowded so having CalculatorRepository/Service is more defined. let me know what you think.

Thanks for the feedback, please review the fixed code.

AlexZorkin (approved):
thank you for the updates, looks good 👍

PR #2417, opened 2025-04-08
fix: Add timezone to available balance check

dhaselhan:
* Was only checking the day, did not include transactions up until the end of March 31st in Vancouver time.


AlexZorkin (approved):
well that explains it! 👍

PR #2421, opened 2025-04-09
fix: Return full chain on update

dhaselhan:
* Change update to return the same model as create for use in the cache


AlexZorkin (approved):
Looks good ty 👍

PR #2452, opened 2025-04-10
fix: session timeout and 403 errors

### LINKED ISSUE #2325: LCFS - session timeouts
**Describe the task**
Investigate and fix the timeout and screen refresh issue in the application. The current workaround refreshes the screen every 5 minutes to prevent timeouts, but this solution is not working. A better solution is needed to ensure users are not logged out due to inactivity when they are actively using the application.

**Purpose**
To prevent user frustration and loss of work by ensuring active users do not experience session timeouts during normal use of the application.

**Acceptance Criteria**
- [ ] Identify root cause of session timeout despite active usage.
- [ ] Implement a reliable method to detect user activity (e.g. clicks, keystrokes).
- [ ] Maintain session activity without relying solely on periodic screen refreshes.
- [ ] Ensure users remain logged in while actively interacting with the application.
- [ ] Validate that the fix works across major browsers and user roles.
- [ ] Test with realistic user interaction scenarios (e.g. data entry, navigation).

**Additional context**
- The existing 5-minute screen refresh approach is insufficient.
- Consider evaluating the session management strategy (e.g., token expiration, activity polling, etc.).
- Coordinate with the security team to ensure compliance with session timeout policies.

kevin-hashimoto:
- fix: addresses 403 errors that happen when tokens refresh
- refactor: more modern approach to setting/refreshing tokens, closes #2325 

many changes are just file name changes. the real stuff is in `KeycloakProvider.jsx`


dhaselhan (approved):
LGTM 👍

PR #2453, opened 2025-04-10
feat: Disable new versions if there is already a draft

dhaselhan:
* Add is newest_flag to report chain to confirm whether newer versions exist, but dont return their data
* Use this flag to control the supplemental and assessment buttons with dialogs + text

### DIFF
--- a/backend/lcfs/web/api/compliance_report/services.py
+++ b/backend/lcfs/web/api/compliance_report/services.py
@@ -37,6 +37,7 @@ from lcfs.web.api.compliance_report.schema import (
     ComplianceReportListSchema,
     ComplianceReportStatusSchema,
     ComplianceReportViewSchema,
+    ChainedComplianceReportSchema,
 )
 from lcfs.web.api.organization_snapshot.services import OrganizationSnapshotService
 from lcfs.web.api.organizations.repo import OrganizationsRepository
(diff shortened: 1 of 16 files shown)


backend/lcfs/web/api/compliance_report/services.py
```diff
@@ -410,13 +411,45 @@ def _mask_report_status(self, reports: List, user: UserProfile) -> List:
...
+        """
+        Fetches a specific compliance report by ID.
+        """
+        report = await self.get_compliance_report_by_id(report_id, user)
+
+        compliance_report_chain = await self.repo.get_compliance_report_chain(
+            report.compliance_report_group_uuid
+        )
+
+        is_newest = len(compliance_report_chain) - 1 == report.version
```
AlexZorkin, line 429: nice one 👍

frontend/src/views/ComplianceReports/components/AssessmentCard.jsx
```diff
@@ -200,20 +195,29 @@ export const AssessmentCard = ({
...
-                          createSupplementalReport()
-                        }}
-                        startIcon={<Assignment />}
-                        sx={{ mt: 2 }}
-                        disabled={isLoading}
+                      <Tooltip
+                        title={
+                          reportData.isNewest
+                            ? ''
+                            : 'Government has a reassessment in progress.'
```
AlexZorkin, line 202: Just want to confirm that business area is ok with telling suppliers this is the case?

AlexZorkin (approved):
Looks good, just want to confirm that one point.

frontend/src/views/ComplianceReports/components/AssessmentCard.jsx
dhaselhan: yep!

PR #2456, opened 2025-04-10
feat: Move Assessment Statement into history

dhaselhan:
* Combine HistoryCard.tests into a single file
* Add new tests for assessment statement
* default expand first history item


prv-proton (approved):
Looks good 👍

PR #2465, opened 2025-04-11
Compliance Report Repo Cleanup

dhaselhan:
Split into two commits:
* Remove is_model and add hard typing to get compliance reports
* Split CR Repo (large) into two smaller reports with summary logic go into its own repo


AlexZorkin (approved):
Nice refactor thank you  👍

Issue #2139, opened 2025-03-01
LCFS - Early issuance report IDIR analyst submitted


airinggov:
**Description:**  
Create a new version of the existing compliance report specifically for IDIR to process Early Issuance reporting. Including implementing a validation rule that prevents processing of reports containing fuels that generate **negative compliance units**. Additionally, modify system behavior to exclude assessment-related language, restrict analyst adjustments/reassessments, and ensure transactions are created per quarter.

**Wireframe:**  
https://preview.uxpin.com/da68049c0af15edd7ce4e47e4bee431fed40c30d#/pages/200308747

**Purpose and Benefit to User:**  
- Provides IDIR staff the ability to process early issuance reports.

### **Acceptance Criteria:**
- [x] If an **Early Issuance Report** contains **any fuel generating negative compliance units**, it **must be returned to the supplier** and cannot be processed.
- [x] **Assessment-type language** must be removed from Early Issuance reports. 
- [x] **Analyst adjustments and reassessments are not allowed** on Early Issuance reports. If an issue is identified, the report must be **Returned to the supplier** instead.
- [ ] Ensure that **Early Issuance reports generate transactions per quarter**.

### **Development Checklist:**
- [x] Implement a validation rule preventing processing of reports with negative compliance units.
- [x] Include workflow to **Return reports to suppliers** same logic as already implemented in compliance reporting (returning report to draft status, putting schedules back in edit mode, updating/overwriting previous History info and Summary snapshots.
- [x] Modify report UI and messaging to remove **assessment-type language**.
- [x] Restrict **analyst adjustments and reassessments** for Early Issuance reports.
- [ ] Implement logic to **create transactions per quarter even if multiple quarters are issued at one time**.

Issue #2137, opened 2025-02-28
LCFS - Early issuance report BCeID draft


airinggov:
**Description:**
Create a new version of the existing compliance report specifically for Early Issuance reporting, segmented by quarter. This report will be visible only to suppliers that have the "Early Issuance Reporting" flag set to **Yes** in their Organization Profile. Early issuance reports are version increments of a compliance report. Only one version can be submitted, Q2 cannot be submitted until Q1 has been issued. Q1 and Q2 numbers can be submitted together in Q2 if they missed reporting in Q1. Each quarter will have a transaction, even if issued at the same time. If the supplier needs to make a supplemental type adjustment before the report is issued it can be returned to the supplier by the analyst.

**Wireframe:**  
https://preview.uxpin.com/da68049c0af15edd7ce4e47e4bee431fed40c30d#/pages/198364252

**Purpose and Benefit to User:**  
- Provides a structured and systematic way to report quarterly fuel supply data.  

### **Acceptance Criteria:**
- [x] Display the Early Issuance Report only if the supplier’s Organization Profile has the **Early Issuance Reporting** flag set to **Yes**.
- [x] The title should follow the format: `"YYYY Early Issuance Report - Quarter X"` (e.g., "2025 Early Issuance Report - Quarter 2").
- [x] Breadcrumb should remain `"YYYY Compliance Report"` (e.g., "2025 Compliance Report").
- [ ] The Introduction text should be updated (exact content to be determined).
- [x] The Report Activities card should display:  
  _"Report your Quarter X fuel supply as well as any supporting documents."_
- [x] Remove the links to the following schedules:
  - Notional Transfers
  - Fuels for Other Use
  - Exporting Fuel
  - Allocation agreement
  - FSE
- [x] Ensure the **Fuel supply schedule** allows reporting quantities by quarter (as implemented in Card #1118).
- [ ] Ensure the new **Summary Format** for Early Issuance reports (as implemented in Card #2358  ) (future work will integrate this into the full summary for the **Q4 year-end Compliance Report, Line 15**).
- [ ] Implement logic that prevents a new early issuance report being submitted until the previous report has been issued.

### **Development Checklist:**
- [ ] Implement report versioning logic for Early Issuance submissions.
- [ ] Modify UI components to display Early Issuance reports only for eligible suppliers.
- [ ] Update the title format dynamically based on the year and quarter.
- [ ] Modify the Introduction text field in the UI.
- [ ] Update the Report activities card content.
[...]

PR #2468, opened 2025-04-14
Feat: LCFS - Early issuance report BCeID draft & IDIR 2137  2139

prv-proton:
#2137 & #2139


backend/lcfs/db/models/compliance/ComplianceReport.py
```diff
@@ -123,6 +122,11 @@ class ComplianceReport(BaseModel, Auditable):
         default=ReportingFrequency.ANNUAL,
         comment="Reporting frequency",
     )
+    quarter = Column(
```
AlexZorkin, line 125: Each compliance report will be associated with all quarters, can you explain the use case of this quarter field?

AlexZorkin (changes requested):
The supplementals will allow editing of all quarters so just want to check the logic for quarter on the compliance report.

backend/lcfs/db/models/compliance/ComplianceReport.py
prv-proton: updated logic as per our discussion

AlexZorkin (approved):
looking good ty 👍

PR #2478, opened 2025-04-15
Fix: LCFS BCeID credit calculator 2308

prv-proton:
fix redirect issue that is still persisting for compliance-periods endpoint


kevin-hashimoto (approved):
LGTM, good catch!

PR #2485, opened 2025-04-16
fix: Load org code from correct entity + Copy FSE

dhaselhan:
* There is no org code on compliance report model, query from repo directly
* Update front end to set level of equipment as a string the same as get


backend/lcfs/web/api/compliance_report/services.py
```diff
@@ -182,19 +185,26 @@ async def create_analyst_adjustment_report(
...
+            new_report.compliance_report_id,
+            current_report.organization_id,
+        )
+
         return ComplianceReportBaseSchema.model_validate(new_report)
 
     @service_handler
     async def create_supplemental_report(
-        self, report_id: int, user: UserProfile = None, legacy_id: int = None
+        self, original_report_id: int, user: UserProfile = None, legacy_id: int = None
```
AlexZorkin, line 199: why the switch to original_report_id? this id can be any report in the chain, might add a bit of confusion

backend/lcfs/web/api/final_supply_equipment/services.py
```diff
@@ -314,3 +320,33 @@ async def get_compliance_report_by_id(self, compliance_report_id: int):
     @service_handler
     async def delete_all(self, compliance_report_id: int):
         return await self.repo.delete_all(compliance_report_id)
+
+    @service_handler
+    async def copy_to_report(
+        self, original_report_id: int, target_report_id: int, organization_id: int
```
AlexZorkin, line 326: we want to be copying the previous reports FSE not the original report FSE records. If they update FSE in a supplemental those changes should be carried forward.

AlexZorkin (changes requested):
Just want to confirm the copy logic around which report FSE records we're copying to the next report in the chain.

backend/lcfs/web/api/compliance_report/services.py
dhaselhan: I wanted a way to differentiate between the two reports, and report_id does not do that.

backend/lcfs/web/api/final_supply_equipment/services.py
dhaselhan: Correct that is how it works. Is this a question or a suggestion?

dhaselhan: Should I name this copy_from_id?

AlexZorkin: all good. Just wanted to clarify.

AlexZorkin (approved):
Looks good ty 👍

PR #2486, opened 2025-04-16
fix: Add join for provision of the act

dhaselhan:
* Was not being loaded resulting in a null reference
* Rename objects for better comprehension of new vs old


AlexZorkin (approved):
Looks good 👍

PR #2487, opened 2025-04-16
fix: Use Get All Allocation Agreements

dhaselhan:
This will return all records including changelog ones, get allocation agreements does not.


prv-proton (approved):
Looks good 👍

PR #2491, opened 2025-04-17
fix: Return same response schema all the time

dhaselhan:
* Union types cause issues with Schema validators, it will run the wrong validators
* Update update to always return the standard response
* Add SelectRenderers where missing
* Add onSettled back to GetFuelSuppliesList


AlexZorkin (approved):
looks good 👍

dhaselhan:
Closing as someone made another fix for this issue.

Issue #2477, opened 2025-04-15
LCFS - Pagination issue on "Compliance report" view table


Grulin:
**Describe the Bug**
As an IDIR user, when accessing the Compliance Report table—either via the dashboard shortcut or the "Compliance Report" navigation tab—the pagination control does not retain the selected page size. For example, if the user sets the page size to 100 and then navigates to the next page using the right arrow icon, the page size resets to the default value of 10.

**Expected Behaviour**
Once a user selects a specific page size (e.g., 100 reports per page), that preference should persist across pagination navigation until the user manually changes it.

**Actual Behaviour**
The page size resets to 10 when navigating to the next page, despite the user previously selecting a different page size.

**Implications**
This behaviour is inconsistent and disrupts data scanning workflows, especially for bulk reviews of compliance report submissions.

**Steps To Reproduce**
Steps to reproduce the behaviour:
User/Role: IDIR
1. Go to Navigate to the Compliance Report table via the dashboard or the navigation tab.
2. Change the page size to 100.
3. Click the right arrow icon to navigate to the next page.
4. Observe that the page size resets to 10.

Issue #2476, opened 2025-04-15
LCFS - Home dashboard compliance report counter is incorrect


Grulin:
**Describe the Bug**
The dashboard counter that displays the number of submitted compliance reports is not accurately reflecting the true number of reports submitted by obligated parties. Discrepancies have been noted between the actual submissions (as viewed in the backend/reporting tables) and the counter shown on the dashboard.

**Expected Behaviour**
The dashboard should display an accurate, real-time count of submitted compliance reports based on the finalized reports for the active compliance period. This should include the following statuses: submitted, recommended by analyst and recommended by manager. 

**Actual Behaviour**
The counter is either underreporting or overreporting the number of submitted compliance reports. 

**Implications**
Incorrect summary of compliance reports in progress

**Steps To Reproduce**
Steps to reproduce the behaviour:
User/Role: IDIR
1. Go to home dashboard
2. See Compliance reports widget showing 315 reports in progress

PR #2495, opened 2025-04-22
Fix: LCFS -compliance report counter and pagination issues 2476 - 2477

prv-proton:
#2476
#2477
- remove caching for useCurrentUser hook


backend/lcfs/web/api/compliance_report/services.py
```diff
@@ -325,7 +323,7 @@ async def get_compliance_reports_paginated(
         user: UserProfile,
     ):
         """Fetches all compliance reports"""
-        is_bceid_user = user_has_roles(user, [RoleEnum.GOVERNMENT])
+        is_bceid_user = user_has_roles(user, [RoleEnum.SUPPLIER])
```
dhaselhan, line 326: This looks a little odd, do we need to rename the variable?

dhaselhan (changes requested):
Just want to confirm the role check change

backend/lcfs/web/api/compliance_report/services.py
prv-proton: I believe this is fine, Since it's for BCeID users that we need to pull all report ID's that are in either of "Recommended" or "Submitted" status when the supplier user selects status filter of type "Submitted". We've a different logic to mask the statuses to "Submitted", if they're in recommendation.

dhaselhan: You are totally right!

PR #2496, opened 2025-04-22
fix: Re-load Data and Add Missing Select Renderer

dhaselhan:
* Fixes EER rounding by re-fetching
* Fixes validation error by clearing errors after deletes
* Fixes missing arrows by adding select rendereers


AlexZorkin (approved):
Looks good 👍

PR #2497, opened 2025-04-22
feat: Expand Locust Tests to include Create

dhaselhan:
* Remove compliance_period from FuelExports, it was the only schedule sending it from the FE
* Delete Export Validation as it was an empty unused class


AlexZorkin (approved):
Nice cleanup ty 👍

PR #2503, opened 2025-04-23
Feat: Other Use Form Text - 2492

AlexZorkin:
- updated other uses text on form
- removed slow irrelevant test


dhaselhan (approved):
Love it 👍

PR #2508, opened 2025-04-24
fix: Fix FSE not saving correctly

dhaselhan:
* Return flattened fields and add standardized mapping function


AlexZorkin (approved):
great ty 👍

PR #2512, opened 2025-04-25
Fix: Summary Penalty Decimals - 2505

AlexZorkin:
-  more robust decimal support added to summary calculations


dhaselhan (approved):
Nice work! Good call on using Decimal!

PR #2522, opened 2025-04-25
feat: Change query analyzer to not use side effects

dhaselhan:
* Query analyzer used to require importing and ran as a side effect
* Convert to a function call to register the hooks to make it explicit


AlexZorkin (approved):
👍

PR #2546, opened 2025-05-02
fix: compliance report list status label

### LINKED ISSUE #2533: LCFS - Recommended by analyst display
**Describe the Bug**  
The label "Recommended by analyst" is incorrectly appearing on the BCeID side of the Compliance Report Index. This label is intended only for internal use (IDIR users) and should not be visible to BCeID users it should display as "Submitted".

Image

kevin-hashimoto:
- fix: compliance report list status label, closes #2533

feel free to merge as I will be away until May 20th


prv-proton (approved):
Looks good 👍

PR #2562, opened 2025-05-06
Feat: IDIR Supplemental Reports

AlexZorkin:
- Added new endpoint to handle idir side creating supplemental reports
- button logic on when actions are visible when supplemental draft exists
- tests updated and added


prv-proton (approved):
Looks good 👍 
Just few suggestions

- it would be nice to redirect the page to index page once the analyst creates the supplemental report.
- There is no way to view in the UI, that the supplemental report was requested from Analyst, would be nice to have one.

hamed-valiollahi (approved):
Looks great! Tested locally and everything works as expected. Thanks!

PR #2572, opened 2025-05-07
Fix: Keep submitted report visible to BCeID users - 2537

### LINKED ISSUE #2537: LCFS - Government Adjustment visibility
**Describe the Bug**  
When an analyst creates a Government Adjustment on a submitted compliance report, the original submitted report disappears from view for BCeID users. This breaks expected behavior where BCeID users should continue to see the submitted version of the report, even after an adjustment is initiated.

**Expected Behaviour**  
Once a report is submitted, it should remain visible in its submitted state to BCeID users, regardless of whether a Government Adjustment is created by an analyst.

**Actual Behaviour**  
Creating a Government Adjustment causes the submitted report to become hidden from BCeID users, making it seem as though the report no longer exists in their view.

**Implications**  
- BCeID users lose visibility of their submitted compliance reports, leading to confusion and support inquiries.  
- Lack of access to the submitted report hinders reference, accountability, and communication around the Government Adjustment.  
- This may result in unnecessary resubmissions or misaligned expectations.

**Steps To Reproduce**  
User/Role:  
1. BCeID user submits a compliance report.  
2. Analyst logs in and creates a Government Adjustment on the submitted report.  
3. BCeID user logs back in and navigates to the report list.  
4. Observe that the submitted report is no longer visible in their list.

**Notes:**  
- Review report visibility logic tied to adjustment creation.  
- Ensure the report’s original version remains accessible in the BCeID user interface after adjustments begin.  
- Consider a new status for legacy submitted reports. If the analyst deletes the Government adjust the legacy report would need to be returned to Submitted status. (related to card #2516 )

hamed-valiollahi:

- Ensures BCeID users can still view the original submitted report after an analyst creates a Government Adjustment.
- Fixes the issue where BCeID users were seeing reports in "Recommended" status by Analyst or Manager instead of "Submitted".

Closes #2537


AlexZorkin (approved):
👍

hamed-valiollahi:
Thanks for the review!

PR #2575, opened 2025-05-08
Fix: line 17 pulled from available balance - 2489

AlexZorkin:
- line 17 now checks available balance on supplemental create


prv-proton (approved):
Looks good 👍

hamed-valiollahi (approved):
Looks great, clean and well tested. Thanks for the fix!

PR #2608, opened 2025-05-14
Fix Compliance report 400 error on dev

prv-proton:


hamed-valiollahi (approved):
Looks good to me! Thanks for the fix! 👍

PR #2610, opened 2025-05-14
Fix: Restore default selection and clean up Compare Reports UI - 2534

### LINKED ISSUE #2534: LCFS - table header
**Describe the Bug**  
The "Non-compliance penalty payable" summary table in compare mode displays incorrect column headings. Specifically, two column headers do not match the expected headings from the equivalent summary tables above, leading to inconsistency in the report layout and potential confusion for users comparing data across sections.

Image

hamed-valiollahi:
- Restores default selection between the two most recent reports
- Removes extra header text

Closes #2534


frontend/src/views/CompareReports/CompareReports.jsx
```diff
@@ -45,6 +45,17 @@ export const CompareReports = () => {
     if (complianceReport) {
       const { chain } = complianceReport
       setReportChain(chain)
+
+      // Compare the latest report with the oldest one
+      if (!report1ID && !report2ID && chain?.length) {
+        const lastIndex = chain.length - 1
+        const newest = chain[lastIndex]?.complianceReportId ?? null
+        const oldest = chain[0]?.complianceReportId ?? newest
```
prv-proton, line 53: I think as per the comment, we should be comparing between the most recent ones.
https://github.com/bcgov/lcfs/issues/2534#issuecomment-2859028755

hamed-valiollahi:
> Need to compare between the latest two version of the reports by default.

You’re right! Not sure how I missed that part! thanks so much for pointing it out! 👍

prv-proton (approved):
Looks good 👍

PR #2614, opened 2025-05-14
Fix: LCFS - BUG - Paging controls on schedule #2557

### LINKED ISSUE #2557: LCFS - BUG - Paging controls on schedule
**Describe the Bug**  
Paging controls are appearing on certain schedules even when the number of entries does not require pagination. This behavior is inconsistent with the intended design, which only shows paging controls when necessary (i.e., when there are more rows than can be displayed in the current view).

Image

**Expected Behaviour**  
Paging controls should only be displayed when the number of entries exceeds the visible limit for a schedule. If all entries can be displayed without scrolling or overflow, the paging should be hidden.

**Actual Behaviour**  
Paging controls are being shown unnecessarily on a few schedules, despite the limited number of entries that do not require pagination.

**Implications**  
- Creates visual clutter and confusion for users.
- Suggests that more data exists than actually does.
- Breaks consistency with established UI behavior across schedules.

**Steps To Reproduce**  
User/Role: Any  
1. Navigate to a schedule that currently has only a few entries.  
2. Observe that paging controls are visible even though all rows are already displayed.  
3. Confirm that this behavior deviates from other schedules with similar entry counts.

**Notes:**  
- Refer to the attached screen capture for an example of the issue.
- Review UI logic responsible for rendering pagination and ensure it respects row thresholds before activating paging controls.

prv-proton:
Fix #2557


AlexZorkin (approved):
👍

PR #2626, opened 2025-05-15
Fix: LCFS - BUG - Uploading documents supplemental reporting #2612

### LINKED ISSUE #2612: LCFS - BUG - Uploading documents supplemental reporting
**Describe the Bug**  
Uploading documents during the supplemental reporting process fails with a 400 error. Users are unable to successfully attach required files, which is preventing completion of the report submission workflow.

**Expected Behaviour**  
Users should be able to upload supporting documents in supplemental reports without encountering errors. Files should be processed and attached to the report successfully.

**Actual Behaviour**  
When attempting to upload a document, the system returns a 400 error and the upload fails. This occurs consistently, regardless of file type or user role.

**Implications**  
- Users are blocked from submitting complete supplemental reports.
- The issue can delay reporting deadlines and regulatory compliance.
- May lead to user frustration and increase support burden.

**Steps To Reproduce**  
User/Role: BCeID / Compliance Reporter  
1. Log in to the LCFS Portal.  
2. Navigate to a supplemental report.  
3. Attempt to upload a document (e.g., PDF, DOCX).  
4. Observe that the upload fails and a 400 error is returned.

**Notes**  
- Check request payload format and size constraints.
- Verify backend validation rules and response structure for document uploads.
- Test across environments to determine if the issue is environment-specific.

prv-proton:
Fix #2612

### DIFF
--- a/backend/lcfs/services/s3/client.py
+++ b/backend/lcfs/services/s3/client.py
@@ -165,7 +165,7 @@ class DocumentService:
         if (
             RoleEnum.SUPPLIER in user.role_names
             and compliance_report.current_status.status
-            != ComplianceReportStatusEnum.Draft.value
+            != ComplianceReportStatusEnum.Draft
         ):
             raise HTTPException(
                 status_code=400,


hamed-valiollahi (approved):
Looks good to me! 👍

Issue #2622, opened 2025-05-15
LCFS - BUG - FSE Edited flag


airinggov:
**Describe the Bug**  
The FSE accordion is incorrectly displaying an 'Edited' flag when a supplemental report is created, even though no edits have been made. Additionally, FSE does not have changelog functionality enabled and should not display change-tracking flags at all.

**Expected Behaviour**  
The FSE section should not show an 'Edited' flag or any such changelog tracking flags.

**Actual Behaviour**  
- An 'Edited' flag is automatically applied to FSE data when a supplemental report is initiated.
- This occurs despite no user-initiated edits.
- The flag appears even though FSE is not subject to changelog tracking.

**Implications**  
- Misleads users into thinking changes have been made to FSE data.

**Steps To Reproduce**  
User/Role: BCeID
1. Create a supplemental report based on an existing report with FSE data.  
2. Navigate to the FSE section.  
3. Observe that the 'Edited' flag is present despite no modifications.  
4. Confirm that FSE changelog is not enabled.

**Notes**  
- Investigate FSE component's logic to ensure changelog flags are not displayed.

airinggov:
@prv-proton I'm still seeing the Edited flag on test. It should never show up under any circumstance at this point as there is no changlelog set up for FSE.

![Image](https://github.com/user-attachments/assets/cc937c75-5849-416b-8169-f0a827c7161b)

Issue #2639, opened 2025-05-15
LCFS - BUG - Notional transfers table rendering


airinggov:
**Describe the Bug**  
When toggling from the "Change log" view back to the current view in the Notional Transfers section, the table remains in a distorted or incorrect format. This layout issue persists until the user re-selects the table, at which point it returns to the expected format.

**Expected Behaviour**  
Switching from the "Change log" back to the current view should automatically restore the standard table layout without requiring any additional user interaction.

**Actual Behaviour**  
- Table remains incorrectly formatted after returning from the Change log view.  (Extra long gray bar)
- The layout is only corrected when the table is manually reselected.
![Image](https://github.com/user-attachments/assets/f1b19986-fd5c-4f43-a496-0d0f13c88392)

**Implications**  
- Confusing and unprofessional visual experience.  
- Potential user frustration due to unclear or broken interface behavior.  
- May lead to misinterpretation of data during review.

**Steps To Reproduce**  
User/Role: Any  
1. Navigate to the Notional Transfers section of a report.  
2. Select the “Change log” view.  
3. Switch back to the current view.  
4. Observe the table remains in an incorrect format until reselected.

**Notes**  
- Review rendering and refresh logic when toggling between Change log and current views.  
- Ensure consistent formatting behavior across all report sections with similar toggle functionality.

Issue #2631, opened 2025-05-15
LCFS - BUG - Schedule added in changelog


airinggov:
**Describe the Bug**  
When adding a new schedule and data to a supplemental report, the changelog view does not correctly flag the newly added row as "Added" in green. This results in inaccurate or incomplete visual feedback in the changelog, reducing its effectiveness for tracking modifications.

**Expected Behaviour**  
Any newly added rows to a supplemental report—whether as part of a new schedule or within an existing one—should be clearly identified and flagged as "Added" in the changelog view.
![Image](https://github.com/user-attachments/assets/92d01840-21c6-464d-9142-3b20eb5df367)

**Actual Behaviour**  
The changelog fails to display added rows as "Added" when new schedule data is added to a supplemental report. This includes visual indicators and change annotations that are otherwise present for modifications.
![Image](https://github.com/user-attachments/assets/aadd66b6-73ea-465e-9154-7dac86cc4a0b)

**Implications**  
- Reduces the utility of the changelog for validation and review processes.  
- May lead to missed updates or misunderstandings about changes made in a supplemental report.  

**Steps To Reproduce**  
User/Role: IDIR or BCeID   
1. Open or create a supplemental compliance report.  
2. Add a new schedule and populate it with new data.  
3. Save the data and view the changelog.  
4. Observe that the new row is not marked or highlighted as "Added."

**Notes**  
- Review changelog logic for detecting and flagging additions to new schedules within supplemental reports et al.  
- Ensure consistent behavior with how changes and additions are flagged across different report versions and types.  
- Refer to the attached screen capture for visual reference of the issue.

PR #2650, opened 2025-05-16
Fix: Compliance report schedules bug fixes

prv-proton:
Bug fixes: #2622 , #2631 , #2639

### DIFF
--- a/frontend/src/views/FuelExports/FuelExportChangelog.jsx
+++ b/frontend/src/views/FuelExports/FuelExportChangelog.jsx
@@ -66,8 +66,7 @@ export const FuelExportChangelog = () => {
   return (
     <Box>
       {changelogData?.map((item, i) => {
-        const isCurrentOrOriginalVersion =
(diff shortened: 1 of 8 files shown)


frontend/src/views/FuelExports/FuelExportChangelog.jsx
```diff
@@ -66,6 +66,8 @@ export const FuelExportChangelog = () => {
   return (
     <Box>
       {changelogData?.map((item, i) => {
+        const isCurrentOrOriginalVersion =
+          i === 0 || item.nickname.includes('Original')
```
AlexZorkin, line 70: I am a little hesitant to use nickname as an identifying variable. Can we not just use version = 0 instead? that would be more explicit.

AlexZorkin (changes requested):
Looks good just one concern.

prv-proton:
> Looks good just one concern.

Thanks for the review, now it has been addressed.

AlexZorkin (approved):
great 👍

PR #2651, opened 2025-05-16
Fix: analyst adjustment transaction creation - 2630

AlexZorkin:
- analyst adjustment didn't have a transaction already existing like submitted status report, so we had to check and create a new one once assessed.


prv-proton (approved):
Looks good 👍

PR #2652, opened 2025-05-16
Fix: Supplemental Line 17 - 2632

AlexZorkin:
- line 17 on summary supplemental creation now calculates available balance


prv-proton (approved):
Looks good 👍

PR #2658, opened 2025-05-20
Fix: LCFS - BUG - FSE form inputs #2625

### LINKED ISSUE #2625: LCFS - BUG - FSE form inputs
**Describe the Bug**  
Two issues have been identified in the Final Supply Equipment (FSE) input section when working within a supplemental report:

1. The 'Intended use' and 'Intended users' dropdown selections are not being saved after input. They appear after being saved but not prior, they also can't be edited to add additional selections.
2. The Address field's lookup functionality only works when entering 1–3 digits; entering 4 or more digits causes the address suggestions to stop appearing.

**Expected Behaviour**  
1. Selected values in the 'Intended use' and 'Intended users' fields should be saved and persist when the form is saved or revisited.
2. The Address field should provide lookup suggestions consistently, regardless of the number of digits entered.

**Actual Behaviour**  
1. Inputs in the 'Intended use' and 'Intended users' fields appear selectable but are not retained after saving or navigating away.
2. The address auto-suggestion functionality fails when four or more digits are entered in the Address field, making it difficult or impossible to complete the field using lookup.

**Implications**  
- Users may unknowingly lose input data, resulting in incomplete or invalid reports.
- The broken address lookup disrupts workflow and can lead to user frustration or incorrect address entries.
- Impacts accuracy and completeness of FSE reporting, especially in supplemental reports.

**Steps To Reproduce**  
User/Role: BCeID   
1. Open or create a supplemental report with FSE data.  
2. Attempt to select and save values for 'Intended use' and 'Intended users.'  
3. Save and return to verify that inputs were not retained.  
4. In the Address field, enter 1–3 digits and observe the lookup working.  
5. Enter 4 or more digits and observe that no suggestions are returned.

**Notes**  
- Review save logic for multi-select fields to ensure state persistence.  
- Validate the address lookup service and input parsing for digit thresholds.

prv-proton:
Fix #2625

### DIFF
--- a/frontend/src/views/FinalSupplyEquipments/AddEditFinalSupplyEquipments.jsx
+++ b/frontend/src/views/FinalSupplyEquipments/AddEditFinalSupplyEquipments.jsx
@@ -9,7 +9,9 @@ import { defaultColDef, finalSupplyEquipmentColDefs } from './_schema'
 import {
   useFinalSupplyEquipmentOptions,
   useGetFinalSupplyEquipments,
-  useSaveFinalSupplyEquipment
+  useSaveFinalSupplyEquipment,
+  useImportFinalSupplyEquipment,
+  useGetFinalSupplyEquipmentImportJobStatus
 } from '@/hooks/useFinalSupplyEquipment'
 import { useCurrentUser } from '@/hooks/useCurrentUser'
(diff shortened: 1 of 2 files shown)


AlexZorkin (approved):
👍

PR #2660, opened 2025-05-20
fix: updated return and supplemental button logic

AlexZorkin:
- fixed button logic to show return to supplier for submitted supplemental reports


hamed-valiollahi (approved):
Looks good to me! Thanks 👍

PR #2663, opened 2025-05-21
fix: remove empty accordions

### LINKED ISSUE #2640: LCFS - BUG - Remove empty accordions
**Describe the Bug**  
In supplemental reports (including reassessments and government adjustments), accordions for empty sections currently display with an "Empty" tag and remain visible even when they contain no information. This creates unnecessary visual clutter and detracts from the usability of the interface.

**Expected Behaviour**  
- Empty accordions should be hidden entirely in supplemental reports unless they contain content.
- The only exception is the "Supporting Documents" accordion, which should remain visible to IDIR Analysts and display the pencil icon for uploads.

**Actual Behaviour**  
- All accordions are displayed regardless of content, each tagged as "Empty" if no data is present.
- This includes unnecessary sections that have no relevance to the current report version.
Image

**Implications**  
- Clutters the UI with irrelevant information.  
- Increases cognitive load for users trying to identify relevant content.  
- May create confusion around which sections require attention or are intentionally blank.

**Steps To Reproduce**  
User/Role: IDIR Analyst or BCeID user  
1. Open a supplemental report (reassessment or government adjustment) with no data in certain sections.  
2. Observe that all section accordions are displayed, even those with no content, and are marked as "Empty."  
3. Note that only the "Supporting Documents" section requires visibility in this state.

**Notes**  
- Update rendering logic to conditionally hide empty accordions in supplemental reports.  
- Ensure the "Supporting Documents" section is always displayed for IDIR users with the pencil icon for file upload.  
- Verify that changes do not impact original compliance reports or break conditional UI rules for non-analyst roles.

kevin-hashimoto:
fix: hide empty accordions on reports, closes #2640

### DIFF
--- a/frontend/src/views/ComplianceReports/components/ReportDetails.jsx
+++ b/frontend/src/views/ComplianceReports/components/ReportDetails.jsx
@@ -386,7 +386,9 @@ const ReportDetails = ({ canEdit, currentStatus = 'Draft', userRoles }) => {
           scheduleData.every((item) => item.actionType === 'DELETE')
 
         return (
-          ((scheduleData && !isArrayEmpty(scheduleData)) || hasVersions) && (
+          ((((scheduleData && !isArrayEmpty(scheduleData)) || hasVersions) &&
+            !isDisabled) ||
+            index === 0) && (
             <Accordion
               sx={{
                 '& .Mui-disabled': {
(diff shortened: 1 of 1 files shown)


hamed-valiollahi:
Since my solution is a bit different from yours, just wanted to let you know that here's my commit for your review.
ed88762fb5f1f3afc16fe6ad350649a41f70318a

hamed-valiollahi (approved):
Sounds good! Thanks, Kevin!

PR #2673, opened 2025-05-22
fix: fuel export energy density

### LINKED ISSUE #2661: LCFS - BUG in Export schedule column "Energy Density" for "Other" fuel type
**Describe the Bug**
When the "Other" fuel type is selected in the export schedule, the "Energy density" column incorrectly displays "null MJ/". This column should allow users to input an energy density figure specific to the unrecognized fuel they are reporting.

**Expected Behaviour**
The "Energy density" column should be editable, allowing users to input a specific energy density figure for the "Other" fuel type.

**Actual Behaviour**
After entry, the "Energy density" column displays "null MJ/" and does not save any input.

Image

**Implications**
Does not allow a user to input a fuel type "Other" correctly. 

**Steps To Reproduce**
User/Role: BCeID
1. Navigate to the Export Schedule.
2. Select "Other" as the fuel type.
3. Observe the Energy Density column displaying "null MJ/".

kevin-hashimoto:
- fix: fuel export other option energy density null and not saving, closes #2661

### DIFF
--- a/backend/lcfs/web/api/fuel_export/actions_service.py
+++ b/backend/lcfs/web/api/fuel_export/actions_service.py
@@ -74,7 +74,9 @@ class FuelExportActionService:
         fuel_export.ci_of_fuel = fuel_data.effective_carbon_intensity
         fuel_export.target_ci = fuel_data.target_ci
         fuel_export.eer = fuel_data.eer
-        fuel_export.energy_density = fuel_data.energy_density
+        fuel_export.energy_density = (
+            fuel_export.energy_density or fuel_data.energy_density
+        )
         fuel_export.uci = fuel_data.uci
 
         # Calculate total energy if energy density is available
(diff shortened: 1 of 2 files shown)


backend/lcfs/web/api/fuel_export/actions_service.py
```diff
@@ -74,7 +74,9 @@ async def _populate_fuel_export_fields(
         fuel_export.ci_of_fuel = fuel_data.effective_carbon_intensity
         fuel_export.target_ci = fuel_data.target_ci
         fuel_export.eer = fuel_data.eer
-        fuel_export.energy_density = fuel_data.energy_density
+        fuel_export.energy_density = (
+            fuel_export.energy_density or fuel_data.energy_density
```
prv-proton, line 78: using or can have it's implications. If fuel_export.energy_density = 0 and fuel_data.energy_density has some truthy value, it will choose fuel_data.energy_density, so is it ok to not preserve 0 as a valid value?

kevin-hashimoto: my thinking is that because fuel export 'other' does not have a set energy density (`fuel_export.energy_density` - which can be provided by the user), if the provided value is 0, `fuel_data.energy_density` will always be 0/false. when types other than 'other' are selected, energy density is automatically set from the default value.

prv-proton (approved):
Looks good 👍

PR #2681, opened 2025-05-22
Fix: Set Session Txn Order

AlexZorkin:
- cleanup of our session dependencies to avoid competing transactions


prv-proton (approved):
Nice fix, looks good 👍

PR #2683, opened 2025-05-22
Feat: Idir Uploads - 2543

AlexZorkin:
allow analysts to upload documents in more compliance report states
return uploader and timestamp information from document API
show username and PST timestamp for uploaded documents
restrict delete button to the document uploader only
display supporting-documents accordion even when no documents exist


prv-proton (approved):
Looks good 👍

PR #2694, opened 2025-05-26
Fix: Penalties in Compare Mode - 2629

### LINKED ISSUE #2629: LCFS - BUG - Penalties in compare mode
**Describe the Bug**  
In Compare mode, penalty amounts from the bottom summary table are not being displayed. This affects users trying to assess differences between versions of compliance reports, as critical penalty-related data is missing in compare mode.

**Expected Behaviour**  
Penalty amounts from the bottom summary table should be displayed in Compare mode. Additionally, the default behavior for Compare mode should automatically select the latest two versions of the report, with the older version displayed in the left column and the latest version in the center column.

**Actual Behaviour**  
- Penalty amounts are not visible in Compare mode.
- The default selections for Compare mode do not automatically set them in the expected order.
Image
Image

**Implications**  
- Users may overlook or misinterpret critical penalty data when comparing reports.
- Increased risk of non-compliance due to undetected penalty changes.
- Confusion or inefficiencies due to unexpected report version selections in the comparison interface.

**Steps To Reproduce**  
User/Role: Any  
1. Open Compare mode for a compliance report.  
2. Observe that the penalty amounts from the summary section are not visible.  
3. Note the default version selections and their positioning (not ordered as expected).

**Notes**  
- Update the Compare mode logic to include the display of penalty amounts from the summary table.  
- Adjust default selection logic to automatically pick the latest two versions of a report, placing the older on the left and the newer in the center.

hamed-valiollahi:
This PR includes the following:
- Fixed missing penalty amounts in Compare mode
- Improved default version selection (older left, newer right)

Closes #2629

### DIFF
--- a/frontend/src/views/CompareReports/CompareReports.jsx
+++ b/frontend/src/views/CompareReports/CompareReports.jsx
@@ -51,8 +51,8 @@ export const CompareReports = () => {
         const sortedChain = [...chain].sort(
           (a, b) => new Date(b.timestamp) - new Date(a.timestamp)
         )
-        setReport1ID(sortedChain[0].complianceReportId)
-        setReport2ID(sortedChain[1].complianceReportId)
+        setReport1ID(sortedChain[1].complianceReportId)
+        setReport2ID(sortedChain[0].complianceReportId)
       }
 
       setIsLoading(false)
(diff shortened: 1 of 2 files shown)


AlexZorkin (approved):
👍

Issue #1789, opened 2025-01-24
LCFS - BUG - Dashboard Summary Widget: Missing Data for Unregistered Organizations


prv-proton:
**Describe the Bug**
The summary widget on the dashboard only pulls details for registered organizations. Unregistered organizations are excluded, which results in incomplete data being displayed in the widget.

**Expected Behaviour**
The widget should display compliance units for both registered and unregistered organizations to provide a complete summary of all organizations.

**Actual Behaviour**
The widget currently only shows compliance units for registered organizations, omitting any data related to unregistered organizations.

**Implications**
This could lead to confusion or misrepresentation of the total compliance units for "All organizations" and might cause incorrect decision-making based on incomplete information.

**Steps To Reproduce**
Steps to reproduce the behaviour:
User/Role: IDIR role
1. Go to the dashboard.
2. Locate the "Summary" widget.
3. Observe the data displayed under "All organizations."
4. Confirm that only registered organizations' compliance units are shown, excluding unregistered ones.

![Image](https://github.com/user-attachments/assets/cfdae3d8-6c71-4e39-95c9-147c38519194)

airinggov:
I changed an organization to unregistered and the compliance balance still changed as a result, this shouldn't happen

PR #2698, opened 2025-05-27
Fix: Dashboard Summary Widget missing unregistered organizations - 1789

hamed-valiollahi:
Fixes bug #1789 where unregistered organizations were missing from the Dashboard Summary Widget. Now shows both Registered and Unregistered orgs, excluding Suspended and Canceled.


AlexZorkin (approved):
Passing the statuses through is much cleaner nice 👍

hamed-valiollahi:
> Passing the statuses through is much cleaner nice 👍

Glad you think so! Thanks so much for the code review.

PR #2706, opened 2025-05-27
Fix: Line 17 Update, Document Link - 2702 2700

AlexZorkin:
- updated line 17 calc to take into account locked status
- document link text formatting updated


prv-proton (approved):
Looks good 👍

PR #2716, opened 2025-05-27
Fix: Nested Transactions Removed

AlexZorkin:
- Removed all nested transactions to fix cpu errors


hamed-valiollahi (approved):
Everything looks great! Thanks 👍

PR #2731, opened 2025-05-28
Fix: LCFS - BUG - Change log #2699

### LINKED ISSUE #2699: LCFS - BUG - Change log
**Describe the Bug**  
The change log comparison view between "Edit new" and "Edit old" does not currently include changes to the 'Fuel type' and 'Fuel category' fields. These fields are not being highlighted or tracked in the change log when modifications occur.

**Expected Behaviour**  
- The change log should include both 'Fuel type' and 'Fuel category' when displaying differences between old and new entries.
- If either field has changed, the previous value should be shown with a yellow background and strikethrough text.
- The updated value should be shown normally, clearly indicating what was changed.
- All fields in the change log should follow this consistent logic for visual highlighting and formatting.

**Actual Behaviour**  
- Changes to 'Fuel type' and 'Fuel category' are not displayed in the comparison.
- Users may miss important modifications, which reduces the utility of the change tracking feature.

Image

**Implications**  
- Incomplete change tracking undermines transparency and accountability in data reporting.
- Users cannot easily identify key updates, increasing the risk of oversight or error.

**Steps To Reproduce**  
User/Role: Any user reviewing changes in the compliance report  
1. Edit a previously saved schedule entry and modify either the 'Fuel type' or 'Fuel category'.
2. Open the change log comparison view.
3. Observe that the changes to these fields are not shown or highlighted.

**Notes:**  
- Ensure the visual indicators (yellow background and strikethrough) are implemented uniformly across all tracked fields.
- Validate changes are captured correctly in both the frontend display and underlying data model.

prv-proton:
Fix #2699
- change log wasn't showing the correct differences being highlighted, as the property names had mismatch.

### DIFF
--- a/frontend/src/views/FuelExports/_schema.jsx
+++ b/frontend/src/views/FuelExports/_schema.jsx
@@ -740,18 +740,18 @@ export const changelogCommonColDefs = (highlight = true) => [
   {
     headerName: i18n.t('fuelExport:fuelExportColLabels.fuelTypeId'),
     field: 'fuelType.fuelType',
-    cellStyle: (params) => highlight && changelogCellStyle(params, 'fuelTypeId')
+    cellStyle: (params) => highlight && changelogCellStyle(params, 'fuelType')
   },
   {
     headerName: i18n.t('fuelExport:fuelExportColLabels.fuelCategory'),
     field: 'fuelCategory.category',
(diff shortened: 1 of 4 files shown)


AlexZorkin (approved):
Good fix ty 👍

PR #2738, opened 2025-05-29
fix: signing declaration showing for idir

### LINKED ISSUE #2719: LCFS - BUG - Signing Authority Declaration
**Describe the Bug**  
When an IDIR analyst returns a supplemental compliance report to a supplier, the Signing Authority Declaration section becomes visible to the analyst, which is not intended behavior. This declaration is meant only for BCeID users during submission.

**Expected Behaviour**  
The Signing Authority Declaration should only be visible to BCeID users who are submitting a compliance report. IDIR analysts should never see this section, regardless of the report status or actions taken.

**Actual Behaviour**  
After returning a supplemental report to the supplier, the IDIR analyst sees the Signing Authority Declaration on the screen.

Image

**Implications**  
- This can create confusion for IDIR users who are not required to and should not interact with signing declarations.
- It may lead to erroneous assumptions about IDIR responsibilities or system behavior.
- It represents a role-based access issue that could have broader implications if similar logic errors exist elsewhere.

**Steps To Reproduce**  
User/Role: IDIR (Analyst)  
1. Open a supplemental compliance report.  
2. Return the report to the supplier.  
3. Observe that the Signing Authority Declaration section appears, despite being irrelevant to IDIR users.

**Notes:**  
- Verify role-based UI rendering rules to ensure the Signing Authority Declaration only displays for eligible BCeID users.  
- Check if similar sections are mistakenly shown to IDIR users in other report workflows.

kevin-hashimoto:
- fix: signing declaration showing for idir users after returning to supplier. closes #2719


AlexZorkin (approved):
👍

PR #2742, opened 2025-05-29
Fix: Supplemental Assessment Visibility - 2703

AlexZorkin:
- updated assessment visibility logic for draft supplemental scenarios.


hamed-valiollahi (approved):
Looks good to me! 👍

PR #2765, opened 2025-06-03
Fix: environment default patch

AlexZorkin:
- added patch for environment variable that was not getting set properly in the backend


prv-proton (approved):
Looks good 👍

PR #2785, opened 2025-06-06
Fix: Analyst Adjustment Logic - 2784 2781

AlexZorkin:
- updated return logic on analyst adjustments
- fixed delete button visibility logic


prv-proton (approved):
Looks good 👍

PR #2788, opened 2025-06-06
Feat: Editing Changelog Styling - 2782

AlexZorkin:
- styling updated for edit row to only highlight first cell
- tests updated


prv-proton (changes requested):
Not  able to update the existing rows in a government adjustment report. All the original rows have green background and disabled from editing.

prv-proton (changes requested):
Everything looks good, just for Delete, it doesn't highlight the entire row instead just the action cell is highlighted.

Issue #2733, opened 2025-05-28
LCFS - Optimize Compliance Report Data Loading - Reduce Database Connection Hammering


AlexZorkin:
**Describe the task**
When opening a compliance report, the application currently makes 14+ separate API requests to load all the required data (fuel supplies, notional transfers, allocation agreements, fuel exports, other uses, final supply equipment, organization snapshots, etc.). This excessive number of requests is hammering our database connections and creating performance bottlenecks.

**Purpose**
- **Performance**: Reduce database connection overhead and improve page load times
- **Scalability**: Prevent database connection pool exhaustion under high load
- **User Experience**: Faster compliance report loading with fewer network round trips
- **Resource Optimization**: More efficient use of server resources and bandwidth

**Acceptance Criteria**
- [ ] Analyze current API request patterns when opening compliance reports
- [ ] Design consolidated API endpoints that combine related data fetching
- [ ] Implement a single endpoint for all schedule data (fuel supplies, notional transfers, allocation agreements, fuel exports, other uses, final supply equipment)
- [ ] Implement a separate endpoint for compliance report metadata and summary information
- [ ] Update frontend hooks and components to use the new consolidated endpoints
- [ ] Ensure data structure maintains compatibility with existing UI components
- [ ] Add proper error handling and loading states for the consolidated requests
- [ ] Verify that all existing functionality continues to work correctly
- [ ] Measure and document performance improvements (request count reduction, load time improvement)
- [ ] Update API documentation to reflect the new endpoints

**Additional context**
- Current API requests observed when opening a compliance report:
  - `/api/fuel-supply/list-all`
  - `/api/notional-transfers/list-all`
  - `/api/allocation-agreement/list-all`
  - `/api/fuel-exports/list-all`
  - `/api/other-uses/list-all`
  - `/api/final-supply-equipments/list-all`
  - `/api/organization_snapshot/:reportID`
  - `/api/reports/:reportID/summary`
  - Multiple table-options endpoints for each schedule type
  - Various other supporting data requests

**Technical Considerations**
- Maintain backward compatibility during transition period
- Ensure proper caching strategies for the consolidated endpoints
- Consider pagination for large datasets within the consolidated response
- Evaluate impact on existing frontend state management and React Query cache keys
- Review database query optimization opportunities when fetching related data together

**Estimated Impact**
- Reduce API requests from 14+ to 2-3 requests per compliance report load
- Improve page load performance by reducing network overhead
- Reduce database connection pressure and improve overall system scalability

PR #2789, opened 2025-06-07
Feat: LCFS - Optimize Compliance Report Data Loading - Reduce Database Connection Hammering

prv-proton:
#2733 
- Add caching and stale time to all retrieval react query hook components and invalidating in case of updates.
- Reduce notification api call to get count on every click, instead set fetch retrieval interval to 1 min.
- use zustand to store the compliance report fetch, as it is being use in multiple components.
- Modify in all ReportDetails child components to fetch the compliance report data from zustand store instead of making api calls.
- Since list-all API is fetching all rows for the schedule summary and changelog components, implement client-side pagination logic for them.
- Fix getRowID issues in FuelExports, OtherUses of AddEdit components.


frontend/src/hooks/useCurrentUser.js
```diff
@@ -11,22 +11,29 @@ export const useCurrentUser = () => {
...
+      setUser(userData) // Set user data in store on successful fetch
+      return userData
     },
     enabled: !!keycloak.authenticated && initialized,
     retry: false,
-    onSuccess: setUser,
-    onError: (error) => {
-      console.error('Error fetching current user:', error)
-    }
+    staleTime: 5 * 60 * 1000, // 5 minutes - data stays fresh for 5 minutes
```
AlexZorkin, line 23: hopefully this fixes it! 👍

AlexZorkin (approved):
This is quite comprehensive so I think we will have to all test this in dev and see how it performs. Nice overhaul 👍

PR #2800, opened 2025-06-10
Fix: Notional transfer summary calculations

AlexZorkin:
- fixed notional transfer quarterly summary calculations
- tests updated


kevin-hashimoto (approved):
LGTM 👍🏻

PR #2816, opened 2025-06-11
Fix: LCFS - BUG - Empty schedules #2690

### LINKED ISSUE #2690: LCFS - BUG - Empty schedules
**Describe the Bug**  
When data is entered into one schedule, all other schedules unexpectedly appear as empty. This behavior is a recent change and was not the original intended functionality.

**Expected Behaviour**  
Schedules should retain and display their respective data independently of each other. Entering data into one schedule should not impact the visibility of other schedules.

**Actual Behaviour**  
Once data is added to any one schedule, all other schedules appear as empty.

Image

Image

prv-proton:
Fix #2690


AlexZorkin (approved):
👍

PR #2859, opened 2025-06-12
Feat: CR Excel Download Updates - 2774

AlexZorkin:
- Updated column sizing logic to be more dynamic
- Updated quarterly reports to display quarterly columns in download
- Tests


prv-proton (approved):
Looks good 👍

PR #2860, opened 2025-06-12
Feat: Transaction view status column refinement - 2624

### LINKED ISSUE #2624: LCFS - Transaction view table refinement
**Describe the task**
The status column on the transaction view table is not visible at first until you scroll across the AG grid. The Status column is maybe the most important information for transactions when we are reviewing them. It would be better to have this column be the first column on the furthest left side of the view table, similar to how other tables in the Portal are arranged. We could also reduce the width of the ID and Year columns to fit more information on the initial view of the table

Wireframe: 

Transaction view table refinement
(https://preview.uxpin.com/e0e2f15fc1b52987fd0287ad8c8103452dd16d60#/pages/206345755?mode=dm)

**Purpose**
Gives the most important information to the user in the first column to the left of the table and avoids them having to scroll across the page to see the status of the transaction.

**Acceptance Criteria**
- [ ] Move the "Status" column in the Transactions table to the far left.
- [ ] Decrease the width of other columns to try to maximize the visibility of all transactions information. Specifically the ID and year columns can be narrower as the information in these is not as long. 
- [ ] Ensure readability across the table is still clear after adjustments

**Additional context**
- We may want to further consider refining some of the other tables in the Portal as well. If a table includes a status, the status should always be first to the left.

hamed-valiollahi:
This PR includes the following changes to improve usability of the Transactions table:

- Moved the "Status" column to the far left to make it immediately visible.
- Reduced width of "ID" and "Year" columns to optimize table layout.
- Moved the "Status" column to the far left in the Compliance Report list.

Closes #2624

### DIFF
--- a/frontend/src/views/ComplianceReports/components/_schema.jsx
+++ b/frontend/src/views/ComplianceReports/components/_schema.jsx
@@ -9,6 +9,40 @@ import { useGetComplianceReportStatuses } from '@/hooks/useComplianceReports'
 
 export const reportsColDefs = (t, isSupplier) => [
   {
+    field: 'status',
+    headerName: t('report:reportColLabels.status'),
+    width: 300,
+    valueGetter: ({ data }) => data.reportStatus || '',
+    filterParams: {
+      textFormatter: (value) => value.replace(/\s+/g, '_').toLowerCase(),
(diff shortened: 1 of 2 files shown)


prv-proton (approved):
Looks good 👍

PR #2870, opened 2025-06-13
Fix: LCFS - BUG - Role change refresh #2868

### LINKED ISSUE #2868: LCFS - BUG - Role change refresh
**Describe the Bug**  
When an IDIR or BCeID user's role is changed, the interface does not refresh immediately to reflect the new role. As a result, the user may not see the correct features, buttons, or permissions until they manually refresh the page or log out and back in.

**Expected Behaviour**  
After a user role is updated and saved, the UI should automatically refresh to reflect the new role and its associated features and access levels.

**Actual Behaviour**  
The interface does not refresh on role change. Users remain in their previous role view until a manual refresh is performed, which can lead to confusion or restricted access.

**Implications**  
- Users may not see new buttons or features appropriate to their updated role.  
- Could result in miscommunication, failed workflows, or incorrect assumptions about permissions.  
- Inconsistent behavior undermines the integrity of role-based access control.

**Steps To Reproduce**  
1. Log in as an IDIR or BCeID user.  
2. Navigate to the Administration > Users section.  
3. Change the role of a user and click Save.  
4. Observe that the interface does not update to reflect the new role.

**Notes**  
- Implement an automatic refresh or re-render on the Save action for role changes.  
- Confirm that the refreshed UI reflects all permissions and controls specific to the new role.  
- Validate behavior across both IDIR and BCeID environments.

prv-proton:
Fix #2868
Migrate react queries to hooks in AddEditUser

### DIFF
--- a/frontend/src/hooks/useUser.js
+++ b/frontend/src/hooks/useUser.js
@@ -166,7 +166,7 @@ export const useCreateUser = ({
       })
 
       // Invalidate organization users query if in supplier context
-      if (isSupplierContext && organizationId) {
+      if (isSupplier && organizationId) {
         queryClient.invalidateQueries({
           queryKey: ['organization-users', organizationId]
         })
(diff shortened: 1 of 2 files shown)


AlexZorkin (approved):
Looks good, thanks for the quick fix 👍

PR #2878, opened 2025-06-14
Feat: Accordion Default Expansion

AlexZorkin:
- Added logic to expand accordions after editing schedule information


prv-proton (approved):
Looks great 👍 , especially the alert about which schedule was updated.

Issue #2826, opened 2025-06-11
Metabase - Refine the "Vw Fuel Supply Analytics Base" database view


Grulin:
**Describe the task**
Refine and expand the "Vw Fuel Supply Analytics Base" view in Metabase to include a more comprehensive set of columns. Each line of reported fuel supply should include all data from the Fuel Supply section of the compliance report, as well as extended Fuel Code information for fuel reported using a code.

**Purpose**
To provide a complete and accurate dataset for analysis and reporting by including detailed fuel supply and fuel code attributes. This will support better query design, increase data usability for compliance evaluation, and align analytics outputs with the structure of compliance reports.

**Acceptance Criteria**

- [ ]  Modify the "Vw Fuel Supply Analytics Base" to include the following columns:
Report Status
Compliance Report ID
Organization ID
Supplier Name
Compliance Year
Compliance Units
Fuel Type
Fuel Category
End Use
Determining Carbon Intensity
Quantity Supplied
Units
Target CI
RCI
UCI
Energy Density
EER
Energy Content
Fuel Code
Fuel Code Company
Feedstock
Feedstock Location
Effective Date
Expiry Date
Misc.
Fuel Production Facility City
Fuel Production Facility Province
Fuel Production Facility Country
Facility Nameplate Capacity
Facility Nameplate Capacity Unit
Facility Nameplate Transport Mode
Finished Fuel Transport Mode
- [ ]  Validate that all new fields accurately reflect the source data from the LCFS Portal.
- [ ]  Ensure each row corresponds to a single, line-item entry from the compliance report.
- [ ]  Maintain performant query execution despite expanded view size.
- [ ]  Provide documentation or column definitions for the updated view.
- [ ]  Test against recent compliance report data to confirm integrity and completeness.
- [ ] Leave blank any fuel code info for default determining carbon intensity, but they must have the fuel supply info like RCI and TCI and EER and Energy content etc. 

**Additional context**

- This update will serve as a foundation for advanced queries and dashboards in Metabase.
- Helps eliminate data silos and manual lookups by analysts.
-

PR #2892, opened 2025-06-17
Metabase: Refine the Vw Fuel Supply Analytics Base database view #2826 

prv-proton:
- #2890 
- drop views before creating
- fix base view of compliance reports.


AlexZorkin (approved):
👍

PR #2910, opened 2025-06-18
Fix:  LCFS - BUG - Download to Excel 500 error - #2905

### LINKED ISSUE #2905: LCFS - BUG - Download to Excel 500 error
**This might just be an issue on the dev site.**
Compliance report work in test but transaction gives 405 error.
Image

**Describe the Bug**  
All "Download to Excel" features across the platform are currently throwing a 500 Internal Server Error, preventing users from exporting data.

**Expected Behaviour**  
Clicking any "Download to Excel" button should generate and download an Excel file containing the appropriate data without any server errors.

**Actual Behaviour**  
A 500 error is returned whenever a user attempts to use the "Download to Excel" feature, regardless of the page or data context.

**Implications**  
This issue blocks users from exporting critical reports and data sets for offline analysis or record-keeping. It may disrupt compliance processes and reduce trust in the platform's reliability.

**Steps To Reproduce**  
User/Role: Any user with access to export functionality  
1. Navigate to any page with a "Download to Excel" feature (e.g., Transactions, Compliance Reports).  
2. Click the "Download to Excel" button.  
3. Observe that a 500 error is triggered instead of an Excel download.

**Notes**  
- Investigate backend endpoints and server logs to determine the root cause of the 500 error.  
- Confirm whether the error is due to recent code changes, data format issues, or environment-specific configurations.  
- Test thoroughly across all modules that include Excel download functionality after applying a fix.

prv-proton:
Fixes #2905

### DIFF
--- a/backend/lcfs/web/api/organization/views.py
+++ b/backend/lcfs/web/api/organization/views.py
@@ -181,15 +181,14 @@ async def get_transactions_paginated_for_org(
 async def export_transactions_for_org(
     request: Request,
     format: str = Query(default="xlsx", description="File export format"),
+    pagination: PaginationRequestSchema | None = Body(None),
     txn_service: TransactionsService = Depends(),
 ):
     """
     Endpoint to export information of all transactions for a specific organization
     """
     organization_id = request.user.organization.organization_id
-    return await txn_service.export_transactions(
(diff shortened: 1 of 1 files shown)


hamed-valiollahi (approved):
Looks good to me! Thanks for the fix

PR #2922, opened 2025-06-19
Fix: Supplemental Txn Logic - 2903 2904

AlexZorkin:
- transaction released and reinstated on superseded supplemental reports
- tests


prv-proton (approved):
Looks good 👍

PR #2926, opened 2025-06-19
Feat: Prepend C- Prefix to Canadian Fuel Codes - 2540

### LINKED ISSUE #2540: LCFS - Canada fuel codes
**Description:**  
Update the fuel code select list in all schedules that use fuel codes to prepend the text "(C-)" to the fuel code prefix when the Fuel Production Facility Country is equal to "Canada". This should be applied to 2025 reports forward. This change applies to the following schedules: Fuel Supply, Export Fuels, Fuels for Other Use, and Allocation Agreements. Example: "BCLCF" should display as "C-BCLCF" when the facility country is Canada.

**Wireframe:**  
N/A – Applies to dropdown/select list rendering in the schedules listed above.

Improves visibility of domestic fuel production within selection interfaces, supporting clearer user input and enabling accurate identification of Canadian-produced fuel codes at the point of data entry.

**Acceptance Criteria:**  
- [ ] In all relevant schedules (Fuel Supply, Export Fuels, Other Use, Allocation Agreements) for 2025 compliance reports and later, the fuel code select list displays a "C-" prefix for fuel codes where the production facility country is "Canada".  
- [ ] If the production facility country is not Canada, the code appears without the "C-" prefix.  
- [ ] Fuel codes are matched accurately regardless of case sensitivity in country name.  
- [ ] Prefix formatting is consistent and does not duplicate or conflict with other display logic.  
- [ ] The displayed value does not affect the stored or transmitted value unless explicitly required elsewhere.

**Development Checklist:**  
- [ ] Modify select list rendering logic to check facility country and format prefix accordingly.  
- [ ] Validate behavior across all four schedule forms in 2025 compliance reports and later.  
- [ ] Add test cases for Canadian and non-Canadian fuel codes.  
- [ ] Ensure the new format is applied consistently across reactive form refreshes and page loads.  
- [ ] Review with stakeholders to confirm visual intent and impact on downstream processing (if any).

**Notes:**  
- This is a display-only change in select lists; ensure internal identifiers remain unmodified unless a backend change is separately required.  
- Future enhancements may include other country-based prefixes—structure logic to accommodate extensibility if needed.

hamed-valiollahi:
This PR includes the following:

- Prepends `C-` to fuel codes when the facility country is Canada
- Applies to all 2025+ schedule dropdowns

Closes #2540


AlexZorkin (changes requested):
This is looking good but please add frontend unit tests for this new logic.

AlexZorkin:
Not sure where we landed with the decision on frontend or backend solution. Where are we at with this pr?

hamed-valiollahi:
> Not sure where we landed with the decision on frontend or backend solution. Where are we at with this pr?

Yeah, I’m not sure either.

AlexZorkin (approved):
Solution looks good 👍

hamed-valiollahi:
> Solution looks good 👍

Thanks Alex for the code review

Issue #2895, opened 2025-06-17
LCFS - Supplemental draft


airinggov:
**Description:**  
When a supplemental report is created, capture and store the Draft creation event — including the user, date/time stamp — as a permanent entry in the report history. If the draft is created by an IDIR user, it will display as "Government of B.C." to the BCeID user. This permanent record is only applicable to Supplemental reports due to the 30-day completion requirement and is not needed for other report formats. Much of this logic is already in place, we just need to ensure the permanence of history display.

The IDIR Compliance Report index page must display all draft supplemental reports, regardless of whether they were created by BCeID or IDIR users.

**Wireframe:**  
N/A – UI impacts limited to report history log display and Compliance Report index visibility for IDIR users.

**Purpose and benefit to user:**  
Ensures there is a traceable record of when a supplemental report draft was initiated, supporting accountability and compliance with the 30-day timeline. Enables IDIR users to track all in-progress supplemental reports for monitoring and follow-up.

**Acceptance Criteria:**  
- [x] On supplemental report creation, log a Draft creation entry with:
    - [ ] Username and role
    - [x] Date/time stamp
    - [x] Label as “Government of B.C.” on the BCeID side if created by IDIR.
- [x] Ensure the Draft creation event is permanently stored in the report’s history log
- [ ] Display all draft supplemental reports in the IDIR Compliance Report index, regardless of creator role

**Development Checklist:**  
- [ ] Update backend logic to log draft creation as a report history line specific to supplemental reports  
- [ ] Adjust front-end history display to render user role labeling correctly for BCeID views  
- [ ] Ensure filtering logic on the IDIR Compliance Report index includes all draft supplemental reports  
- [ ] Add tests for report creation and history persistence  
- [ ] Validate history integrity across report lifecycle events

**Notes:**  
- Existing logic for draft creation and user roles may be reused; ensure persistence is implemented.

airinggov:
This has been implemented with the original method of showing a Supplemental requested status on the IDIR side when created by the analyst. We still might consider changing it to behave closer to what this card is documenting.

PR #2955, opened 2025-06-23
Feat: LCFS - Supplemental draft #2895

prv-proton:
#2895


backend/lcfs/db/migrations/versions/2025-06-20-16-41_413eef467edd.py
```diff
@@ -0,0 +1,85 @@
...
+
+
+def upgrade() -> None:
+    # Step 1: Add the new enum value and commit it
+    connection = op.get_bind()
+
+    # Add the enum value in its own transaction
+    connection.execute(
+        sa.text(
+            "ALTER TYPE compliancereportstatusenum ADD VALUE IF NOT EXISTS 'Supplemental_requested'"
```
AlexZorkin, line 45: So this an official new status, where does it get set?

backend/lcfs/web/api/compliance_report/services.py
```diff
@@ -430,7 +430,7 @@ async def create_government_initiated_supplemental_report(
             reporting_frequency=current_report.reporting_frequency,
             compliance_report_group_uuid=current_report.compliance_report_group_uuid,  # Same group
             version=new_version,
-            supplemental_initiator=SupplementalInitiatorType.SUPPLIER_SUPPLEMENTAL,  # Supplier edits it
+            supplemental_initiator=SupplementalInitiatorType.GOVERNMENT_INITIATED,  # Supplier edits it
```
AlexZorkin, line 433: update comment explaining the purpose here

AlexZorkin (changes requested):
Looks good, just two minor questions.

backend/lcfs/db/migrations/versions/2025-06-20-16-41_413eef467edd.py
prv-proton: We won't be setting this in any report directly, but since the index page gets it's data from the  compliance_report view, I'm just setting this in the database view only. But in edit/view of the compliance report would still continue to have the original status.

AlexZorkin: 👍

backend/lcfs/web/api/compliance_report/services.py
AlexZorkin: 👍

AlexZorkin (approved):
Looks good ty 👍

PR #2963, opened 2025-06-25
LCFS - BUG - Other fuel, Energy density

### LINKED ISSUE #2919: LCFS - BUG - Other fuel, Energy density
**Describe the Bug**  
In both the Fuel Supply and Export Fuel data entry schedules, when "Other" is selected as the fuel type, the Energy Density field incorrectly allows the default value of zero to pass validation. Users should be required to input a valid Energy Density greater than zero.

**Expected Behaviour**  
When "Other" fuel type is selected:
- The Energy Density field must not accept a value of zero.
- Validation should require users to enter a value greater than zero before saving.

**Actual Behaviour**  
Energy Density is currently validated as acceptable when left at the default value of zero, even when "Other" fuel type is selected.

**Implications**  
- Zero values may result in invalid or misleading compliance calculations.  
- Undermines data integrity and could impact downstream reporting.  
- Users may not realize the requirement to manually input a valid energy density.

**Steps To Reproduce**  
User/Role: Any user entering Fuel Supply or Export Fuel data  
1. Add a new row in the Fuel Supply or Export Fuel schedule.  
2. Select "Other" as the fuel type.  
3. Leave the Energy Density value at zero.  
4. Save the row and observe that it passes validation without error.

**Notes**  
- Add validation logic to enforce Energy Density > 0 for “Other” fuel types.  
- Ensure this rule is applied consistently across all relevant schedules.  
- Include a tooltip or validation message to inform users of the requirement.

kevin-hashimoto:
- fix: fuel type other energy density non 0, closes #2919


hamed-valiollahi (approved):
Looks good to me! Great test coverage too. 👍

PR #2964, opened 2025-06-25
LCFS - BUG - Other use input

### LINKED ISSUE #2930: LCFS - BUG - Other use input
**Describe the Bug**  
When the Other Use schedule input form loads, it displays two empty rows by default and incorrectly makes the second row active. This behavior is inconsistent with other schedules, which initialize with only one active row.

**Expected Behaviour**  
On initial load, the Other Use schedule should:
- Display only one empty row
- Set the first row as the active input row

**Actual Behaviour**  
Two empty rows appear upon loading, and the second row is unexpectedly activated.

**Implications**  
- Introduces UI inconsistency across schedules  
- Can confuse users who expect to begin entry in the first row  
- May cause unintentional data entry in the wrong row

**Steps To Reproduce**  
User/Role: Any user accessing the Other Use schedule  
1. Open a new compliance report  
2. Navigate to the Other Use schedule  
3. Observe that two empty rows are displayed and the second row is active by default

**Notes**  
- Update row initialization logic to match the behavior of other schedules  
- Test across BCeID and IDIR roles to ensure consistent experience  
- Validate that only the first row is focused and editable on initial load

kevin-hashimoto:
- fix: double row other uses bug, closes #2930


hamed-valiollahi (approved):
Looks good to me! Tested it locally and it works as expected. 👍

PR #2968, opened 2025-06-27
Fix: Hide Action Buttons for IDIR Analysts - 2940

### LINKED ISSUE #2940: LCFS - Action buttons disabled
**Description:**  
When an IDIR analyst creates or returns a supplemental report, the action buttons are displayed but disabled. To improve clarity and user experience, these buttons should be completely hidden when the user does not have permission to use them in the current report state.

**Wireframe:**  
Image

Hiding inactive action buttons eliminates confusion and better reflects the user's current permissions and the report's workflow state. This improves usability and aligns with interface behavior across other sections of the application.

**Acceptance Criteria:**  
- [ ] When an IDIR analyst creates a new supplemental report, all action buttons for which they lack permissions should be hidden (not disabled).  
- [ ] When an IDIR analyst returns a supplemental report, they should also not see disabled action buttons; these should be hidden entirely.  
- [ ] Visibility logic should be consistent across all report states and aligned with role-based access controls.  
- [ ] Confirm no impact on BCeID user button visibility or behavior.

**Development Checklist:**  
- [ ] Identify all conditions under which action buttons appear as disabled for IDIR analysts.  
- [ ] Refactor UI to hide buttons instead of disabling them in those scenarios.

hamed-valiollahi:
This PR hides action buttons for IDIR users in supplemental reports when they lack permission.

Closes #2940


frontend/src/views/ComplianceReports/buttonConfigs.jsx
```diff
@@ -516,8 +527,8 @@ function shouldShowButton(buttonName, context) {
       return !context.isEarlyIssuance
 
     case 'createReassessment':
-      // Disabled if there's a conflicting supplemental
-      return !context.hasDraftSupplemental
+      // Additional condition: only show for specific conditions beyond draft supplemental check
```
prv-proton, line 530: since createReassessment button is handled in a different component, is this change necessary?

hamed-valiollahi: You're absolutely right. I've removed the unnecessary code. Thanks for catching that!

PR #2969, opened 2025-06-27
Fix: Auto-expand file attachments accordion after upload - 2917

### LINKED ISSUE #2917: LCFS - BUG - File attachments accordion
**Describe the Bug**  
After uploading file attachments, the file attachment accordion remains collapsed, requiring the user to manually expand it to view the uploaded files. This extra step can be confusing and causes users to miss confirmation that their files were added successfully.

**Expected Behaviour**  
Once file attachments are added, the file attachment accordion should automatically expand to display the uploaded files without requiring additional user interaction.

**Actual Behaviour**  
The accordion stays collapsed after file upload, forcing the user to expand it manually to confirm the attachments are present.

**Implications**  
- Users may think their file uploads failed if they do not expand the accordion.  
- This creates unnecessary friction in the workflow and may result in duplicate uploads or user support requests.

**Steps To Reproduce**  
User/Role: Any user with access to file attachment functionality  
1. Upload a file in a section where attachments are added via an accordion UI.  
2. Observe that the accordion remains collapsed.  
3. Manually expand the accordion to confirm the file was uploaded.

**Notes**  
- Update UI logic to ensure the accordion expands automatically after a file is added.  
- Test the behavior across all relevant modules where file attachments are supported.

hamed-valiollahi:
This PR includes the fix to auto-expand the attachments accordion after uploading a file.  

Closes #2917

### DIFF
--- a/frontend/src/views/ComplianceReports/EditViewComplianceReport.jsx
+++ b/frontend/src/views/ComplianceReports/EditViewComplianceReport.jsx
@@ -613,32 +613,30 @@ export const EditViewComplianceReport = ({ isError, error }) => {
               chain={reportData?.chain}
             />
           </Stack>
-          {!location.state?.newReport && (
-            <>
-              <ReportDetails
+          <>
+            <ReportDetails
+              canEdit={canEdit}
+              currentStatus={currentStatus}
+              hasRoles={hasRoles}
+              complianceReportData={reportData}
+            />
+            {!showEarlyIssuanceSummary && (
+              <ComplianceReportSummary
+                reportID={complianceReportId}
+                enableCompareMode={reportData?.chain?.length > 1}
                 canEdit={canEdit}
                 currentStatus={currentStatus}
-                hasRoles={hasRoles}
-                complianceReportData={reportData}
+                compliancePeriodYear={compliancePeriod}
+                setIsSigningAuthorityDeclared={setIsSigningAuthorityDeclared}
+                buttonClusterConfig={buttonClusterConfig}
+                methods={methods}
+                alertRef={alertRef}
               />
-              {!showEarlyIssuanceSummary && (
-                <ComplianceReportSummary
-                  reportID={complianceReportId}
-                  enableCompareMode={reportData?.chain?.length > 1}
-                  canEdit={canEdit}
-                  currentStatus={currentStatus}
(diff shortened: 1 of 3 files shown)


frontend/src/views/ComplianceReports/components/ReportDetails.jsx
```diff
@@ -518,6 +524,24 @@ const ReportDetails = ({ canEdit, currentStatus = 'Draft', hasRoles }) => {
     []
   )
 
+  const prevHasDataRef = useRef(new Map())
+
+  useEffect(() => {
```
prv-proton, line 529: @hamed-valiollahi, there's already an existing pattern implemented in this PR for handling this requirement in other schedules. You could consider following a similar approach for supporting documents to maintain consistency..
https://github.com/bcgov/lcfs/pull/2878/files

hamed-valiollahi: Great! I reviewed the PR and it uses a much smarter pattern. I'll update the code. Thanks for pointing this out.

hamed-valiollahi: Thanks for the review, Prashanth. I’ve updated the code to follow the same pattern we’re using in the other schedules. Let me know if you see any issues or if there’s anything else I should adjust.

frontend/src/views/ComplianceReports/EditViewComplianceReport.jsx
```diff
@@ -613,32 +613,30 @@ export const EditViewComplianceReport = ({ isError, error }) => {
               chain={reportData?.chain}
             />
           </Stack>
-          {!location.state?.newReport && (
```
prv-proton, line 616: removing this would show ComplianceReportSummary panel to the users. I think when a new Draft report is created we should only be showing introduction panel that's the reason this check was added.

hamed-valiollahi: Thanks for the information. I kept the ComplianceReportSummary hidden for new draft reports, but made the ReportDetails visible so the supporting documents section shows up. This way, as soon as the user uploads a file, they can see it in draft mode.

hamed-valiollahi:
> Looks good 👍

Thanks Prashanth for the code review.

PR #2971, opened 2025-06-27
fix: remove nulls from potential recipients

AlexZorkin:
- discovered a bug where nulls were being included in some recipient lists and failing the email send.


prv-proton (approved):
Looks good 👍

PR #2983, opened 2025-07-02
Feat: Python based migrations for TFRS Shutdown

AlexZorkin:
- added python based migrations for tfrs shutdown based on groovy scripts
- allows for faster testing and resetting of database


backend/lcfs/db/migrations/versions/2025-07-18-15-25_87592f5136b3.py
```diff
@@ -0,0 +1,218 @@
...
+# revision identifiers, used by Alembic.
+revision = "87592f5136b3"
+down_revision = "5c3de27f158b"
+branch_labels = None
+depends_on = None
+
+
+def upgrade():
+    # Insert historical gasoline values
+    op.execute(
```
prv-proton, line 21: When credit calculator is used to find Credits before 2024 compliance year. did isn't showing up.

AlexZorkin: ok we can followup with this as a separate ticket

PR #3010, opened 2025-07-15
Feat: Fuel code workflow notifications - 2589

AlexZorkin:
- added 3 new notifications that trigger on fuel code status changes
- tests for new email templates and frontend


backend/lcfs/web/api/fuel_code/services.py
```diff
@@ -255,32 +274,113 @@ async def create_fuel_code(
...
+            action_type=fuel_code.action_type,
         )
+        await self.repo.create_fuel_code_history(history)
+        updated_fuel_code = await self.repo.update_fuel_code(fuel_code)
+
+        # Send notifications based on status change
+        notifications = []
+
+        # Check for specific status transitions and new statuses
+        if status == FuelCodeStatusEnum.Recommended:
```
prv-proton, line 335: may need to use FuelCodeStatusEnumSchema instead, as it could cause mismatch.

prv-proton (changes requested):
Thanks for the update, overall looks good 👍, just minor fixes.

backend/lcfs/db/migrations/versions/2025-07-22-07-37_674478937015.py
```diff
@@ -0,0 +1,46 @@
...
+from alembic import op
+
+# revision identifiers, used by Alembic.
+revision = "674478937015"
+down_revision = "18ca5084ea87"
+branch_labels = None
+depends_on = None
+
+
+def upgrade():
```
prv-proton, line 19: During the code merges some of the changes that was done on vw_fuel_code_base view is lost and it's causing issue with index page. You might want to consider to revert these changes back from below 
https://github.com/bcgov/lcfs/blob/e3c1d9e98ef0899a3b73a58db18915ac9c2cdd25/backend/lcfs/db/sql/views/metabase.sql

prv-proton (changes requested):
Since related organization is now set to undefined for fuel codes in the notifications, it's causing issue once the fuel-code notifications are there in the notifications index page.
image

may need to handle this at else condition in /lcfs/frontend/src/views/Notifications/NotificationMenu/components/_schema.jsx

`
  {
    colId: 'organization',
    field: 'organization',
    headerName: t('notifications:notificationColLabels.organization'),
    valueGetter: (params) => {
      const { service, toOrganizationId, fromOrganization } = JSON.parse(
        params.data.message
      )
      if (
        service === 'Transfer' &&
        toOrganizationId === currentUser?.organization?.organizationId
      ) {
        return fromOrganization
      }
      return params.data.relatedOrganization.name
    }
  }
`

backend/lcfs/db/migrations/versions/2025-07-22-07-37_674478937015.py
AlexZorkin: ok thanks

prv-proton (approved):
👍

PR #3013, opened 2025-07-15
fix: changelog errors

kevin-hashimoto:
fixes errors related to compliance report changelog


AlexZorkin (changes requested):
So main issue was missing joins? just need to make sure the backend tests are passing correctly.

AlexZorkin (approved):
👍

PR #3049, opened 2025-07-25
Feat: big int quantity fields - 2668

AlexZorkin:
- Updated all quantity fields in the app to use BigInteger
- Migration drops and recreates views in order to change type


prv-proton (approved):
Looks good 👍

PR #3075, opened 2025-07-31
Fix LCFS - BUG for downloading fuel code info to Excel #3070

### LINKED ISSUE #3070: LCFS - BUG for downloading fuel code info to Excel
**Describe the Bug**
When an IDIR user attempts to download the fuel code information into an Excel spreadsheet from the LCFS Portal, the system generates an error message and the download does not complete. The error appears consistently regardless of browser or device used.

**Expected Behaviour**
The fuel code information should successfully download into an Excel file without error when initiated by an IDIR user. The file should be complete, correctly formatted, and accessible.

**Actual Behaviour**
Instead of generating a usable Excel file, the portal produces an error message. See screenshot: 

Image

**Implications**
IDIR users cannot access fuel code data in Excel format, impeding compliance monitoring and enforcement work.

**Steps To Reproduce**
User/Role:

1. Log into the LCFS Portal using an IDIR account.
2. Navigate to the fuel code information section.
3. Select the option to download the data in Excel format.
4. Observe the error message and failed download.

prv-proton:
Fixes #3070

### DIFF
--- a/backend/lcfs/tests/fuel_code/test_fuel_code_exporter.py
+++ b/backend/lcfs/tests/fuel_code/test_fuel_code_exporter.py
@@ -16,8 +16,8 @@ async def test_export_success():
 
     mock_fuel_codes = [
         MagicMock(
-            fuel_code_status=MagicMock(status=FuelCodeStatusEnum.Draft),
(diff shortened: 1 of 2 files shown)


hamed-valiollahi (approved):
Looks good to me! 👍

PR #3099, opened 2025-08-06
Fix: LCFS - Bug Fix: to_transaction_id not being persisted when transfers are recorded + data backfill

### LINKED ISSUE #3067: LCFS - Bug Fix: to_transaction_id not being persisted when transfers are recorded + data backfill
**Describe the task**
Fix a bug where `to_transaction_id` is not being properly persisted in the Transfer table when transfers are recorded. Additionally, create a data recovery script to backfill the missing `to_transaction_id` values for already recorded transfers.

**Purpose**
When a transfer is recorded, the system creates a transaction for the receiving organization but fails to persist the `to_transaction_id` reference in the Transfer table. This breaks the bidirectional link between transfers and transactions, causing a loss of transaction traceability for receiving organizations.

**Root Cause Analysis**
In `backend/lcfs/web/api/transfer/services.py`, the `director_record_transfer` method (line 428):
1. Creates a new transaction for the receiving organization (line 478-482)
2. Sets `transfer.to_transaction = to_transaction` (line 483)
3. Calls `update_transfer(transfer)` (line 485)

However, the `update_transfer` method in the repository only calls `flush()` without explicitly committing the `to_transaction_id` field change, causing the foreign key reference to not be persisted.

**Acceptance Criteria**
- [ ] Fix the bug in `director_record_transfer` to ensure `to_transaction_id` is properly persisted
- [ ] Add explicit field assignment: `transfer.to_transaction_id = to_transaction.transaction_id` before updating
- [ ] Verify the repository's `update_transfer` method properly commits the changes
- [ ] Create a data recovery script/migration to backfill missing `to_transaction_id` values
- [ ] Add unit tests to verify `to_transaction_id` is set when transfers are recorded
- [ ] Verify all recorded transfers have both `from_transaction_id` and `to_transaction_id` after the fix

**Data Recovery Strategy**
For the backfill script, identify orphaned transactions by:
1. Find all transfers with `current_status_id = 6` (Recorded) and `to_transaction_id IS NULL`
2. For each transfer, find the matching transaction by:
   - `organization_id = transfer.to_organization_id`
   - `transaction_action = 'Adjustment'`
   - `compliance_units = transfer.quantity` (positive value)
   - Transaction `create_date` matches the transfer's recorded date (from transfer_history or update_date)
   - Transaction is not already linked to another transfer
3. Update the transfer with the matched `transaction_id`
4. Log any transfers that cannot be matched for manual review

**Additional context**
- The bug is in: `backend/lcfs/web/api/transfer/services.py:483-485`
- Transfer model: `backend/lcfs/db/models/transfer/Transfer.py`
- Repository: `backend/lcfs/web/api/transfer/repo.py`
- Recorded status ID = 6 (confirmed in migrations and views)
- The `to_transaction_id` field is nullable, which allowed this bug to go unnoticed
- SQL views like `vw_transfer_base` rely on both transaction IDs being present

prv-proton:
Fixes #3067

### DIFF
--- a/backend/lcfs/db/migrations/versions/2025-08-14-13-28_e5799ecd8dda.py
+++ b/backend/lcfs/db/migrations/versions/2025-08-14-13-28_e5799ecd8dda.py
@@ -1,7 +1,7 @@
 """backfill to_transaction_id in transfers table
 
 Revision ID: e5799ecd8dda
-Revises: 3f5a7b9c1d2e
+Revises: 2b82ade06987
 Create Date: 2025-08-06 13:28:05.038494
 
 """
@@ -11,7 +11,7 @@ from alembic import op
 
 # revision identifiers, used by Alembic.
 revision = "e5799ecd8dda"
-down_revision = "3f5a7b9c1d2e"
+down_revision = "2b82ade06987"
 branch_labels = None
 depends_on = None
 


backend/lcfs/db/migrations/versions/2025-08-10-13-28_e5799ecd8dda.py
```diff
@@ -0,0 +1,157 @@
...
+        from_tx_create_date,
+    ) in transfers_to_fix:
+
+        # Use from_transaction create_date if available, otherwise use transfer create_date
+        reference_date = (
+            from_tx_create_date if from_tx_create_date else transfer_create_date
+        )
+
+        # Find the corresponding transaction for the receiving organization.
+        # Select the one with the closest timestamp to the reference date
```
AlexZorkin, line 70: 'closest timestamp' is concerning to me. It should be exact match or else it could choose the wrong transaction. Let's discuss this in more detail.

AlexZorkin (changes requested):
Fix looks good but the migration we should review to ensure we're getting the correct txn ids.

backend/lcfs/db/migrations/versions/2025-08-10-13-28_e5799ecd8dda.py
prv-proton: Thanks @AlexZorkin  for the review. Yes, it seemed bit concerning, but after going thru each transaction I found that they had a mismatch of a second or two between transfer create date and transaction create date, hence used closest date approach to resolve it. I did check the data in my local with prod data and there were around 128 record updates and all seemed ok, but we can go thru it once again if required.

PR #3142, opened 2025-08-12
Fix: line 6 persistence logic updates

AlexZorkin:
- Updated logic around line 6 calculation to properly cap based on line 4 volume amounts
- Updated tests


prv-proton (approved):
Looks good 👍

PR #3160, opened 2025-08-14
LCFS - Bug: Filters on Organization Page Cause 400 Errors and Table Disappearance

### LINKED ISSUE #3138: LCFS - Bug: Filters on Organization Page Cause 400 Errors and Table Disappearance
**Describe the Bug**
On Prod, when using filters or sorting on the Organization page, applying filters or sorting on the registration status or early issuance columns causes the page to break. Specifically, users receive 400 error messages and the data table disappears. Filters on status and organization name work correctly without issue.

See screenshot: 

Image

**Expected Behaviour**
Applying filters or sorting on any column, including registration status and early issuance, should update the table correctly without errors or the table disappearing.

**Actual Behaviour**
Filters and sorting on registration status or early issuance columns cause a 400 error response from the server and the table disappears from the page.

**Implications**
Users cannot effectively filter or sort on key columns, limiting usability and data access.

**Steps To Reproduce**
User/Role: IDIR

1. Navigate to the Organization page.
2. Try to sort or apply a filter on the registration status column.
3. Observe the 400 error and disappearance of the data table.
4. Repeat for the early issuance column filter or sorting.

Note that sorting/filtering on status or organization name does not cause errors.

kevin-hashimoto:
fix: filtering orgs by registration/early issuance causing 400 errors, closes #3138


backend/lcfs/web/api/organizations/services.py
```diff
@@ -87,10 +92,24 @@ def apply_organization_filters(self, pagination, conditions):
                 )
                 continue
 
-            if filter.field == "status":
+            if field_name == "registrationStatus":
```
prv-proton, line 95: might need to be in snake case instead of camel Case. Since registration_status is same as status field, can't we  reuse the status check on line 109?

prv-proton (changes requested):
still facing issues w.r.t registration status column, early_issuance looks good though

prv-proton (approved):
Looks good 👍

PR #3164, opened 2025-08-15
update readme in order to trigger the pipeline

kuanfandevops:


AlexZorkin (changes requested):
can we delete this now?

PR #3165, opened 2025-08-15
Feat: Line 7&9 Auto-populate and Locking - 2893

AlexZorkin:
- auto-populate and lock Lines 7 & 9 in compliance report summaries for 2025+ reports
- pulls in previous report values if exists, otherwise leaves fields editable


prv-proton (approved):
looks good 👍

PR #3172, opened 2025-08-19
Feat: 2015-2030 Lookup Table Values - 3157

AlexZorkin:
- migration updating lookup table values to 2030


backend/lcfs/db/migrations/versions/2025-08-19-10-52_988aeb471e14.py
```diff
@@ -0,0 +1,291 @@
...
+            WHERE ed.fuel_type_id = ft.fuel_type_id
+            AND ed.compliance_period_id = cp.compliance_period_id
+        );
+    """
+    )
+
+    # Extend energy_effectiveness_ratio table from 2026 (id=17) to 2030 (id=21)
+    op.execute(
+        """
+        INSERT INTO energy_effectiveness_ratio (
```
prv-proton, line 80: Query isn't inserting all the necessary rows into energy_effectiveness_ratio table for each compliance_period.

prv-proton (changes requested):
Thanks for the migration. Not all data is being inserted into required tables.

prv-proton (approved):
Looks good 👍

backend/lcfs/db/migrations/versions/2025-08-19-10-52_988aeb471e14.py
prv-proton, line 80: why some of the fuel types like "Biodiesel" is excluded here? Due to which eer isn't populated for them.

PR #3199, opened 2025-08-25
Fix: Supplemental Logic Updates

AlexZorkin:
- Moved summary locking back a status to Recommended by Analyst from Recommended by Compliance Manager
- Removed the re-calculate summary on Assessment to keep the summary consistent with what was recommended
- Exclude current report from compliance period lookups
- Updated non-assessment logic to lock summary at time of status change


prv-proton (approved):
Looks good 👍

PR #3247, opened 2025-09-05
Fix: add back missing triggers and one new transaction trigger - 3223

AlexZorkin:
- a previous migration cascaded and removed a set of triggers, this migration adds them back and one new one


prv-proton (approved):
👍

PR #3310, opened 2025-09-22
LCFS - Negative credit balance

### LINKED ISSUE #3262: LCFS - Negative credit balance
**Describe the Bug**
On test environment: Org 10  seems to be two issues with transfers and what happens to account balances:

1. A transfer after the 2024 compliance date while a report was awaiting to be assessed resulted in the credits from the transfer being applied like it was being applied to the 2024 report (although the report summary was fine).
2. After the report was assessed, due to net debits exceeding the credit balance, the organization went into negative compliance units in their account balance, which should not be possible.

**Expected Behaviour**
The assessed report should only pull credits available on or before the compliance date of March 31. In this case, the supplier had 50,000 credits when the report was submitted and the report resulted in net debits of 58,104 in 2024 reporting, the report should pull the 50,000 credits and confirm penalties related to 8,104 debits. 

A transfer of 5,000 credits was received by the organization after the 2024 compliance date. Any transfers after the compliance date for 2024 cannot be applied for 2024.

The account balance should reflect this.:
- The organization account balance should be 50,000 in reserve waiting for the report to be assessed. 
- After the transfer and still waiting for the report to be assessed, the organization should have 5,000 credits available to transfer and 50,000 in reserve. 
- After the report being assessed the organization should have 5,000 available to transfer and nothing else

**Actual Behaviour**
The report appears to have functioned appropriately.

The account balance however had issues:
- The organization account balance was 50,000 in reserve waiting for the report to be assessed. 
- After the transfer and still waiting for the report to be assessed, the organization had 55,000 in reserve (instead 5,000 available).
- After assessment the organization had negative compliance units of -3,104. (50,000 + 5,000 - 58,104).

**Implications**
This can create erroneous credit balances when suppliers make transfers before their report has been assessed. I think it would only affect those in a debit situation and do not have enough credits to make up the total

**Steps To Reproduce**
User/Role:
1. BCeID user: Create and submit compliance report (might need a supplemental as this was during a supplemental)
2. two BCeID users make a transfer with the receiving party being the one who submited the report
3. View the account balance to see if balance and reserve are incorrect
4. Approve the transfer
5. Reveiw account balance to see if incorrect

kevin-hashimoto:
fix: date attribution for transactions causing records to be wrongly attributed to reports from previous years, fixes #3262 

Prevents post-deadline transfers from incorrectly affecting prior compliance periods and eliminates negative balance scenarios.


AlexZorkin (approved):
logic looks good. the logger.debugs should be removed.

PR #3312, opened 2025-09-23
feat: Max Retention, Obligation/Deferment Summary Logic - 3265

AlexZorkin:
- Updated the backend summary service so Line 8 (obligation deferred) automatically zeroes out whenever renewable supply meets or exceeds the requirement, while still capping deferment at 5 % for shortfalls. Net
  Line 10 values now reflect the removed deferment instead of carrying stale volumes.
  - Exposed the previous-year caps and refined the frontend schema: Line 7 uses the backend-provided maxima only when Lines 7/9 are locked; first-year reports now have unrestricted inputs for gasoline,
  diesel, and jet fuel.
  - Added unit coverage for the new deferment behaviour and the unlocked Line 7 scenario to protect the new rules.


prv-proton (approved):
Looks good 👍

PR #3315, opened 2025-09-25
Fix: LCFS - Supplemental summary Line 17 not carrying previous report value #3289

### LINKED ISSUE #3289: LCFS - Supplemental summary Line 17 not carrying previous report value
### Describe the Bug
On the Low carbon fuel target summary table, Line 17 "Available compliance unit balance for the compliance period" shows 0 in the Supplemental column for Supplemental report 1. The Original Report shows a non-zero value (e.g., 12,215), and the Supplemental column should match this value when the supplemental is created.

### Expected Behaviour
When a supplemental report is created, Line 17 in the Supplemental column should carry forward the value from the Original Report for the same period, so the Delta remains 0 unless changes are made within the supplemental.

### Actual Behaviour
The Supplemental column for Line 17 is populated with 0, resulting in an incorrect negative Delta (e.g., -12,215) and cascading incorrect balances further down the summary (Lines 20–22).

Image

### Implications
- **Incorrect balances** and delta values on the summary
- **Potentially incorrect end-of-period available balance** and non-compliance calculations
- **Confusion for reporters and reviewers**; risk of erroneous assessments/approvals

### Steps To Reproduce
User/Role: Report submitter or reviewer
1. Create a Supplemental report for a compliance period where the Original Report has a non-zero Line 17 value.
2. Navigate to the Summary view.
3. Locate Line 17: "Available compliance unit balance for the compliance period".
4. Observe that the Supplemental column shows 0 while the Original Report shows a non-zero value, producing an incorrect negative Delta.

### Additional context
This appears to be a carry-forward/initialization bug where the Supplemental report is not pulling the prior report's value for Line 17.

### Suggested Priority
High — impacts core balances and can misrepresent compliance status.

prv-proton:
Fix #3289


AlexZorkin (approved):
This was definitely missing. Good fix. 👍

PR #3317, opened 2025-09-26
Fix: Auth connection issue fix.

prv-proton:
- Updated the authentication workflow to resolve signing keys directly from the cached JWKS payload, handle missing or rotated keys, and continue JWT validation without additional network calls to avoid connection failures.


AlexZorkin (approved):
nice optimization 👍

PR #3338, opened 2025-10-14
Fix: Updated Lessor Retention Logic

AlexZorkin:
- line 10 and 4 used to calculate lessor of retention and deferral amounts
- tests updated


prv-proton (approved):
Looks good 👍

Issue #3372, opened 2025-10-17
LCFS - Standardize Download/Upload Excel Template Buttons


airinggov:
**Description:**  
Update the **Download Excel template** and **Upload Excel template** buttons on the **Add Charging Site** and **Add FSE** pages to use a consistent text and colour.  
Both buttons should appear as **blue outline buttons** (not solid blue), matching the style used on the **Charging Site** page. These buttons should **not include any dropdown options**.

**Wireframe:**  
<img width="545" height="168" alt="Image" src="https://github.com/user-attachments/assets/3d58a9f2-b201-4f65-9402-04a8542be78d" />
<img width="539" height="162" alt="Image" src="https://github.com/user-attachments/assets/7a90604d-4f66-4ed4-9282-c54833057431" />

**Purpose and benefit to user:**  
Ensures visual consistency and usability across related pages, making it easier for users to identify and interact with Excel upload/download actions in a predictable manner.

**Acceptance Criteria:**  
- [ ] Both **Download Excel template** and **Upload Excel template** buttons are blue outline style.  
- [ ] Button styling matches the format used on the **Charging Site** page.  
- [ ] Buttons appear without dropdowns or extra options.  
- [ ] Functionality of both buttons remains unchanged.  
- [ ] Consistent layout and spacing are applied across all affected pages.  

**Development Checklist:**  
- [ ] Update button components on **Add Charging Site** and **Add FSE** pages.  
- [ ] Remove dropdown configurations if present.  
- [ ] Verify that actions (download/upload) work correctly post-update.  
- [ ] Check UI for alignment and consistent spacing.  
- [ ] Perform regression testing to confirm no visual or functional impact on other button types.  

**Notes:**  
- Use the existing button component styles from the **Charging Site** page as reference.  
- Confirm with design team if hover and focus states match the standard blue outline button style.

PR #3390, opened 2025-10-22
Fix: LCFS - Excel templates for Charging Sites and FSE - #3355

### LINKED ISSUE #3355: LCFS - Excel templates for Charging Sites and FSE
**User Story:**
As a BCeID user, I want to download empty Excel templates for creating new Charging Sites and FSE, so that I can add new data without overwriting or editing any existing records.

**Description:**
The Excel templates for both Charging Sites and FSE should always be empty when downloaded. These templates are intended **only for creating new records**, not for editing or overwriting existing data.

Image

Image

Each template should include **select lists** and **multi-select lists**  that match the options available in the corresponding web form to ensure data consistency and prevent invalid entries.

**Acceptance Criteria:**

* [ ] Downloaded Charging Site and FSE Excel templates contain no pre-existing data.
* [ ] Templates include select lists that match the web form options.
* [ ] Upload validation prevents overwriting or modifying existing records.

**Notes:**

* Confirm that select list values dynamically reflect current web form options.
* Consider including instructions or a note within the template header about “new entries only.”

prv-proton:
Fix #3355, #3372

### DIFF
--- a/frontend/src/views/ComplianceReports/EditViewComplianceReport.jsx
+++ b/frontend/src/views/ComplianceReports/EditViewComplianceReport.jsx
@@ -476,13 +476,14 @@ export const EditViewComplianceReport = ({ isError, error }) => {
     }
 
     const shouldShowAssessmentStatement =
-      isGovernmentUser && !qReport?.isQuarterly && !hasDraftSupplemental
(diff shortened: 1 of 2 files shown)


AlexZorkin (approved):
Let's merge this before my latest changes. Looks good, I will handle the conflicts. 👍

Issue #3379, opened 2025-10-21
LCFS - Keep 'Intended users' on FSE form and remove from Charging sites


airinggov:
**Describe the Story**  
Retain the **‘Intended users’** field on the **FSE input form** and **remove it** from the **Charging site form** and associated **data model**. 

The field was previously relocated to the Charging site form, but should remain at the FSE level for data accuracy and reporting alignment.

**Purpose and Benefit to User**  
Ensures that user intent data is captured accurately at the FSE level, where it varies by equipment, rather than at the site level.  
Simplifies data maintenance and eliminates redundancy between FSE and site records.

**Acceptance Criteria**  
- [ ] ‘Intended users’ field remains visible and functional on the FSE input form  
- [ ] Field is removed from the Charging site input form UI  
- [ ] Field and related data attributes are removed from the Charging site data model  
- [ ] Excel templates are modified to cater for it.
- [ ] Migration scripts are modified as required.
- [ ] Validation and saving logic reflect field presence only at the FSE level  
- [ ] Existing FSE ‘Intended users’ data is preserved during the change  

**Additional Context**  
- Current duplication between site-level and FSE-level fields can cause confusion and inconsistent reporting.  
- Any related database references or API endpoints for the site-level ‘Intended users’ field should be updated or deprecated.

Issue #3373, opened 2025-10-17
LCFS - Filter and relocate Allocating organization field


airinggov:
**Describe the Story**  
Update how and where the **Allocating organization** field is displayed and populated.  
The field should be **filtered** to show only organizations associated with allocation agreements for the **logged-in supplier** from any **current draft report** or **previously assessed reports**.  
If the supplier has no allocation agreements, the field should be **disabled** (as it is optional).  
Additionally, **relocate** this field from the **FSE input form** to the **Charging site input form**.

**Purpose and Benefit to User**  
Ensures that users can only select valid, relevant allocation organizations tied to their existing agreements.  
Improves data accuracy, simplifies FSE input, and aligns allocation management at the site level.

**Acceptance Criteria**  
- [ ] Allocating organization select list filters results based on allocation agreements for the logged-in supplier  
- [ ] Data source includes both current draft and previously assessed reports  
- [ ] Field is disabled when no allocation agreements exist for the supplier  
- [ ] Field is moved from the FSE input form to the Charging site input form and data model
- [ ] Validation and saving logic updated to reflect new field location and rules  
- [ ] Existing FSE records retain allocation data correctly after migration  
- [ ] Excel templates are modified to cater for it.
- [ ] Migration scripts are modified as required.
- [ ] Add text to the end of existing instructional text: Allocating organizations tied to your allocation agreements, if an organization isn't listed you must first enter an allocation agreement in your compliance report.

**Additional Context**  
- Field currently resides on the FSE form but will be tied more logically to site-level configuration.  
- May require backend API adjustment to retrieve allocation agreements for the logged-in supplier.  
- UI should visually indicate when the field is disabled due to no available allocations.

PR #3397, opened 2025-10-24
Feat: fse updates to sites and equipment with intended users

AlexZorkin:
- Tickets #3373, #3379


prv-proton (approved):
Looks good 👍 . Thanks for the updates, lots of things are taken care of.
Please do consider to add "update_date"  in the charging_site/repo.py

PR #3398, opened 2025-10-27
LCFS - Bug - “Fuel Produced in Canada” Select List Should Only Be Active for Default Carbon Intensity

### LINKED ISSUE #3394: LCFS - Bug - “Fuel Produced in Canada” Select List Should Only Be Active for Default Carbon Intensity
**Describe the Bug**  
The **“Fuel produced in Canada”** select list is currently active for all fuel types, including those with defined fuel codes. This is incorrect behavior, as the location can already be determined from the selected fuel code.

**Expected Behaviour**  
The **“Fuel produced in Canada”** select list should only be **active** when the user selects **“Default carbon intensity”** as the fuel type.  
For any other **fuel code**, the field should be **locked and greyed out**, automatically set based on the fuel code’s known production location.

**Actual Behaviour**  
The select list remains editable for all fuel code selections, allowing users to change a value that should be system-determined.

**Note**: the top line shown for 'Prescribed carbon intensity' for fossil derived/non renewable fuels is not part of this requirement. It is correctly shown not having any active fields or Yes/No displayed.

Image

**Implications**  
- Users can unintentionally override system-defined data.  
- Creates potential inconsistencies in reporting and carbon intensity calculations.  
- Leads to incorrect country associations for specific fuel codes.

**Steps To Reproduce**  
**User/Role:** IDIR or BCeID  
1. Navigate to the **Fuel Code input form**.  
2. Select any valid **Fuel code** (not “Default carbon intensity”).  
3. Observe that the **“Fuel produced in Canada”** field remains editable.  
4. Change the selection — no system restriction is applied.  

**Expected Result:**  
The field should be **locked and greyed out** for all fuel codes except **“Default carbon intensity.”**  

**Additional Context**  
- The system can determine production location from fuel code metadata.  
- Ensure the UI reflects this by disabling the field appropriately.

kevin-hashimoto:
- fix: produced in canada field state fix based on determining carbon intensity, closes #3394


AlexZorkin (approved):
👍

PR #3401, opened 2025-10-28
Fix: LCFS - Incorrect FSE, Document, and Comment Copy Behaviour When Creating Supplemental or Analyst Adjustment Reports #3392

### LINKED ISSUE #3392: LCFS - Incorrect FSE, Document, and Comment Copy Behaviour When Creating Supplemental or Analyst Adjustment Reports
**Describe the Bug**

When creating a Supplemental Report or an Analyst Adjustment Report, the system currently copies the Final Supply Equipment (FSE) data, documents, and internal comments associations from the existing report to the newly created one.
However, this can lead to unintended data deletion if the new report (still in draft state) is deleted — as the copied associations remain linked, leading to cascade deletions of shared data.

Additionally, with the new Charging Equipment system, the FSE data for supplemental reports must be pulled specifically from the previous original report, not the latest version, but the current implementation does not enforce this.

**Expected Behaviour**

- FSE, documents, and internal comments should not be directly copied as linked associations when creating supplemental or analyst adjustment reports.
- Instead, data should either be cloned (disassociated from original objects) or fetched anew from the original report only.
- Deletion of a draft supplemental report should never affect existing data tied to the original report.

**Actual Behaviour**

- The copy_to_report, copy_documents, and copy_internal_comments methods are invoked during supplemental or analyst adjustment report creation, creating linked associations with the original entities.
- If the new draft report is deleted, associated FSE, documents, or internal comments may also be deleted unintentionally due to shared linkage.
- FSE data is being sourced incorrectly from the latest report rather than strictly from the original report in the report lineage.

**Implications**

- Data loss risk: Deleting a draft supplemental report can cascade-delete shared entities (FSE, documents, internal comments).
- Data integrity issues: The Charging Equipment system may show incorrect or outdated data in newly created reports. 
- Audit and traceability issues: Historical consistency between reports and original submissions could be compromised.

**Steps To Reproduce**

1. User/Role: Analyst or Government User
2. Go to Compliance Reports.
3. Select an existing report and create a Supplemental Report or Analyst Adjustment Report.
4. Observe that FSE, documents, and internal comments are copied over.
5. Delete the newly created draft supplemental report.
6. Return to the original report and check FSE or documents — note that data may be deleted or missing.

prv-proton:
Fix #3392.

1. Instead of fse_compliance_reporting table make use of existing compliance_report_charging_equipment table.
2. use group_uuid or chained Ids to retrieve charging equipments, document and internal comments.
3. remove cloning of data for every new supplemental report id

### DIFF
--- a/backend/lcfs/db/migrations/versions/2025-10-28-06-48_1f3ce398db1c.py
+++ b/backend/lcfs/db/migrations/versions/2025-10-28-06-48_1f3ce398db1c.py
@@ -75,16 +75,6 @@ def upgrade() -> None:
     )
     op.drop_column("compliance_report_charging_equipment", "date_of_supply_from")
     op.drop_column("compliance_report_charging_equipment", "date_of_supply_to")
-    # Fix supply dates for wrongly entered data where from_date > to_date
-    op.execute(
-        """
-        UPDATE final_supply_equipment fse
-        SET 
-            supply_from_date = make_date(2024, EXTRACT(MONTH FROM fse.supply_from_date)::int, EXTRACT(DAY FROM fse.supply_from_date)::int),
-            supply_to_date   = make_date(2024, EXTRACT(MONTH FROM fse.supply_to_date)::int, EXTRACT(DAY FROM fse.supply_to_date)::int)
-        WHERE fse.supply_from_date > fse.supply_to_date;
-        """
-    )
     # Migrate data from final_supply_equipment to compliance_report_charging_equipment
     op.execute(
         """


backend/lcfs/db/models/compliance/ComplianceReportChargingEquipment.py
```diff
@@ -100,13 +127,13 @@ def __repr__(self):
...
             f")>"
         )
 
     @property
     def supply_period_days(self):
         """Calculate the number of days in the supply period."""
-        if self.date_of_supply_from and self.date_of_supply_to:
-            return (self.date_of_supply_to - self.date_of_supply_from).days + 1
+        if self.supply_from_date and self.supply_to_date:
+            return (self.supply_to_date - self.date_of_supply_from).days + 1
```
AlexZorkin: self.supply_from_date typo? should be supply_from_date?

backend/lcfs/web/api/final_supply_equipment/repo.py
```diff
@@ -749,7 +752,13 @@ async def get_fse_reporting_list_paginated(
         # Apply pagination
         offset = (pagination.page - 1) * pagination.size
         paginated_query = final_query.offset(offset).limit(pagination.size)
-
+        print("Generated SQL Query:")
```
AlexZorkin: remove print statements

AlexZorkin (changes requested):
If FSE data is now shared across report versions using group_uuid, should this constraint be on (group_uuid, charging_equipment_id, organization_id) instead of (compliance_report_id, ...)?

prv-proton:
> If FSE data is now shared across report versions using group_uuid, should this constraint be on (group_uuid, charging_equipment_id, organization_id) instead of (compliance_report_id, ...)?

Thanks for the review, fixed them.

AlexZorkin (approved):
Looks good 👍

PR #3402, opened 2025-10-28
Feat: summary lines 7 and 9 mutual exclusive - 3364

AlexZorkin:
- updated business logic to only allow one of 7 of 9 to have a value at a time


prv-proton (approved):
Looks good 👍

PR #3452, opened 2025-11-03
Fix: Correct organization status filter options - 3427

### LINKED ISSUE #3427: LCFS - Bug: Invalid Status Filter Options on Organizations Page (IDIR)
**Describe the Bug**
On the Organizations page (IDIR view), the Status column in the view table includes filter options for "Canceled" and "Suspended", which are not valid statuses for organizations. The only valid statuses should be "Registered" and "Unregistered".

**Expected Behaviour**
Only "Registered" and "Unregistered" should appear as filter options.

**Actual Behaviour**
"Canceled" and "Suspended" are incorrectly included as filter options.

Image

**Implications**
Misleading filter options may confuse users.

**Steps To Reproduce**
User/Role:

1. Log in as an IDIR user.
2. Navigate to the Organizations page.
3. Open the filter options for the Status column in the table.
4. Observe the available filter values.

hamed-valiollahi:
This PR ensures only Registered and Unregistered statuses appear on the Organizations list for IDIR users.

Closes #3427

### DIFF
--- a/backend/lcfs/tests/compliance_report/test_summary_repo.py
+++ b/backend/lcfs/tests/compliance_report/test_summary_repo.py
@@ -707,3 +707,253 @@ async def test_penalty_override_user_relationship(
     assert result.penalty_override_user == users[0].user_profile_id
     # Test the relationship is properly configured
     assert hasattr(result, "penalty_override_user_profile")
+
+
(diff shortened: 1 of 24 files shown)


AlexZorkin (approved):
👍

PR #3456, opened 2025-11-04
Fix: FSE and Summary updated - 3425 3440

AlexZorkin:
- Diesel placeholder updated
- Charging site tab color updated
- FSE draft select logic updated


prv-proton (approved):
Looks good 👍

PR #3486, opened 2025-11-07
Fix: LCFS - Bug: Blank Save & Return Modal After Excel Upload (BCeID User)

### LINKED ISSUE #3405: LCFS - Bug: Blank Save & Return Modal After Excel Upload (BCeID User)
**Describe the Bug**
When a BCeID user uploads an Excel file to the Add Charging Sites section, they are presented with a Save & Return button. Clicking this button triggers a modal window that appears blank and non-functional. Although the Excel upload auto-saves successfully, the modal creates confusion as it implies further action is needed but provides no content or feedback.

**Expected Behaviour**
The modal should either:

- Display a confirmation message indicating the upload has been saved and the user can safely return, or
- Redirect the user without showing a modal, if no further interaction is needed.

**Actual Behaviour**
A blank modal window appears with no content or actionable elements. The upload is saved, but the user is not informed, leading to confusion.

Image

**Implications**
User confusion due to lack of feedback.

**Steps To Reproduce**
User/Role:

1. Log in as a BCeID user.
2. Navigate to Add Charging Sites.
3. Upload a valid Excel file.
4. Click the Save & Return button.
5. Observe the modal window that appears.

prv-proton:
Fixes #3405
- Allow delete on charging site on if it's  in Draft status.

### DIFF
--- a/backend/lcfs/tests/compliance_report/test_compliance_report_repo.py
+++ b/backend/lcfs/tests/compliance_report/test_compliance_report_repo.py
@@ -12,6 +12,7 @@ from lcfs.db.models.compliance.ComplianceReport import ReportingFrequency
 from lcfs.db.models.user import UserProfile
 from lcfs.web.api.base import (
     PaginationRequestSchema,
+    SortOrder,
 )
 from lcfs.web.api.compliance_report.schema import ComplianceReportBaseSchema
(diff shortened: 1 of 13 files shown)


AlexZorkin (approved):
👍

PR #3499, opened 2025-11-14
Fix: Charging Site Upload Append Mode & Template Rules - 3481

### LINKED ISSUE #3481: LCFS - BUG - Charging site upload utility
**Describe the Bug**  
The Charging Site upload utility throws an error and fails to process the uploaded Excel file.  
The upload modal appears to be in **Overwrite mode** when it should be in **Append mode**.  
This feature should only allow adding new sites, not overwriting existing ones.

**Expected Behaviour**  
- The upload utility should operate in **Append mode**, allowing users to add new charging sites.  
- If a duplicate **Site name** (within the same organization) is found, the upload should return a clear validation error message instructing the user to remove duplicates from the file.  
- The upload should succeed for valid new records only.  

**Actual Behaviour**  
- Uploads fail with an unexpected error message (see attached).  
- The system behaves as if in Overwrite mode, attempting to replace existing entries instead of adding new ones.  

Image

**Implications**  
- Users cannot upload new charging sites.  
- Risk of overwriting existing site data if not corrected.  

**Steps To Reproduce**  
User/Role: BCeID Supplier  
1. Navigate to **Charging Sites** page.  
2. Click **Upload Excel**.  
3. Upload a valid template file with new charging site data.  
4. Observe the modal behavior and resulting error message.  

**Additional Issues Identified**  
- The **Excel template** should remove the following system-managed columns:  
  - Organization  
  - Site Code  
  - Status  
- The **Allocating Organization** field in the Excel upload should only allow organizations that exist in the supplier’s **Allocation Agreements**.  
  - A select list should be implemented to enforce this constraint.

hamed-valiollahi:
This PR includes the following, 
- Fix append mode behaviour
- Add duplicate site validation
- Clean up template columns
- Restrict allocating org list

Closes #3481

### DIFF
--- a/frontend/src/views/AllocationAgreements/AddEditAllocationAgreements.jsx
+++ b/frontend/src/views/AllocationAgreements/AddEditAllocationAgreements.jsx
@@ -509,7 +509,12 @@ export const AddEditAllocationAgreements = () => {
             ))}
           </BCBox>
         </div>
-        {isFeatureEnabled(FEATURE_FLAGS.ALLOCATION_AGREEMENT_IMPORT_EXPORT) &&
+        {/* 
+          Temporarily hidden
+          Excel import/export causing validation issues
+          TODO: Re-enable after validation logic is fixed
+        */}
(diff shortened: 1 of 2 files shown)


prv-proton (approved):
Looks good 👍

PR #3518, opened 2025-11-18
Fix: Quarterly report Q1 null - 3510

AlexZorkin:
- updated Q1 default in compliance unit calculation
- invalidated allocation agreement cache on update so selection list is refreshed


prv-proton (approved):
Looks good 👍

PR #3520, opened 2025-11-19
Feat: LCFS - FSE utilization assessment

### LINKED ISSUE #3294: LCFS - FSE utilization assessment
**Description:**
Add a new IDIR-only view column to the FSE forms that calculates electricity reasonableness as a **percentage** using the following formula:

Capacity utilization (%) = (Total energy reported (kWh)/Reasonable max value) x100
- Reasonable max value is = Charger power output × number of FSE (if line by line, this will always be 1)× operational hours × operational days (Can be pulled from the dates of supply in the FSE table)

This should only display if the user has input an kWh volume amount in the FSE form. This column is not a required field, therefore may not always be visible to the analyst if there was no volume input by the BCeID user.

**Wireframe:** (https://preview.uxpin.com/e0e2f15fc1b52987fd0287ad8c8103452dd16d60#/pages/215889141/simulate/no-panels)

To provide IDIR users with a quick reference metric that helps assess the reasonableness of electricity usage data submitted in FSE forms. This supports validation and review processes. Automating will reduce staff time and speed up processing significantly. 

**Acceptance Criteria:**
For FSE form the calculations use the following power output per application (if kWh usage field completed):

Image

**Development Checklist:**
- [ ]  Column is visible only to IDIR users, for a submitted compliance report
- [ ]  Column calculates and displays the electricity reasonableness percentage using the specified formula
- [ ]  Column is added to all relevant FSE form views without disrupting existing layout or functionality

**Notes:**
- See Card attachment in Teams folder named "3294 reporting scenarios"
- This column is for internal review purposes and should not be visible to BCeID users.
- Consider formatting the percentage to two decimal places for clarity.
- May require backend support to access all variables used in the calculation.

prv-proton:
Closes #3294

### DIFF
--- a/frontend/package-lock.json
+++ b/frontend/package-lock.json
@@ -53,7 +53,7 @@
         "react-i18next": "^15.4.1",
         "react-input-mask": "^2.0.4",
         "react-leaflet": "^4.2.1",
-        "react-leaflet-custom-control": "^1.4.0",
+        "react-leaflet-custom-control": "^1.3.2",
         "react-number-format": "^5.4.3",
         "react-quill": "^2.0.0",
         "react-router-dom": "^7.5.2",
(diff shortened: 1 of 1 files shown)


AlexZorkin (approved):
Very interesting feature! I wonder how accurate the numbers will be in a real world setting. Did you test any real values against this? 👍

PR #3522, opened 2025-11-20
LCFS - Schedule save

### LINKED ISSUE #3515: LCFS - Schedule save
see card #3424 

**Describe the task**  
Ensure all compliance report schedules save correctly when a user clicks **"Save and return to report"** before the final auto-save verification is triggered. Currently, auto-save only executes when a cell loses focus. Clicking the button should force a final save by triggering a focus reset or equivalent save event.

**Purpose**  
To prevent data loss and ensure user-entered schedule values are consistently saved, even when users click the save button without first clicking outside the edited cell.

**Acceptance Criteria**  
- [ ] Clicking **"Save and return to report"** triggers a forced save event for all active schedules.  
- [ ] Any unsaved cell edits are captured even if the cell has not lost focus.  
- [ ] No duplicate saves or unexpected validation errors occur.  
- [ ] Behaviour is consistent across all report schedules and report years.  
- [ ] Verification testing confirms no data loss scenarios remain.

**Additional context**  
- Current behaviour relies on blur/focus-out events to fire auto-save.  
- Users may click save immediately after typing, resulting in missing entries.  
- A synthetic blur, commit, or save-trigger mechanism may be needed.

kevin-hashimoto:
- feat: save and return button now triggers a stopediting method which should reliably save the unsaved row before navigating away, closes #3515

### DIFF
--- a/backend/lcfs/web/api/charging_equipment/schema.py
+++ b/backend/lcfs/web/api/charging_equipment/schema.py
@@ -67,6 +67,8 @@ class ChargingEquipmentCreateSchema(BaseSchema):
     model: Optional[str] = Field(None, max_length=100)
     level_of_equipment_id: int
     ports: Optional[PortsEnum] = None
+    latitude: Optional[float] = None
+    longitude: Optional[float] = None
     notes: Optional[str] = None
     intended_use_ids: Optional[List[int]] = []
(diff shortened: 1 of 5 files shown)


hamed-valiollahi (approved):
Looks good to me! 👍

PR #3523, opened 2025-11-20
Fix: LCFS - Floating Page Navigation Bar Does Not Scroll Left Properly

### LINKED ISSUE #3459: LCFS - Floating Page Navigation Bar Does Not Scroll Left Properly
**Describe the Bug**
When viewing a large list and the floating page navigation bar appears, attempting to scroll left does not work if the user is already partially scrolled to the left. The navigation bar seems to treat the current position as the furthest left, preventing further leftward scrolling.

**Expected Behaviour**
The navigation bar should allow users to scroll left regardless of their current position on the page, especially if they are starting from the far right.

Image

**Actual Behaviour**
The navigation bar locks the leftward scroll based on the user's current position, making it impossible to scroll further left even when more content exists.

**Implications**
Users may be unable to access earlier pages in the list, leading to frustration and reduced usability, especially when navigating large datasets or paginated content.

**Steps To Reproduce**
User/Role: Any

1. Go to a page with a large list that triggers the floating page navigation bar
2. Scroll horizontally to the right
3. Attempt to use the floating navigation bar to scroll left
4. Observe that it does not scroll left beyond the current position

prv-proton:
Fixes #3459

### DIFF
--- a/backend/lcfs/web/api/charging_equipment/schema.py
+++ b/backend/lcfs/web/api/charging_equipment/schema.py
@@ -67,6 +67,8 @@ class ChargingEquipmentCreateSchema(BaseSchema):
     model: Optional[str] = Field(None, max_length=100)
     level_of_equipment_id: int
     ports: Optional[PortsEnum] = None
+    latitude: Optional[float] = None
+    longitude: Optional[float] = None
     notes: Optional[str] = None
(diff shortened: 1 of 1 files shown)


AlexZorkin (approved):
tricky fix 👍

PR #3575, opened 2025-12-04
Convert layout files to ts

### LINKED ISSUE #2808: convert layout files to ts
parent task for layout files

kevin-hashimoto:
- feat: layout files now converted to ts, closes #2808 
- note: disabled ts check in github actions as there will be up/down stream ts errors as we convert other files to ts. this should be turned back on when the majority of the ts conversion work is complete.


AlexZorkin (approved):
Nice progress 👍

PR #3588, opened 2025-12-06
Feat: Summary Fixes and Cleanup

AlexZorkin:
- Updated summary logic for all lines to better adhere to legislative requirements
- Added multiple tests around compliance report edge cases
- Updated staging seeders with edge case reports

### DIFF
--- a/backend/lcfs/tests/compliance_report/test_summary_repo.py
+++ b/backend/lcfs/tests/compliance_report/test_summary_repo.py
@@ -241,6 +241,7 @@ async def transfers(dbsession, organizations):
         Transfer(
             transfer_id=992,
             agreement_date=date,
+            transaction_effective_date=date,
             from_organization_id=organizations[0].organization_id,
             current_status_id=6,
             quantity=1,
@@ -248,6 +249,7 @@ async def transfers(dbsession, organizations):
         Transfer(
             transfer_id=993,
             agreement_date=date,
+            transaction_effective_date=date,
             from_organization_id=organizations[0].organization_id,
             current_status_id=6,
             quantity=1,
@@ -255,6 +257,7 @@ async def transfers(dbsession, organizations):
         Transfer(
             transfer_id=994,
             agreement_date=date,
+            transaction_effective_date=date,
             from_organization_id=organizations[1].organization_id,
             current_status_id=6,
             quantity=1,
@@ -262,6 +265,7 @@ async def transfers(dbsession, organizations):
         Transfer(
             transfer_id=995,
             agreement_date=date,
+            transaction_effective_date=date,
             from_organization_id=organizations[1].organization_id,
             current_status_id=6,
             quantity=1,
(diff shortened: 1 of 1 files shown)


backend/lcfs/db/migrations/versions/2025-12-06-12-00_c4d5e6f7a8b9.py
```diff
@@ -0,0 +1,33 @@
...
+# revision identifiers, used by Alembic.
+revision = "c4d5e6f7a8b9"
+down_revision = "b71c1d2e3f45"
+branch_labels = None
+depends_on = None
+
+
+def upgrade() -> None:
+    # Drop legacy TFRS offset columns; they are no longer used by views or models.
+    op.drop_column("compliance_report_summary", "credits_offset_a")
```
prv-proton: These views are dependent on compliance_report_summary, though the columns are not being used in it directly, drop column is causing error while running migration. Probably adding cascade should fix this.
```
sqlalchemy.exc.DBAPIError: (sqlalchemy.dialects.postgresql.asyncpg.Error) <class 'asyncpg.exceptions.DependentObjectsStillExistError'>: cannot drop column credits_offset_a of table compliance_report_summary because other objects depend on it
DETAIL:  view vw_compliance_report_base depends on column credits_offset_a of table compliance_report_summary
view vw_compliance_report_fuel_volume_history depends on column credits_offset_a of table compliance_report_summary
HINT:  Use DROP ... CASCADE to drop the dependent objects too.
```
vw_compliance_report_base
vw_compliance_report_fuel_volume_history

backend/lcfs/web/api/compliance_report/summary_service.py
```diff
@@ -671,6 +692,32 @@ async def calculate_compliance_report_summary(
...
+                        prev_compliance_report.summary.line_8_obligation_deferred_gasoline or 0
+                    )
+                if summary_model.line_8_obligation_deferred_diesel in [None, 0]:
+                    summary_model.line_8_obligation_deferred_diesel = (
+                        prev_compliance_report.summary.line_8_obligation_deferred_diesel or 0
+                    )
+                if summary_model.line_8_obligation_deferred_jet_fuel in [None, 0]:
+                    summary_model.line_8_obligation_deferred_jet_fuel = (
+                        prev_compliance_report.summary.line_8_obligation_deferred_jet_fuel or 0
+                    )
```
prv-proton: image
Just my view, when we're not enabling the edit option for these cells then should we be showing these values in summary for both 6 & 8?
LCFS1 2025 report in dev

prv-proton (changes requested):
Thanks for the cleanup, overall changes look good, just some minor fixes may need to be made.

backend/lcfs/web/api/compliance_report/summary_service.py
AlexZorkin: we will address this after we merge this branch. Lots of other fixes to get into develop

prv-proton (approved):
👍

PR #3589, opened 2025-12-08
Fix: LCFS - Refine FSE & Charging Site Production Release Data and missing data refill

### LINKED ISSUE #3584: LCFS - Refine FSE & Charging Site Production Release Data
**Describe the Bug**
There are three data issues observed in Prod affecting the Charging Site production release and the FSE index:

1. Charging Site – Allocating Organization Column
When a reporting organization does not have an allocating organization associated with it, the system is copying the reporting organization name into the “Allocating organization” column. These columns should only be populated when the organization has reported an allocated supply (which is currently handled correctly). Organizations without allocated supply should show blank in this column.

2. FSE Index – Registration # Format
The Registration # displayed on the FSE index in Prod is showing 5 digits in the second half of the identifier where it should show only 3 digits (with leading zeros if applicable).

3. FSE Index – Version # Start Value
The Version # on the FSE index is showing 0 for initial entries. The first version should be 1 in all cases, increasing sequentially (1, 2, 3, …).

**Expected Behaviour**
1. Charging Site – Allocating organization
Only populate “Allocating organization” when the reporting organization has reported an allocated supply.
If no allocated supply, the “Allocating organization” column should be blank.

2. FSE Index – Registration #
Display the Registration # with a 3-digit suffix (e.g., ABC123-001), using leading zeros if needed.

3. FSE Index – Version #
Initial version should be 1. Subsequent versions should increment by +1 per update.

**Actual Behaviour**
1. Charging Site – Allocating organization
For organizations without allocated supply, the reporting organization name is erroneously copied into “Allocating organization”.

2. FSE Index – Registration #
Registration # shows a 5-digit suffix (e.g., ...-12345) instead of the required 3-digit suffix.

3. FSE Index – Version #
Initial versions display as 0 instead of 1.

**Implications**
Incorrect data displays and differences between Prod and Test

**Steps To Reproduce**
User/Role: IDIR
1. Go to Charging site and FSE index
2. See errors in table

prv-proton:
Fixes #3584
- Missing data refilling.


prv-proton:
> One issue to resolve: There’s a mismatch between the migration and the model around equipment_number. The migration enforces a 3-digit numeric format with a check constraint (^[0-9]{3}$), but ChargingEquipment.generate_equipment_number now uses next_base36, which will eventually produce non-numeric values and also treats the result inconsistently (string vs int). We should either stick with simple 3-digit integers in the model or relax the DB constraint and fully commit to base36, but not mix the two.

@AlexZorkin , fixed to have the sequence number in int form, please review.

PR #3591, opened 2025-12-08
Fix: Rounding solution

AlexZorkin:
- updated rounding to be more consistent throughout backend calculation of compliance units


prv-proton (approved):
Looks good 👍

PR #3595, opened 2025-12-09
LCFS - HDRD Default Selection Locks “Supplied in Q1” Field Incorrectly

### LINKED ISSUE #3565: LCFS - HDRD Default Selection Locks “Supplied in Q1” Field Incorrectly
**Describe the Bug**
When HDRD fuel type is selected as Canadian produced, the system also asks if it was supplied in Q1. However, if the user initially selects “Supplied in Q1 = Yes” and then changes “Fuel Produced in Canada” from “No” to “Yes,” the “Supplied in Q1” field remains locked as “Yes” instead of being cleared. This behavior is incorrect because the supplied-in-Q1 status should not persist when the Canadian production status changes.

**Expected Behaviour**
- If “Fuel Produced in Canada” is changed from “No” to “Yes,” the “Supplied in Q1” field should reset (blank) rather than remain locked as “Yes.”
- The system should allow users to reselect or leave the field blank after the change.

**Actual Behaviour**
- Changing “Fuel Produced in Canada” from “No” to “Yes” does not clear the “Supplied in Q1” selection.
- The field remains locked as “Yes,” creating unnecessary data.

**Implications**
A field is left with "yes" when, as a Canadian produced fuel, it doesn't matter when it was supplied for 2025 reporting. 

**Steps To Reproduce**
User/Role: BCeID

1. Select HDRD as fuel type in fuel supply schedule.
2. Set “Fuel Produced in Canada” to “No.”
3. Select “Supplied in Q1 = Yes.”
4. Change “Fuel Produced in Canada” to “Yes.”
5. Observe that “Supplied in Q1” remains locked as “Yes” instead of clearing.

kevin-hashimoto:
- feat: Q1 column now clears when canada supplied is set to 'yes', closes #3565


AlexZorkin (approved):
👍

PR #3603, opened 2025-12-11
Fix: LCFS – In-Line FSE Date Change in Compliance Report Not Saving

### LINKED ISSUE #3539: LCFS – In-Line FSE Date Change in Compliance Report Not Saving
**Describe the Bug**
As a BCeID user in a compliance report, changing the FSE section dates of supply inline in compliance report does not save. Default date remains.

**Expected Behaviour**
You should be able to change the dates of supply in line with the FSE in case the supply was not for the entire year

**Actual Behaviour**
You can edit and change the dates of supply in the FSE section of the compliance report, but it does not save and persist in the main report table after. 

**Implications**
This leads to confusion for users and potentially incorrect dates of supply for FSE listed in the compliance report.

**Steps To Reproduce**
User/Role: BCeID user
1. Go to compliance report with an FSE reported
2. Click on the dates of supply fields and pick a date
3. Save changes and return to main report
4. See error where dates were note saved

prv-proton:
Closes #3539
- Union query modified to prioritise the equipment that matches to the current compliance report group uuid.
- if a row throws error to add to current compliance period in the backend then ensure at the UI the row is not selected for edit.
- add compliance report group uuid to the unique constraint check on compliance_report_charging_equipment table.

### DIFF
--- a/backend/lcfs/db/migrations/versions/2025-12-11-10-35_64ab3b361bde.py
+++ b/backend/lcfs/db/migrations/versions/2025-12-11-10-35_64ab3b361bde.py
@@ -1,7 +1,7 @@
 """Modify constraint on compliance report charging equipment
 
 Revision ID: 64ab3b361bde
-Revises: b8c9d0e1f2a3
+Revises: bbfbe5d4c7e8
 Create Date: 2025-12-11 10:35:57.365155
 
 """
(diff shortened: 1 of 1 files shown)


AlexZorkin (approved):
tricky fix, nice one 👍

PR #3605, opened 2025-12-11
Feat: BCeID permission fix for compliance reports - 3596

### LINKED ISSUE #3596: LCFS – BCeID Role Permission Issue for Compliance Report Access
**Describe the Bug**
Currently, users must have both the Compliance Reporting and Signing Authority roles to view a compliance report. If a user has only one of these roles, they receive a message stating they do not have permission. This is incorrect because role permissions should allow partial functionality based on the assigned role.

**Expected Behaviour**
Users with Compliance Reporting role should be able to:
Create, edit, and view compliance reports.

Users with Signing Authority role should be able to:
View, sign, and submit compliance reports.

Users with both roles should be able to:
Create, edit, view, sign, and submit compliance reports.

**Actual Behaviour**
Users with only one role cannot view compliance reports and receive a “no permission” message.
Full access is only granted when both roles are assigned.

**Implications**
Prevents users from performing tasks aligned with their role.
Causes workflow delays and confusion for organizations managing compliance reports.

**Steps To Reproduce**
User/Role:BCeID

1. Assign only the Compliance Reporting role to a user.
2. Attempt to view a compliance report.
3. Observe permission error message.
4. Repeat with only Signing Authority role and confirm same issue.

hamed-valiollahi:
This PR corrects role-based access so each BCeID role gets its proper report permissions.

Closes #3596

### DIFF
--- a/backend/lcfs/web/api/organization/views.py
+++ b/backend/lcfs/web/api/organization/views.py
@@ -240,7 +240,7 @@ async def update_transfer(
     response_model=ComplianceReportBaseSchema,
     status_code=status.HTTP_201_CREATED,
 )
-@view_handler([RoleEnum.COMPLIANCE_REPORTING, RoleEnum.SIGNING_AUTHORITY])
+@view_handler([RoleEnum.COMPLIANCE_REPORTING])
 async def create_compliance_report(
     request: Request,
     organization_id: int,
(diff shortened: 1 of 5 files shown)


hamed-valiollahi:
> looks good, its been requested to make this a hot fix to prod so please also create a pr with these changes directly to main branch.

Of course, thanks for pointing it out. I’ve created the PR: https://github.com/bcgov/lcfs/pull/3606 Thanks for the review.

PR #3623, opened 2025-12-17
Fix: BUG - LCFS role switcher formatting

### LINKED ISSUE #3569: BUG - LCFS role switcher formatting
**Describe the Bug**
When the role switcher is activated, the button formatting shifts and does not align with its default state before selection.

Image

**Expected Behaviour**
The role switcher button should maintain consistent alignment and formatting whether it is in its default state or after being selected.

**Actual Behaviour**
After activation, the button appears misaligned compared to its original position and style.

**Implications**
* Creates a visual inconsistency that may confuse users.
* Reduces overall UI polish and perceived quality of the interface.

prv-proton:
Fixes #3569


AlexZorkin (approved):
👍

PR #3651, opened 2025-12-19
Feat: C-Proxy Prefix, Summary Text update, FSE Save 

AlexZorkin:
- Prevent PROXY prefix from being auto-changed to C-BCLCF when country is set to Canada; auto-change now only applies to BCLCF prefix, and adds C-PROXY as a new prefix option.
- Fix Line 5 summary description to say "(transferred away)" instead of "(received)"
- Fix FSE single edit save by correcting isRowValid function to properly access properties from the row parameter instead of referencing undefined variables.


prv-proton (approved):
👍

PR #3660, opened 2025-12-23
Feat: Renewable Table Visibility - 3610

AlexZorkin:
- if lines 3 and 9 are zero or null in all fuel categories, then we hide the renewable table and display message in place
- Tests added for new logic


hamed-valiollahi (approved):
Looks great! 👍

PR #3667, opened 2025-12-24
Feat: Map, FSE and Director workflow fixes

hamed-valiollahi:
Here is the list of issues resolved.
 
**1. Map**
- Added clustering for markers on the Charging Sites map & FSE map tab.
- Fixed the default zoom issue that caused markers not to appear on the Charging Sites map.
- Resolved an issue on the FSE map where empty Charging Sites without FSEs were not shown.
- Removed the map from the Schedule view.

**2. Enhanced Director permissions**
- Director issuance is now permitted at the end of the process.
- When a Director acts as an Analyst or Compliance Manager, the browser no longer redirects to the Compliance Reporting list.
- Fixed the issue where the non-assessment checkbox was not showing in supplemental reports.

**3. Charging equipment**
- Fixed the issue where the Save and return button was disappearing.
- Resolved issues with validation messages.
- Tests have been updated.


hamed-valiollahi:
Thanks for the review, Alex! 👍

PR #3686, opened 2025-12-30
Feat: BCGrid Tables Default Widths - 3636

AlexZorkin:
- Updated BCGrid tables default column widths and min widths

### DIFF
--- a/backend/lcfs/tests/charging_site/test_charging_site_repo.py
+++ b/backend/lcfs/tests/charging_site/test_charging_site_repo.py
@@ -1,5 +1,5 @@
 import pytest
-from unittest.mock import AsyncMock, MagicMock, patch
+from unittest.mock import AsyncMock, MagicMock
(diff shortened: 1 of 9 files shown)


hamed-valiollahi (approved):
Looks good to me! Thanks 👍

prv-proton (approved):
Thanks for the changes — overall, this looks great. Just a few things to note:

Since defaultMinWidth has been reduced to 50, columns that don’t explicitly define a minWidth in the column definitions appear cramped on initial load.
Example: FSE Index page on the Charging Site maintenance pages.

In a few places, the width property is used instead of minWidth in the column definitions. This prevents users from resizing columns (expanding or shrinking).
Example: FSE Processing editor page accessed via the compliance report.

AlexZorkin:
I updated all the inconsistencies in other grid tables so they all work the same way now.

PR #3688, opened 2025-12-31
Hotfix: Fuel code expiry - 3682

AlexZorkin:
- added new date tracking field to fuel code to prevent multiple emails being sent for expired fuel codes
- when expiration date is updated on fuel code, notification sent date is reset
- updated tests


hamed-valiollahi (approved):
Looks good to me! Thanks for the fix 👍

PR #3704, opened 2026-01-05
Feat: Org Snapshot Geocoding - 85

AlexZorkin:
-  Added lat lon to organization snapshot prioritizing address types
- added geocoding script that generates sql update statements
- meta base view added for fuel supply lookups


prv-proton (approved):
Looks good 👍

PR #3767, opened 2026-01-15
Feat: Delete Gov Notification - 3760

AlexZorkin:
- Added delete endpoint and feature for gov notifications
- only available for compliance manager and director on the IDIR side


prv-proton (approved):
👍

PR #3770, opened 2026-01-15
Fix: Compliance Year URL - 3731

AlexZorkin:
- frontend views no longer use the url parameter for display purposes and properly pull it from the compliance report object


prv-proton (approved):
Looks good 👍

PR #3851, opened 2026-02-06
Fix: out of range transactions fix on supplemental txns - 3821

AlexZorkin:
# Fix Summary

**Issue:** Supplemental reports assessed after the compliance period ended were excluded from Line 17, causing incorrect Line 22 values.

**Root Cause:** Code checked transaction `create_date` instead of the report's `compliance_period` to determine if a transaction should be included in the period's available balance.

**Fix:** Modified `calculate_line_17_available_balance_for_period()` to:
1. Join to `CompliancePeriod` table to get the report's period year
2. Check if `report_compliance_period <= target_period` instead of `create_date <= period_end_date`
3. Maintain fallback to `create_date` for historical data without period info

**Result:** Supplemental reports are now correctly included in Line 17 based on which period they belong to, regardless of when they were assessed. This automatically fixes Line 22 since it's calculated as `Line 17 + Line 20`.


prv-proton (approved):
Looks good 👍

PR #3852, opened 2026-02-06
Fix: removed Fuel code from data table output - 3842

AlexZorkin:
- Filtering out Fuel code options from table output


prv-proton (approved):
👍

PR #3880, opened 2026-02-11
Fix: LCFS - BUG - redundant Organization ID in filter

### LINKED ISSUE #3854: LCFS - BUG - redundant Organization ID in filter
**Describe the Bug**  
When selecting an Organization from the filter select list, the filter display shows both the Organization Name and the Organization ID. The Organization ID appears as a duplicate and should not be displayed.

Image

**Expected Behaviour**  
Only the Organization Name should appear in the applied filter display/pill when an Organization is selected from the select list. The Organization ID should be hidden.

**Actual Behaviour**  
The applied filter display includes the Organization ID in addition to the Organization Name, resulting in a duplicate and cluttered filter pill.

**Implications**  
- Creates confusion for users by displaying redundant information.  
- Makes the filter display harder to read and less consistent with expected UI behavior.  
- Reduces clarity when multiple filters are applied.

**Steps To Reproduce**  
User/Role: Any user with access to a page that includes the Organization filter  

1. Go to a page with an Organization select-list filter.  
2. Open the Organization select list.  
3. Select an Organization.  
4. Observe the applied filter display.  
5. See that both the Organization Name and Organization ID are shown instead of only the Organization Name.

prv-proton:
Fixes #3854

### DIFF
--- a/backend/lcfs/web/api/calculator/services.py
+++ b/backend/lcfs/web/api/calculator/services.py
@@ -225,7 +225,7 @@ class CalculatorService:
         for row in data:
             # Skip rows where determining CI is "Fuel code"
             # Only show Default CI and Prescribed CI rows
-            if row.provision_of_the_act == "Fuel code":
+            if row.provision_of_the_act and row.provision_of_the_act.startswith("Fuel code"):
                 continue
 
             # Create a unique key for each combination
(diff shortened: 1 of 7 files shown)


hamed-valiollahi (approved):
Looks good to me! 👍

PR #3891, opened 2026-02-12
Fix: UTC to Pacific time conversion on dates

AlexZorkin:
- updated recorded date and approval dates to show in pacific time


hamed-valiollahi (approved):
Looks good to me! 👍

PR #3899, opened 2026-02-13
Fix: LCFS - FSE status filter

prv-proton:
- Filter list for status column
- create a new version and a record when a validated charging_site/charging_equipment is modified.
- Fix all the filters for FSE index page.
- Suggestion Editor fix to work with enter key


AlexZorkin (approved):
Well done, nice fixes on the chain logic. 👍

PR #3920, opened 2026-02-20
Feat: LCFS - FSE reporting, edit mode checkbox

### LINKED ISSUE #3859: LCFS - FSE reporting, edit mode checkbox
**Description:**  
Update the compliance report FSE edit behavior so that unchecking a row checkbox does not immediately delete entered data. When a BCeID or IDIR user unchecks a row containing **Dates of supply**, **kWh usage**, or **Compliance notes**, the system should lock the row and remove it from display mode rather than deleting the values immediately. Once the user returns to display mode and returns to edit mode, the previously entered values should no longer be present.

**Wireframe:**  
See existing FSE edit and display mode screens.  
Behavioral change only; no layout or visual redesign required.

This change prevents accidental data loss while users are actively editing compliance report FSE data. Locking and hiding unchecked rows provides a clearer, less destructive workflow and aligns user expectations when toggling row inclusion during report preparation.

**Acceptance Criteria:**  
- When a user unchecks an FSE row checkbox in edit mode:
  - The row is locked.
  - Entered values (**Dates of supply**, **kWh usage**, **Compliance notes**) are not immediately deleted.
- The unchecked row does not appear in display mode after being unchecked.
- If the user re-enters edit mode after unchecking the row:
  - The previously entered values are no longer present.
  - The row behaves as a fresh, unchecked entry.
- Behavior is consistent for both BCeID and IDIR users.
- No unintended side effects occur to validation, saving, or submission logic.

**Development Checklist:**  
- [ ] Review current FSE edit/display mode toggle logic  
- [ ] Update uncheck behavior to lock and hide rows instead of deleting immediately  
- [ ] Ensure values persist only until exiting edit mode  
- [ ] Confirm values are cleared when returning to edit mode  
- [ ] Test behavior for both BCeID and IDIR roles  
- [ ] Regression test FSE validation and save flows  

**Notes:**  
- This change applies only to compliance report FSE edit mode.  
- Ensure the behavior is consistent across all report statuses where editing is permitted.

prv-proton:
Fixes #3859

### DIFF
--- a/frontend/src/views/FinalSupplyEquipments/__tests__/FinalSupplyEquipmentReporting.test.jsx
+++ b/frontend/src/views/FinalSupplyEquipments/__tests__/FinalSupplyEquipmentReporting.test.jsx
@@ -20,7 +20,8 @@ vi.mock('@/hooks/useFinalSupplyEquipment', () => ({
   useGetFSEReportingList: vi.fn(),
   useSaveFSEReporting: vi.fn(),
   useDeleteFSEReportingBatch: vi.fn(),
-  useSetFSEReportingDefaultDates: vi.fn()
+  useSetFSEReportingDefaultDates: vi.fn(),
+  useUpdateFSEReportingActiveStatus: vi.fn()
 }))
 
 vi.mock('@/hooks/useComplianceReports', () => ({
@@ -55,7 +56,8 @@ import {
   useGetFSEReportingList,
   useSaveFSEReporting,
   useDeleteFSEReportingBatch,
-  useSetFSEReportingDefaultDates
+  useSetFSEReportingDefaultDates,
+  useUpdateFSEReportingActiveStatus
 } from '@/hooks/useFinalSupplyEquipment'
 import { useComplianceReportWithCache } from '@/hooks/useComplianceReports'
 import { handleScheduleSave } from '@/utils/schedules'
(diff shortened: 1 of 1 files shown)


backend/lcfs/web/api/final_supply_equipment/repo.py
```diff
@@ -991,6 +998,23 @@ async def update_fse_reporting(
         await self.db.flush()
         return {"id": charging_equipment_compliance_id, **data}
 
+    @repo_handler
+    async def update_reporting_active_status(
```
AlexZorkin: No authorization scoping on `update_reporting_active_status` — any user can toggle any org's records
The repo method only filters by `reporting_ids`:
```python
.where(
    ComplianceReportChargingEquipment.charging_equipment_compliance_id.in_(
        reporting_ids
    )
)
```
It doesn't scope by `organization_id` or `compliance_report_id`, even though the schema includes these fields. A supplier user who knows (or guesses) `reporting_ids` belonging to another organization could toggle their records.

Compare with `bulk_update_reporting_dates` which correctly scopes:
```python
.where(and_(
    ...in_(data.equipment_ids),
    ...compliance_report_id == data.compliance_report_id,
    ...organization_id == data.organization_id,
    ...is_active.is_(True),
))
```

The fix should add `organization_id` and `compliance_report_id` to the WHERE clause. Note: the existing `delete_fse_reporting_batch` has the same gap — that's pre-existing tech debt, but this PR shouldn't replicate it.

AlexZorkin (changes requested):
Migration order then one business logic issue to review. ty

AlexZorkin (approved):
changes look good 👍

PR #3931, opened 2026-02-24
Feat: Historical Supplemental Updates

AlexZorkin:
- EER default added for pre 2024 years
- Grey out un-assessed report columns in summary compare


prv-proton (approved):
Looks good.👍 Ty

PR #4171, opened 2026-03-30
feat: enable analyst assignment for compliance managers and directors

AlexZorkin:
## Summary
- Extended the analyst assignment feature on the compliance report list view to allow **Compliance Manager** and **Director** IDIR roles to assign analysts, in addition to the existing Analyst role
- Updated both backend API endpoint authorization (`/analysts` and `/{report_id}/assign`) and frontend role check in `AssignedAnalystCell`

## Test plan
- [ ] Log in as a Compliance Manager and verify the analyst assignment dropdown appears on the compliance report list
- [ ] Log in as a Director and verify the same
- [ ] Verify Analyst role still works as before
- [ ] Verify supplier users still cannot see the column
- [ ] Verify non-IDIR users without the required roles still see read-only view


prv-proton (approved):
👍

PR #4212, opened 2026-04-02
Fix: LCFS - validation logic on the Fuels for other use

### LINKED ISSUE #4193: LCFS - validation logic on the Fuels for other use
**Describe the task**  
Update validation logic on the Fuels for Other Use schedule so that when "Other" is selected for the "Expected use" field, the "If other, enter expected use" field becomes mandatory. Currently, the row passes validation and saves even when this field is left null.

**Purpose**  
To ensure data integrity by requiring users to provide additional details when "Other" is selected, preventing incomplete or ambiguous entries.

**Acceptance Criteria**  
- [ ] When "Other" is selected in the "Expected use" field, the "If other, enter expected use" field is enabled.  
- [ ] The "If other, enter expected use" field is required when enabled.  
- [ ] Validation prevents saving the row if the field is empty or null when required.  
- [ ] Appropriate validation message is displayed to the user when the field is missing.  
- [ ] Validation does not trigger when "Other" is not selected.  
- [ ] Existing functionality for other "Expected use" values remains unchanged.

**Additional context**  
- Current behaviour allows saving with a null value, which leads to incomplete data.  
- Likely requires both frontend and backend validation updates to ensure consistency.

prv-proton:
Fixes #4193

### DIFF
--- a/backend/lcfs/tests/other_uses/test_other_uses_validation.py
+++ b/backend/lcfs/tests/other_uses/test_other_uses_validation.py
@@ -1,8 +1,9 @@
 import pytest
 from unittest.mock import MagicMock
 from fastapi import HTTPException, Request
+from fastapi.exceptions import RequestValidationError
 
-from lcfs.web.api.other_uses.schema import OtherUsesCreateSchema
+from lcfs.web.api.other_uses.schema import OtherUsesCreateSchema, OtherUsesSchema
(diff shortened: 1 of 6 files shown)


hamed-valiollahi (approved):
Nice, solid implementation! Thank you 👍

PR #4219, opened 2026-04-07
Fix: updated dates to use UTC - 4215

AlexZorkin:
- updated date date time objects to use UTC so we have consistent times in the database.


hamed-valiollahi (approved):
Nice improvement with UTC consistency 👍

PR #4232, opened 2026-04-10
Chore: backfill materialized view definition in migrations

AlexZorkin:
- update a previous materialized view definition with correct commands


prv-proton (approved):
👍

Issue #3864, opened 2026-02-09
LCFS - Historical TFRS build testing


Grulin:
**Describe the task**
There seems to still be a few things to iron out in the historical TFRS compliance reports being pulled over in to the LCFS Portal:

1) Report ID 49 showing an energy content value of 35.40 in TFRS but only 35 in the LCFS Portal. 
2) Reports 370 and 755 are showing both the original report Fuel Supply in that schedule and the supplemental 1 lines reported. It should only show the most recent report. Not seeing change log here. **This might be related to a broader issue in Test**

**Purpose**
Further refines the pull over of the historical data pull over. 

**Acceptance Criteria**
- [ ] cell values in the LCFS Portal schedules are correctly displaying the same figure as in TFRS, even decimal places
- [ ] Only the latest report information is showing in the Fuel Supply section of the compliance report. 

**Additional context**

Grulin:
Retested this card again and these reports are still missing the change log toggle

PR #4280, opened 2026-04-20
fix: enable change-log toggle for delete-only supplementals (#3864)

AlexZorkin:
## Summary

Closes part of [#3864](https://github.com/bcgov/lcfs/issues/3864): the change-log toggle was disabled on Fuel Supply (and other activity sections) for historical TFRS reports 370 and 755 even though both have a real supplemental version in the chain.

**Root cause** — the toggle's enabled state was inferred client-side from rows in the activity response. \`mode=view\` filters out any group whose latest action is \`DELETE\`, so for supplementals whose only edits were deletions, every surviving row mapped back to the base report (\`version=0\`) and \`wasEdited\` returned false. Reports 370/755 fall into this case (the supplemental only deleted rows from the original).

**Fix** — compute the flag server-side. New \`ComplianceReportRepository.has_supplemental_changes(report_id, model)\` returns true if any record of the given activity model belongs to a chain entry with \`version > 0\`. Each of the 5 activity list schemas now exposes \`was_edited\`, and the frontend reads \`data.wasEdited\` instead of inferring from row contents. Per-section accuracy is preserved — toggles still enable only on sections that actually have edits in the chain.

DB sanity check on local data:

| report_id | fuel_supply_was_edited |
|---|---|
| 370 (supp.) | t |
| 755 (supp.) | t |
| 754 (base)  | f |
| 3657 (base) | f |

## Test plan

- [ ] Open compliance report 755 → expand Fuel Supply → "Change log" toggle is enabled and shows the deletion history when clicked
- [ ] Open compliance report 370 → same behaviour
- [ ] Open a non-supplemental report (e.g. 754) → "Change log" toggle is disabled (no chain)
- [ ] Open a non-TFRS supplemental that *did* edit Fuel Supply (CREATE/UPDATE) → toggle still enables (existing behaviour preserved)
- [ ] Open a non-TFRS supplemental that *did not* edit Allocation Agreement → that section's toggle stays disabled (per-section accuracy preserved)
- [x] Backend: \`pytest lcfs/tests/{fuel_supply,fuel_export,notional_transfer,other_uses,allocation_agreement,compliance_report}/\` — 657 passed
- [x] Frontend: \`vitest run ReportDetails.test.jsx\` — 22 passed

### DIFF
--- a/backend/lcfs/tests/allocation_agreement/test_allocation_agreement_services.py
+++ b/backend/lcfs/tests/allocation_agreement/test_allocation_agreement_services.py
@@ -1,5 +1,5 @@
 import pytest
-from unittest.mock import MagicMock
+from unittest.mock import AsyncMock, MagicMock
 
 from lcfs.db.base import ActionTypeEnum
 from lcfs.db.models.compliance.AllocationAgreement import AllocationAgreement
@@ -7,6 +7,7 @@ from lcfs.web.api.allocation_agreement.schema import (
     AllocationAgreementCreateSchema,
 )
 from lcfs.web.api.allocation_agreement.services import AllocationAgreementServices
+from lcfs.web.api.compliance_report.repo import ComplianceReportRepository
 from lcfs.web.api.compliance_report.services import ComplianceReportServices
 from lcfs.web.api.compliance_report.dtos import ChangelogAllocationAgreementsDTO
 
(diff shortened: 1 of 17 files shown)


hamed-valiollahi (approved):
Looks great! Thank you 👍

PR #4281, opened 2026-04-21
fix: LCFS - Line 21/22 ignore prior-assessed credits when tx finalized post-deadline (#4235)

### LINKED ISSUE #4235: LCFS - Incorrect Line 21 penalty and Line 22 balance in prod; report ID 3400
**Describe the Bug:**

- In a multi-supplemental reporting scenario, Line 21 (Non-compliance penalty payable) and Line 22 (Available compliance unit balance at the end of the compliance period) are calculated incorrectly.

- The system derives a secondary delta by comparing Line 20 values across supplemental reports, even though Line 20 already represents the net change from assessment. 

**Expected Behaviour:**
- Line 20 in each supplemental report must be treated as the authoritative net change from that
  assessment.
- Line 22 must reflect the remaining issued compliance units held by the organisation following
  the most recent assessed supplemental.
- Line 21 must only apply where a true compliance unit shortfall exists.

**Actual Behaviour:**
- Line 20 values are re-compared across supplemental reports as though they were cumulative values.
- This creates a “delta of a delta” and overstates the reduction in compliance units.
- Line 22 is reduced to zero despite the organisation continuing to hold issued compliance units.
- Line 21 applies a non-compliance penalty based on the incorrect zero balance.

**Implications:**
- Incorrect non-compliance penalties are generated.
- Reduction in balance due to credit decrease is not shown in line 22
- Manual analyst intervention is required to explain discrepancies.

**Steps to Reproduce:**
1. Go to the summary section of report ID 3400
2. View the summary table with a Delta column comparing the two supplementals.
3. Observe that Line 21 and Line 22 reflect a full elimination of credits rather than a revision.

**Acceptance Criteria:**
- Line 21 must not apply a penalty where issued compliance units remain following reassessment.
- Line 22 must reflect the remaining issued compliance units after the latest assessment.
- Summary logic must respect that Line 20 is already a net change in reduction scenarios

**Additional Context:**
The initial issuance associated with the first assessed supplemental report is the only compliance
unit transaction this organisation has ever undertaken, and the organisation continues to hold
those credits. Line 15, Line 17, and Line 20 are correct. The error is isolated to summary logic
affecting Line 21 and Line 22. This issue requires a manual patch in production to correct
affected records.

**There may be other affected reports, but this one is high priority as it is showing a penalty when it should not**

AlexZorkin:
## Summary
- Fixes [#4235](https://github.com/bcgov/lcfs/issues/4235). In a multi-supplemental scenario, if a prior version's assessment was finalized **after** the compliance period end (March 31, year + 1), its Adjustment transaction is intentionally excluded from Line 17's period-end snapshot. The organisation still holds those credits, but `Line 22 = max(line_17 + line_20, 0)` then zeros out the balance and `Line 21` applies a false non-compliance penalty on the reassessment's Line 20 deficit.
- New repo helper `TransactionRepository.get_group_adjustments_excluded_from_line_17` sums Adjustment transactions in the same report group with `update_date > compliance_period_end`. `summary_service.calculate_low_carbon_fuel_target_summary` folds that amount into an `effective_available_balance` used **only** for Line 21 and Line 22. Lines 15, 17, and 20 are unchanged (they were already correct per the bug report).
- Handles multi-level supplemental chains — any combination of pre- or post-deadline assessments along the chain resolves correctly.

## Affected production reports
Two Reports (identified via the detection query below). Manual data patch for both is provided separately to the analyst applying the prod fix — not included in this PR.

## Test plan
- [x] New regression test `test_reassessment_after_deadline_retains_prior_issuance_for_line_22` covers the reassessment-after-deadline scenario.
- [x] Full backend suite green locally (`pytest lcfs/tests` — 1441 passed, 1 skipped).
- [ ] Verify report 3400 summary in dev: Line 21 = 0, Line 22 reflects remaining issued balance after pulling the branch.
- [ ] Verify report 3546 summary in dev: Line 21 = 0, Line 22 reflects remaining issued balance.
- [ ] Confirm an original (v0) report and a pre-deadline supplemental chain are unaffected (no change in Line 21/22).
- [ ] Run the detection query on prod and review any additional candidates before applying the data patch.

### DIFF
--- a/backend/lcfs/tests/compliance_report/conftest.py
+++ b/backend/lcfs/tests/compliance_report/conftest.py
@@ -173,6 +173,7 @@ def mock_trxn_repo():
     repo = AsyncMock(spec=TransactionRepository)
     repo.calculate_available_balance_for_period = AsyncMock(return_value=2000)
     repo.delete_transaction = AsyncMock()
+    repo.get_group_adjustments_excluded_from_line_17 = AsyncMock(return_value=0)
     return repo
 
 
--- a/backend/lcfs/tests/compliance_report/test_summary_service.py
+++ b/backend/lcfs/tests/compliance_report/test_summary_service.py
@@ -1129,6 +1129,88 @@ async def test_supplemental_report_uses_existing_summary_line_17(
 
 
 @pytest.mark.anyio
+async def test_reassessment_after_deadline_retains_prior_issuance_for_line_22(
+    compliance_report_summary_service, mock_trxn_repo, mock_summary_repo, mock_repo
+):
+    """
+    Regression test for bug #4235.
+
+    When a prior supplemental was assessed AFTER the compliance period end
+    (March 31, year + 1), its Adjustment transaction is excluded from Line 17.
+    However, the organization still holds those credits, so a new supplemental
+    that reduces the supply must not wipe the balance to zero or apply a
+    penalty as long as units remain.
+
+    Scenario (based on prod report 3400):
+      - Prior assessed v1 issued 795 credits (post-deadline -> NOT in Line 17).
+      - Current v2 proposes Line 18 = 567 -> Line 20 = -228.
+      - Expected Line 22 = 567 (remaining issued credits), Line 21 = 0.
+    """
+    compliance_period_start = datetime(2024, 1, 1)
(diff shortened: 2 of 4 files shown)


prv-proton (approved):
👍

PR #4283, opened 2026-04-21
Fix: LCFS - Fossil-derived diesel supplemental line missing from Metabase for CR #1609

prv-proton:
Fix #1609


hamed-valiollahi (approved):
Looks good to me 👍

Issue #4269, opened 2026-04-17
LCFS - Penalties incorrectly visible for reports recommended for exemption


Grulin:
**Describe the Bug:**
In Production, when an analyst recommends a compliance report for exemption that has a renewable fuel or low carbon fuel penalty attached, penalties are not consistently removed from the report outputs. While most penalty amounts are hidden, penalties continue to appear in specific summary locations and data outputs despite the exemption recommendation.

Currently impacting report 3472 that is in a recommended status.

**Expected Behaviour:**
Once a compliance report is recommended for exemption by an analyst:
-No penalties should be visible anywhere in the report.
-No penalties should appear in any summary tables or summary lines.
-No penalties should appear in any descriptive text.
-No penalties should be included in any data exports or data outputs.

**Actual Behaviour:**
When a report recommended for exemption includes a penalty:
1. The penalty still appears in the *Non-compliance penalty payable* summary table under the line “Total non-compliance penalty payable”.
2. If the supplier has a debit owing position, the debit amount is still visible in the description text for line 21 in brackets.
3. Penalty values may still be present in underlying data outputs, despite not being intended for display (unsure about this, but worth checking)

**Implications:**
-Creates confusion for analysts, managers, and suppliers reviewing exempted reports.
-Undermines the intent of an exemption recommendation by continuing to surface penalties.
-Increases risk of incorrect interpretation of compliance status and financial liability.
-May result in inconsistent communications with suppliers and internal decision-makers.

**Steps to Reproduce:**
IDIR

1. Open a compliance report in Production with an assessed renewable or low carbon penalty.
2. Set the report status to “Recommended for Exemption” as an analyst.
3. View the compliance summary tables.
4. Review the *Non-compliance penalty payable* summary table.
5. Review line 21 descriptive text for reports with a debit owing position.
6. Review any available data exports or report outputs.

**Acceptance Criteria:**
-When a report is recommended for exemption:
-The “Total non-compliance penalty payable” value is zero or not displayed.
-No debit or penalty values appear in line 21 descriptions.-No penalties appear in any other summary lines.
-No penalties are included in any downloadable or system-generated data outputs

**Additional Context:**
This bug is specific to the “Recommended for Exemption” workflow and occurs in Production.

PR #4294, opened 2026-04-22
fix: LCFS - hide penalties on reports recommended for exemption (#4269)

AlexZorkin:
## Summary
- When an analyst recommends a report for exemption, exemption flags are set on the report but `calculate_compliance_report_summary` / `convert_summary_to_dict` never consulted them — so the non-compliance penalty total, the line 21 "(X units * \$Y CAD per unit)" description, and Excel exports kept surfacing the penalty throughout the Recommended by analyst/manager workflow (fixing [#4269](https://github.com/bcgov/lcfs/issues/4269)).
- Added `_apply_exemption_overrides` helper that zeros lines 4/11 (renewable), 18/20/21 (low carbon), recomputes line 22 from line 17, and rebuilds the non-compliance penalty total + line 11/21 rows based on exemption flags. It also strips the unit/rate breakdown from line 21's description when low-carbon-exempted.
- Helper is invoked at the end of both summary paths, so summary tables, descriptive text, and `/export` Excel output stay consistent from the moment exemption is selected — not just once the report hits Exempted.

## Test plan
- [x] New unit tests in `test_summary_service.py` cover renewable-only, low-carbon-only (description stripped + line 22 recalc), both exemptions, no-op without flags, and defensive null-report handling.
- [x] Full `backend/lcfs/tests/compliance_report/` unit suite passes (382 tests).
- [ ] Verify in a running env: on report 3472 (Recommended for Exemption), "Total non-compliance penalty payable" shows \$0, line 21 description has no debit-in-brackets, and the Excel export reflects the zeros.

### DIFF
--- a/backend/lcfs/tests/compliance_report/test_summary_service.py
+++ b/backend/lcfs/tests/compliance_report/test_summary_service.py
@@ -3093,3 +3093,156 @@ async def test_line_15_16_uses_assessed_report_values(
     # Line 16 should use assessed report's Line 19 value
     line_16 = next(row for row in result.low_carbon_fuel_target_summary if row.line == 16)
     assert line_16.value == 1000, f"Line 16 should be 1000 from assessed report, got {line_16.value}"
+
+
+def _make_summary_schema_with_penalties():
+    """Build a ComplianceReportSummarySchema with non-zero penalty rows for exemption tests."""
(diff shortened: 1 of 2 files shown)


prv-proton (approved):
👍

Issue #4292, opened 2026-04-22
LCFS - Test environment: FSE tabs visible to BCeID signing authority users without compliance reporting role


Grulin:
**Describe the Bug:**  
In the Test environment, BCeID users who have only the Signing Authority role can see three Final Supply Equipment (FSE) tabs. These tabs should not be visible to users without the Compliance Reporting role.

**Expected Behaviour:**  
The three FSE tabs are visible only to users who have the Compliance Reporting role.  
BCeID users with only the Signing Authority role should not see any FSE tabs.

**Actual Behaviour:**  
BCeID users with only the Signing Authority role can see three FSE tabs in the application.

**Implications:**  
- Creates role-based access inconsistencies.  
- May confuse Signing Authority users about actions they are permitted to take.  
- Increases risk of inappropriate interaction with FSE functionality.

**Steps to Reproduce:**  
1. Log into the Test environment as a BCeID user.  
2. Ensure the user has only the Signing Authority role and does not have the Compliance Reporting role.  
3. Navigate to the FSE section.  
4. Observe that three FSE tabs are visible.

**Additional Context:**  
FSE functionality should be restricted to Compliance Reporting users only. This aligns with existing role-based access expectations for external users.

PR #4296, opened 2026-04-22
fix: LCFS - gate FSE and charging-sites tabs on compliance_reporting role (#4292)

AlexZorkin:
## Summary
- In Test, BCeID users with only the Signing Authority role could see the three FSE-related tabs (Manage FSE, FSE map, Manage charging sites) in the compliance reporting menu — the gate in [ReportsMenu.jsx](frontend/src/views/ComplianceReports/ReportsMenu.jsx) only checked the feature flag, not the user's role (fixing [#4292](https://github.com/bcgov/lcfs/issues/4292)).
- Added a `hasRoles(roles.compliance_reporting)` check to the non-IDIR path of both `canAccessChargingSitesTab` and `canAccessFseTab` so signing-authority-only users no longer see the tabs even when the flags are on.
- Tightened the route-level `<Role>` wrapper around the FSE / charging-sites outlets from `roles.supplier` to `roles.compliance_reporting` so direct URL access is also blocked.

## Test plan
- [x] New vitest suite in `ReportsMenu.test.jsx` covers: tabs hidden for signing-authority-only; visible for compliance-reporting + flag on; hidden for compliance-reporting + flag off; visible for IDIR regardless of flag; direct URL to `/fse` and `/charging-sites` blocked for signing-authority-only and allowed for compliance-reporting.
- [x] Broader `ComplianceReports` frontend test suite passes (705 tests).
- [ ] Verify in Test env: log in as BCeID user with only Signing Authority — no FSE / charging-sites tabs visible, direct URLs redirect/null render.

### DIFF
--- a/frontend/src/views/ComplianceReports/ReportsMenu.jsx
+++ b/frontend/src/views/ComplianceReports/ReportsMenu.jsx
@@ -27,9 +27,14 @@ export function ReportsMenu() {
   const location = useLocation()
   const { hasAnyRole, hasRoles } = useCurrentUser()
   const isIDIR = hasAnyRole(...govRoles)
+  const hasComplianceReporting = hasRoles(roles.compliance_reporting)
   const canAccessChargingSitesTab =
-    isIDIR || isFeatureEnabled(FEATURE_FLAGS.MANAGE_CHARGING_SITES)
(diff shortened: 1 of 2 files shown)


prv-proton (approved):
👍

PR #4332, opened 2026-04-28
Fix: LCFS - Performance issues

prv-proton:
Root cause: `ReportDetails` was eagerly calling every schedule endpoint on initial render, and each endpoint ran version-history resolution queries. As compliance-report volume increased, that fan-out made schedule loading noticeably slower.

Changes made:
- Added a lightweight `GET /reports/:reportID/schedule-overview` endpoint to return per-accordion metadata only:
  - record count
  - active/deleted state
  - supplemental edited flag
  - FSE capability flag
- Refactored `ReportDetails` to use the overview first for visibility, chips, disabled state, and auto-expand decisions.
- Kept the existing UX where accordions with data auto-expand on load.
- Changed full schedule payload fetching to lazy loading:
  - if overview shows data, the accordion is shown immediately
  - the full request starts immediately
  - users see a loading spinner in the expanded accordion while the payload is still loading
- Added frontend guards for async gaps so accordions do not crash while data is still resolving:
  - protected schedule render paths from `undefined`
  - normalized supporting document data before rendering
- Added DB indexes for compliance schedule history/version lookup patterns:
  - `compliance_report (compliance_report_group_uuid, version, compliance_report_id)`
  - versioned schedule indexes on `fuel_supply`, `fuel_export`, `allocation_agreement`, `notional_transfer`, and `other_uses`
  - `final_supply_equipment (compliance_report_id, create_date)`
- Added a quarterly off-hours compliance-table reindex job:
  - January, April, July, October
  - day 1 at 03:15 am PST
  - Can be altered to run once at the startup using env variable.
  - configurable via backend settings

### DIFF
--- a/backend/lcfs/db/migrations/versions/2026-04-20-09-00_d4f5e6a7b8c9.py
+++ b/backend/lcfs/db/migrations/versions/2026-04-20-09-00_d4f5e6a7b8c9.py
@@ -1,0 +1,84 @@
+"""Add compliance report schedule performance indexes
+
+Revision ID: d4f5e6a7b8c9
+Revises: c1d2e3f4a5b6
+Create Date: 2026-04-20 09:00:00.000000
+"""
+
+from alembic import op
+
+
+# revision identifiers, used by Alembic.
+revision = "d4f5e6a7b8c9"
+down_revision = "c1d2e3f4a5b6"
+branch_labels = None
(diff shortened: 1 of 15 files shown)


AlexZorkin (approved):
Looks great, lets get this merged 👍

PR #4356, opened 2026-05-04
chore: reduce notifications/count polling and quiet wildcard-role log

AlexZorkin:
## Summary
- Bump `useNotificationsCount` `refetchInterval` from 1m to 5m, set `refetchIntervalInBackground: false`, and keep `refetchOnWindowFocus: true` — badge still feels fresh during active use, but idle/background tabs stop hammering the backend. Existing `invalidateQueries` after read/delete mutations covers in-app updates.
- Downgrade the per-request `Endpoint ... is accessible by all roles` log from `warn` to `debug`. It's static per route (wildcard role list is known at decoration time) and was producing one warn per request — drowning out backend logs, especially from the notifications poll.

## Test plan
- [ ] Sign in, confirm the notifications badge still updates after marking/deleting notifications without a manual refresh.
- [ ] Leave the app open in a background tab for 10+ minutes; confirm no `/notifications/count` calls fire until the tab is refocused.
- [ ] Tail backend logs and confirm the `is accessible by all roles` warnings are gone at default log level.


prv-proton (approved):
👍

PR #4374, opened 2026-05-08
Feat: Add My Fuel Codes Page - 4168

### LINKED ISSUE #4168: LCFS - Create a new My fuel codes page
**Description:**  
Create a new **“My fuel codes”** sub tab/page under the **Fuel codes** main tab for BCeID users.  

This page will display fuel codes associated with the user’s organization and will include both **active and archived** fuel codes. The layout will follow the same structure as existing internal fuel code pages.

The page will be visible only to BCeID users with the **CI applicant** role.

**Wireframe:**  
https://preview.uxpin.com/77159fa6e4246f73978bd68f8170c8bc42744cb7#/pages/220751477

Allows BCeID users to view and manage all fuel codes associated with their organization in one place, including both active and archived records.

**Acceptance Criteria:**  
- A new **My fuel codes** sub tab/page exists under the Fuel codes tab.  
- Page is visible only to BCeID users with the **CI applicant** role.  
- Page displays fuel codes associated with the user’s organization.  
- Both **active and archived** fuel codes are included in the list.  
- Page layout matches existing IDIR fuel code page.  
- Page loads and displays data correctly.

**Development Checklist:**  
- [ ] Create My fuel codes route/page under Fuel codes.  
- [ ] Implement role-based visibility (BCeID CI applicant only).  
- [ ] Query and display fuel codes for the user’s organization.  
- [ ] Include both active and archived fuel codes.  
- [ ] Match layout and behavior of existing fuel code pages.  
- [ ] Test with different BCeID users and data scenarios.

**Notes:**  
- Ensure consistency with existing fuel code UI patterns.  
- Confirm archived records are clearly distinguishable if needed.

hamed-valiollahi:
This PR adds the My Fuel Codes page for BCeID CI applicants.

- Add My Fuel Codes page
- Restrict page to CI applicants
- Display organization fuel codes
- Include active and archived records
- Match existing fuel code layout

Closes #4168


backend/lcfs/web/api/fuel_code/repo.py
AlexZorkin, line 1: one concern: filtering by company name rather than organization id

Three problems with name-based scoping:

- The schema in this very PR includes a formerCompany column (_schema.jsx:441-444). That's direct evidence the system tracks org name changes — meaning fuel codes carry a company value that may not equal the org's current name. A CI applicant whose org was renamed will silently lose access to their pre-rename fuel codes.
- Whitespace / punctuation / suffix drift ("ACME Corp" vs "ACME Corp." vs "ACME Corporation"). Lowercasing doesn't fix any of these. A typo on either side hides records with no error.
- Cross-tenant collisions — if two organizations happen to share a name (or one's company string equals another org's name), a CI applicant would see another org's fuel codes. Unlikely in practice but it's not a principled scope.

The fix is to filter on the organization foreign key, not the string. FuelCodeListView should expose organization_id, pass request.user.organization.organization_id from the view down to the repo and condition on that.

AlexZorkin (changes requested):
just one concern to address otherwise looks good.

backend/lcfs/web/api/fuel_code/repo.py
hamed-valiollahi: Thanks for the feedback, Alex, appreciate the detailed review.

That was actually one of my concerns as well, since I wasn’t sure if updating the Add Fuel Code flow was part of this ticket or not.

I’ve now applied the change and updated the Add Fuel Code flow to use `organization_id`.

What I’m not fully sure about now is the existing fuel code records. Since we added a new `organization_id` column, I assume we need some kind of backfill for existing rows too.

Any thoughts on the best way to handle that? Maybe comparing the company name with the organization legal name or operating name?

AlexZorkin: we should have a migration as part of this that does the backfill. lets chat about it if there are any other concerns. ty

hamed-valiollahi: Thanks Alex. I've added the backfill migration and pushed the changes.

AlexZorkin (approved):
Thanks for the backfill 👍

PR #4382, opened 2026-05-12
Fix: LCFS - Decommissioned FSE reporting 4360

prv-proton:


backend/lcfs/db/sql/views/metabase.sql
AlexZorkin, line 1: The same correlated subquery is inlined 4 times here (SELECT, two CASE WHEN checks, and the division). Postgres won't deduplicate them, so each row runs the join+sort+limit up to 4× instead of once. Please move it back into a LEFT JOIN LATERAL ... power_lookup (or a CTE keyed on charging_equipment_id) and reference power_lookup.power_output everywhere — single source of truth, and a real perf win on long date ranges.

backend/lcfs/web/api/final_supply_equipment/services.py
```diff
@@ -501,9 +512,24 @@ async def get_fse_reporting_list_paginated(
                 only_active=(mode == "summary"),
             )
 
-        data, total = await self.repo.get_fse_reporting_list_paginated(
-            organization_id, pagination, report.compliance_report_id, mode
-        )
+        refresh_decommissioned_fse = self._should_refresh_decommissioned_fse(report)
```
AlexZorkin: Bit of an anti-pattern to have update on a get list request. Better home for this is the equipment-status transition (cascade when charging_equipment becomes Decommissioned), or a dedicated "refresh" call the UI invokes when entering edit mode.

AlexZorkin (changes requested):
Couple of things to consider.

backend/lcfs/db/sql/views/metabase.sql
prv-proton: fixed

AlexZorkin (approved):
looks good ty 👍

PR #4368, opened 2026-05-07
fix: prevent duplicate compliance report submissions (#4367)

AlexZorkin:
## Summary
- Disables the Cancel button and the close (X) icon in `BCModal` while the primary action is in flight
- Blocks the dialog `onClose` (backdrop / Esc) while loading
- Stops BCeID suppliers from cancelling and re-submitting a compliance report mid-request, which previously generated duplicate IDIR notifications

Closes #4367

## Test plan
- [ ] Submit a compliance report as a BCeID supplier; confirm Cancel/X are disabled and clicking the backdrop does not close the modal until the request resolves
- [ ] Confirm only one submission notification is created on the IDIR side per submission
- [ ] If the submission errors, confirm the modal becomes interactive again

PR #4393, opened 2026-05-12
fix: lock compliance_report row to prevent concurrent submission duplicates

AlexZorkin:
## Summary

Follow-up hotfix to #4368. The original idempotency fix closed the sequential repeat-click window but left a TOCTOU race open for genuinely concurrent submissions — two in-flight requests both read `compliance_report.transaction_id = NULL`, both create a Reserved transaction, and the second `UPDATE` orphans the first. The orphaned row then surfaces in the transactions UI as a "Legacy Transaction" (its display label for `StandaloneTransaction` in `mv_transaction_aggregate`), which is what we've been seeing in prod since the original hotfix shipped.

- New `ComplianceReportRepository.lock_compliance_report_row` issues `SELECT compliance_report.transaction_id … FOR UPDATE` and returns the persisted `transaction_id`. Bare row lookup — no eager loads — so it doesn't disturb the ORM graph the caller is already using.
- `handle_submitted_status` and `handle_recommended_by_analyst_status` (government reassessment path) take the lock **before** the idempotency check and resync `report.transaction_id` from the lock's return value. Trusting the in-memory ORM attribute would defeat the lock — it was loaded before we blocked on the other writer.
- Existing orphans in prod will be cleaned up via the out-of-band SQL script previously used for incident #4368.

## Race this closes

```
T0  Req A: SELECT compliance_report   -> transaction_id = NULL
T0  Req B: SELECT compliance_report   -> transaction_id = NULL   (A uncommitted)
T1  Req A: idempotency check          -> None
T1  Req B: idempotency check          -> None
T2  Req A: org_service.adjust_balance -> Transaction N   (Reserved)
T2  Req B: org_service.adjust_balance -> Transaction N+1 (Reserved)
T3  Req A: UPDATE compliance_report SET transaction_id = N   commit
T3  Req B: UPDATE compliance_report SET transaction_id = N+1 commit  (last writer wins)
           -> Transaction N is now an orphan
```

With the lock, Req B blocks at T0 until Req A commits at T3, then wakes up, reads the now-persisted `transaction_id`, and 409s out via the existing idempotency guard.

## Test plan

- [x] Updated `test_handle_submitted_status_blocks_duplicate_when_reserved_exists` to mock the lock return value (regression for sequential repeat clicks still passes).
- [x] New `test_handle_submitted_status_lock_resyncs_transaction_id` covers the concurrent-race path: in-memory `transaction_id=None`, lock returns the freshly-persisted id, idempotency check uses the lock-returned id, request 409s, no new transaction or repo update.
- [ ] Manual: in a staging deploy, open two browser tabs to the same draft report, hit Sign-and-Submit in rapid succession from both — verify only one Reserved transaction is created and the second request returns 409.
- [ ] Post-merge: run the orphan-cleanup SQL script against prod to remove the existing duplicates surfaced in the transactions UI.


hamed-valiollahi (approved):
Looks good to me! 👍  Thank you

PR #4395, opened 2026-05-12
Fix: LCFS - Excel download feature on the BCeID Fuel Code

prv-proton:
Fix: It works on the BCeID screen but not on the anonymous page. It should work on the anonymous page as well.


hamed-valiollahi (approved):
Looks good to me! 👍  Thanks

PR #4414, opened 2026-05-21
Feat: LCFS - CI application internal progress bar

prv-proton:
Fixes #4398, #4405


AlexZorkin (approved):
This is looking solid. Just wondering if we're missing a few of the buttons that our PO is looking for. 
"Generate fuel codes" button
"Further documentation requested" state with specific button
"Reactivate application" button
I think we could merge what we have and then get feedback. This process will most likely evolve a bit.

PR #4415, opened 2026-05-21
Feat: Add Redis-backed rate limiting for public endpoints - 4383

hamed-valiollahi:
This PR adds Redis-backed rate limiting for public endpoints.

- Enforce shared request limits
- Protect public export endpoints
- Support per-route overrides
- Return standard 429 responses
- Document limiter usage

Closes #4383


AlexZorkin (changes requested):
One topic to discuss before we merge this in:

# Suggested change: replace CIDR allow-list with Uvicorn proxy-header handling

## Context

The PR introduces `LCFS_RATE_LIMIT_TRUSTED_PROXIES` (a comma-separated CIDR list) and custom logic in `rate_limit.py` (`_parse_trusted_proxies`, `_is_trusted`, the trust check inside `client_ip()`) to decide when `X-Forwarded-For` can be trusted. This is the *correct* defense against header-spoofing — but it requires per-environment configuration and duplicates work that Uvicorn already does.

Because LCFS pods are never exposed directly to the internet (the OpenShift router is the only possible TCP peer), we can delegate this entirely to Uvicorn and eliminate the env var and ~30 lines of custom logic.

## Proposed change

**1. Configure Uvicorn to handle proxy headers** (Dockerfile or startup script):
```
uvicorn lcfs.web.application:get_app --proxy-headers --forwarded-allow-ips '*'
```

Uvicorn rewrites `request.client.host` to the real client IP based on `X-Forwarded-For`. `'*'` is safe here because the only entity that can connect to the pod is the OpenShift router, which overwrites XFF with the real edge IP.

**2. Remove from `backend/lcfs/settings.py`**:
```python
rate_limit_trusted_proxies: str = ""
```

**3. Remove from `backend/lcfs/web/core/rate_limit.py`**:
- `_parse_trusted_proxies()`
- `_is_trusted()`
- The `ipaddress` import
- The trust-check branch inside `client_ip()`

**4. Reduce `client_ip()` to**:
```python
def client_ip(request: Request) -> str:
    return request.client.host if request.client else "unknown"
```

**5. Replace the `LCFS_RATE_LIMIT_TRUSTED_PROXIES` row in the README env-var table** with a note under "Rate limiting":

> The limiter uses `request.client.host`, which Uvicorn populates from `X-Forwarded-For` because LCFS runs behind the OpenShift router (`--proxy-headers --forwarded-allow-ips '*'`). Pods are never exposed directly to the internet, so the router is the only possible peer.

**6. Update tests** in `backend/lcfs/tests/web/test_rate_limit.py`:
- Delete `test_client_ip_ignores_forwarded_when_proxy_not_trusted`
- Delete `test_client_ip_honours_forwarded_when_proxy_trusted`

## Why this is better

- **Zero per-env configuration.** No `LCFS_RATE_LIMIT_TRUSTED_PROXIES` to set in dev/test/prod, no CIDR to look up, no risk of forgetting it and silently throttling all public traffic into one bucket.
- **Single source of truth.** Every consumer of "real client IP" (logs, audit, rate limit, error reports) sees the same value, populated once at the ASGI layer.
- **Smaller attack surface.** Removes ~30 lines of CIDR parsing, header parsing, and trust logic from the limiter in favor of Uvicorn's battle-tested implementation.
- **Standard idiom.** `--proxy-headers --forwarded-allow-ips` is the documented Uvicorn pattern for "I'm behind a trusted proxy."

## Caveat to document

The `'*'` value is safe **only** under the assumption that pods are not directly reachable from the internet. If LCFS is ever exposed via a `NodePort`/`LoadBalancer` that bypasses the router, or run on a VM with a public IP, this would re-open header spoofing. Worth a one-line comment near the Uvicorn invocation so a future operator doesn't get surprised.

AlexZorkin (approved):
Changes look great thanks. Must easier to manage this way. 👍

PR #4439, opened 2026-05-25
Fix: LCFS - BUG - Encyclopedia Brown and the case of the missing FSE

prv-proton:
Fixes #4409


hamed-valiollahi (approved):
Looks great! 👍

Issue #4385, opened 2026-05-12
LCFS - Add director override for credit transfer category assignment


Grulin:
**Describe the task**  
Add radio button control on the credit transfer page that allows a director to manually change the assigned credit category, overriding the automated date-based category logic used by the system. This new control will appear only to users with the Director role after the transfer has been recorded. This will allow updates to any transfer in Recorded status by the director at any time after the transfer has been completed. A modal requires confirmation for the radio select update.

https://preview.uxpin.com/59145260c7d8cc48969ac79401cc03d84843c3d9#/pages/196166639

If the A1 checkbox is chaecked, the A radio button will be automatically selected. If radio B, C or D is selected, the A1 checkbox cannot be checked.

**Purpose**  
This task ensures credit transfers are assigned to the most accurate category for reporting when verified agreement date information differs slightly from what an organisation entered in the system. It removes the need for manual spreadsheet adjustments by analysts and supports consistent, automated reporting through Metabase.

**Acceptance Criteria**
- [ ] A director-only control is available on credit transfer records to override the system-assigned credit category.
- [ ] The overridden category replaces the automated category in all reporting, exports, and analytics outputs.
- [ ] Modal confirmation when a radio selection is made to change the category.
- [ ] Role-based access controls prevent non-director users from applying overrides.

**Additional context**
- Credit categories are currently assigned automatically based on agreement date ranges.
- Analysts have been manually adjusting categories in spreadsheet outputs where agreement date information was later corrected.
- This creates inconsistencies between the portal data and Metabase outputs and is a barrier to full automation.
- Some existing records will require data patches after implementation to align historical portal data with corrected reporting. Separate card for this

PR #4440, opened 2026-05-25
Feat: LCFS - Add director override for credit transfer category assignment

prv-proton:
Closes #4429, #4385


hamed-valiollahi (approved):
Tested locally, everything worked as expected. Solid work, and nice consideration of the Metabase impact. 👍  Thank you

Issue #4431, opened 2026-05-22
LCFS - Investigate compliance report submission lag and network errors in TEST


Grulin:
**LCFS - Investigate compliance report submission lag and network errors**

**Describe the Bug**
Users experience lag and intermittent network errors when creating and submitting compliance reports. The report appears not to submit initially but eventually processes.

**Expected Behaviour**
Compliance reports should submit immediately with clear confirmation and no perceived failure.

**Actual Behaviour**
Submission appears to fail due to lag or network error, then later completes.

**Implications**
- Reduced user confidence in submission process  
- Risk of duplicate submissions  
- Potential support burden for analysts

**Steps to Reproduce**
1. Create a compliance report in test environment  
2. Submit report  
3. Observe lag and intermittent network error message  

**Acceptance Criteria**
- Submission completes without delay or error messaging  
- User receives clear confirmation immediately  
- No duplicate submissions occur due to retry behaviour  

**Additional context**

PR #4466, opened 2026-05-28
fix: speed up compliance report submission (#4431)

AlexZorkin:
## Summary

Investigated the lag + intermittent "network error" reports on compliance report submission ([#4431](https://github.com/bcgov/lcfs/issues/4431)) and shipped the two highest-leverage, lowest-risk fixes.

### 1. Missing indexes on the submission hot path (backend)

Submission's summary recalc fans out across every versioned schedule (`fuel_supply`, `fuel_export`, `other_uses`, `notional_transfer`, `allocation_agreement`) plus a couple of children of `compliance_report` itself, filtering by `compliance_report_id` and walking version chains via `group_uuid`. Postgres doesn't auto-index FKs, and only `allocation_agreement.group_uuid` was indexed before this PR — so each lookup was a sequential scan.

New alembic revision `a7b3c9d5e1f2` adds 12 indexes:

| Table | Column(s) |
| --- | --- |
| `fuel_supply` | `compliance_report_id`, `group_uuid` |
| `fuel_export` | `compliance_report_id`, `group_uuid` |
| `other_uses` | `compliance_report_id`, `group_uuid` |
| `notional_transfer` | `compliance_report_id`, `group_uuid` |
| `allocation_agreement` | `compliance_report_id` (group_uuid done in `173179aeed95`) |
| `compliance_report` | `compliance_report_group_uuid` |
| `compliance_report_history` | `compliance_report_id` |
| `compliance_report_summary` | `compliance_report_id` |

All statements use `CREATE INDEX IF NOT EXISTS` / `DROP INDEX IF EXISTS`, so the migration is safe to re-run and won't collide with any ad-hoc indexes added out-of-band.

### 2. Disable retry on the submit mutation + add a 90s client timeout (frontend)

`useUpdateComplianceReport` was inheriting the default react-query config (3 retries, 1s/2s/4s exponential backoff). The backend `handle_submitted_status` path serialises concurrent submits with a row lock and rejects the second attempt with a 409 ("This report has already been submitted") — by design, see #4368. When a PUT was slow but eventually succeeded, the silent retry raced its tail and got that 409, which the UI surfaced as a network error toast even though the submit had succeeded. This is the exact "appears not to submit initially but eventually processes" symptom from the ticket.

- `retry: 0` on the submit mutation (read queries are unaffected).
- 90s default axios `timeout` (was `0` = wait forever) — slow requests now fail with a clean client-side error before the infra-layer 504 hits. Caller-supplied opts still override.

### Out of scope (deferred follow-ups)

Two larger wins from the same investigation, intentionally split out:

- **Move synchronous email dispatch off the submit request path** ([notification/services.py:389](https://github.com/bcgov/lcfs/blob/develop/backend/lcfs/web/api/notification/services.py#L389)). Biggest single perceived-latency win; needs its own PR because it touches the notification service contract.
- **Push fossil/renewable filters into SQL inside `calculate_compliance_report_summary`** ([summary_service.py:779-808](https://github.com/bcgov/lcfs/blob/develop/backend/lcfs/web/api/compliance_report/summary_service.py#L779)). Eliminates the in-memory filter loop over the whole effective schedule set. Higher risk — better done after the indexes above are live so query plans are sane to begin with.

## Test plan

- [x] `alembic heads` → `a7b3c9d5e1f2 (head)`, chain intact
- [x] Migration syntax + idempotency (re-run safe via IF NOT EXISTS)
- [x] Frontend vitest: `useApiService.test.js` 28/28, `useComplianceReports.test.js` 12/12, `useComplianceReports.analyst.test.jsx` 13/13
- [ ] Apply migration in dev: `alembic upgrade head`, confirm all 12 indexes appear in `pg_indexes`
- [ ] Reproduce ticket flow in test: create + submit compliance report, confirm no "network error" toast on slow response, no duplicate Reserved transaction
- [ ] Sanity check `EXPLAIN` on `get_effective_fuel_supplies` and `get_compliance_report_chain` queries — should now use index scans, not seq scans

### DIFF
--- a/backend/lcfs/db/migrations/versions/2026-05-28-14-00_a7b3c9d5e1f2.py
+++ b/backend/lcfs/db/migrations/versions/2026-05-28-14-00_a7b3c9d5e1f2.py
@@ -16,7 +16,7 @@ All statements use ``IF NOT EXISTS`` so the migration is safe to
 re-run, and any ad-hoc indexes added out-of-band won't collide.
 
 Revision ID: a7b3c9d5e1f2
-Revises: c2a4f6b8d9e1
+Revises: e60c34503169
 Create Date: 2026-05-28 14:00:00.000000
 
 """
@@ -25,7 +25,7 @@ from alembic import op
 
 
 revision = "a7b3c9d5e1f2"
-down_revision = "c2a4f6b8d9e1"
+down_revision = "e60c34503169"
 branch_labels = None
 depends_on = None
 
--- a/backend/lcfs/db/migrations/versions/2026-05-28-14-00_a7b3c9d5e1f2.py
+++ b/backend/lcfs/db/migrations/versions/2026-05-28-14-00_a7b3c9d5e1f2.py
@@ -1,0 +1,84 @@
+"""Add submission hot-path indexes on schedule tables.
+
+Compliance report submission re-fetches every versioned schedule
+(fuel_supply, fuel_export, other_uses, notional_transfer,
+allocation_agreement) by ``compliance_report_id`` and walks the
+group history via ``group_uuid``. Postgres does not auto-index FKs,
+and only ``allocation_agreement.group_uuid`` was indexed prior to
+this migration, so each lookup was a sequential scan over the
+table.
+
+We also index ``compliance_report.compliance_report_group_uuid`` —
+that column is the join key used to assemble a report's version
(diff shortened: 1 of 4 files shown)


hamed-valiollahi (approved):
Great improvement! 👍  Thank you!

PR #4467, opened 2026-05-28
fix: allocation agreement line edit & delete in supplemental reports (#4444)

### LINKED ISSUE #4444: LCFS - Fix allocation agreement line editing and deletion in supplemental report
**Describe the Bug**  
A production bug was identified in a supplemental report within the allocation agreement table. Two issues were found during testing:
1. User is unable to delete a line in the allocation section of the report. The trash can icon is not there.
2. When attempting to change the quantity on an allocation agreement line, the system throws an error and highlights "Determining carbon intensity" as the issue.

This bug fix should also include a production data patch to delete the duplicate allocation agreement line that currently exists in a production report in draft supplemental status.

**Expected Behaviour**  
BCeID users should be able to:
- delete an allocation agreement entry line from a supplemental report
- edit the quantity on an allocation agreement line without triggering an unrelated validation or error message
- save the updated supplemental report successfully after making either change

**Actual Behaviour**  
- Allocation agreement lines in the supplemental report cannot be deleted.
- Editing the quantity on an allocation agreement line causes an error.
- The error incorrectly points to "Determining carbon intensity".
- A duplicate allocation agreement line was added by the user in the affected production draft supplemental report and requires patching.

Image

**Implications**  
- BCeID users cannot fully correct allocation agreement data in a supplemental report.
- Users may be blocked from submitting accurate supplemental changes.
- The incorrect error message makes troubleshooting difficult and may lead users to believe the issue is related to carbon intensity rather than the allocation agreement line edit.

**Steps to Reproduce**  
1. Open the affected report in draft supplemental status.
2. Navigate to the allocation agreement table within the supplemental report.
3. Attempt to delete an allocation agreement entry line.
4. Observe that there is no ability to delete the line.
5. Attempt to edit the quantity field on an existing allocation agreement line.
6. Observe that an error appears and highlights "Determining carbon intensity".

**Acceptance Criteria**  
- [ ]  BCeID users can delete an allocation agreement entry line within a supplemental report.
- [ ] - BCeID users can change the quantity on an allocation agreement line without error.
- [ ] - The system does not incorrectly trigger or highlight "Determining carbon intensity" when the quantity is edited.
- [ ] - Supplemental reports save successfully after allocation agreement lines are edited or deleted.
- [ ] - The duplicate allocation agreement line in the affected production draft supplemental report is removed through a production patch.

**Additional context**  
- This issue was identified through testing.
- Screenshot of the test bug will be included.
- The fix should address both the application behaviour and the existing production data issue.
- Impacted organization in prod is organization ID 340

AlexZorkin:
## Summary

Fixes #4444. BCeID users could not edit the quantity on, or delete, a pre-existing allocation agreement line in a supplemental report. Editing quantity threw a validation error that incorrectly highlighted the **Determining carbon intensity** column.

## Root cause

The allocation agreement `/list-all` endpoint returns `provisionOfTheAct` as a nested object (`{ provisionOfTheActId, name }`), unlike the other schedules (e.g. fuel supply), whose list response returns it as a plain string. The grid's `valueGetter` displayed it correctly, but the raw nested object stayed on the row until that specific cell was edited.

So when a user edited the **quantity** on a pre-existing line — the common case in a supplemental report — the un-flattened nested object was sent to `/save`, where `provision_of_the_act` is typed as `str`. Pydantic rejected it and the resulting error, keyed to `provisionOfTheAct`, surfaced as a spurious **"Determining carbon intensity"** message (that column's label). The same un-flattened payload also broke delete/undo of pre-existing lines.

A previous fix flattened these nested objects **after** save (for display), but the request payload was still sent un-flattened.

## Fix

Flatten the nested reference fields (`provisionOfTheAct`, `fuelCode`, `fuelType`, `fuelCategory`) **at load time**, via a shared `flattenNestedFields` helper that the existing post-save path now also reuses. Pre-existing rows therefore submit valid payloads on quantity edits and on delete/undo.

## Testing

- Added focused unit tests for `flattenNestedFields` (nested → string flattening, id capture, string pass-through, no mutation).
- Full `AddEditAllocationAgreements` suite passes (30 passed, pre-existing skips unchanged).

## Acceptance criteria

- [x] Change quantity on an allocation agreement line without error
- [x] No spurious "Determining carbon intensity" highlight on quantity edit
- [x] Delete/undo of a pre-existing allocation agreement line works
- [x] Supplemental reports save successfully after edit/delete

## Out of scope

The production data patch to remove the duplicate allocation agreement line (draft supplemental) is **not** included here — it will be handled separately once the specific record is identified in production.

### DIFF
--- a/backend/lcfs/tests/allocation_agreement/test_allocation_agreement_views.py
+++ b/backend/lcfs/tests/allocation_agreement/test_allocation_agreement_views.py
@@ -78,6 +78,42 @@ async def test_save_allocation_agreement_create(
 
 
 @pytest.mark.anyio
+async def test_save_allocation_agreement_missing_provision_returns_422(
+    client: AsyncClient,
+    fastapi_app: FastAPI,
+    set_mock_user,
+    mock_allocation_agreement_service,
+    mock_allocation_agreement_validation,
+):
+    """Saving a row without provision_of_the_act must return a clean field
+    validation error (422) rather than crashing the carbon-intensity lookup
+    with a 500 (regression for #4444)."""
+    with patch(
+        "lcfs.web.api.allocation_agreement.views.ComplianceReportValidation.validate_organization_access"
+    ) as mock_validate_organization_access:
+        set_mock_user(fastapi_app, [RoleEnum.SUPPLIER, RoleEnum.COMPLIANCE_REPORTING])
+        url = fastapi_app.url_path_for("save_allocation_agreements_row")
+        payload = create_mock_schema({"provision_of_the_act": None}).model_dump()
+
+        mock_validate_organization_access.return_value = True
+
+        fastapi_app.dependency_overrides[AllocationAgreementServices] = (
+            lambda: mock_allocation_agreement_service
+        )
+        fastapi_app.dependency_overrides[ComplianceReportValidation] = (
+            lambda: mock_allocation_agreement_validation
+        )
+
+        response = await client.post(url, json=payload)
+
+        assert response.status_code == 422
+        data = response.json()
+        assert data["errors"][0]["fields"] == ["provisionOfTheAct"]
+        # The crashing service path must not be reached
+        mock_allocation_agreement_service.create_allocation_agreement.assert_not_called()
+
+
+@pytest.mark.anyio
 async def test_save_allocation_agreement_update(
     client: AsyncClient,
     fastapi_app: FastAPI,
--- a/backend/lcfs/tests/allocation_agreement/test_allocation_agreement_views.py
+++ b/backend/lcfs/tests/allocation_agreement/test_allocation_agreement_views.py
@@ -153,6 +153,42 @@ async def test_save_allocation_agreement_delete(
         assert data["message"] == "Allocation agreement deleted successfully"
 
 
+@pytest.mark.anyio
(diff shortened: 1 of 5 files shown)


AlexZorkin:
### Added a second fix (delete validation)

Testing surfaced a second, distinct 422 on delete:

```
{'loc': ('quantity',), 'msg': 'field required', 'type': 'value_error'}
```

**Cause:** the `/save` endpoint reuses `AllocationAgreementCreateSchema` for deletes, so its `check_quantity_required` / `check_fuel_code_required` before-validators run on delete requests too. Deleting an **incomplete** row — e.g. the duplicate line added without a quantity — fails required-field validation, blocking the delete. The previous guard `isinstance(values, DeleteAllocationAgreementResponseSchema)` was dead code (the validator's `values` is always the raw request dict, never that response type), so the skip never fired.

**Fix:** guard both before-validators on the `deleted` flag, so deletions bypass required-field checks (quantity is irrelevant when removing a row) while create/update validation is unchanged. Added a regression test covering deletion of a row with no quantity.

This is the root-cause backend complement to the frontend flattening fix; together they cover the full issue (edit quantity, and delete — including incomplete/duplicate rows).

PR #4484, opened 2026-05-29
fix: exclude Other fuel types from pre-2024 compliance reports

AlexZorkin:
## Summary

Fixes #4434. The fuel type options **"Other"** and **"Other diesel fuel"** were appearing in legacy compliance report dropdowns (years 2019–2023) where they should not. Both fuel types were introduced with the 2024 LCFS legislation and must only be selectable from the 2024 compliance year onward.

### Root cause
Both fuel types are seeded as `is_legacy=FALSE`, `fossil_derived=FALSE`, `renewable=TRUE`. The per-year filtering in each schedule's table-options query treats any `is_legacy=False` non-fossil fuel as a "renewable reused across eras" and lets it through for pre-2024 reports — nothing keyed these two specific 2024-era fuel types out of legacy years.

## Changes
A shared `LEGACY_EXCLUDED_FUEL_TYPES` constant is added and referenced from each schedule's pre-2024 filter branch, so the exclusion list lives in one place. Applied across all four affected schedules:

| Area | File | Change |
|---|---|---|
| Constant | `backend/lcfs/utils/constants.py` | `LEGACY_EXCLUDED_FUEL_TYPES = ["Other", "Other diesel fuel"]` |
| Fuel supply | `backend/lcfs/web/api/fuel_supply/repo.py` | Exclude both in pre-2024 branch |
| Fuel export | `backend/lcfs/web/api/fuel_export/repo.py` | Exclude both in pre-2024 branch |
| Other uses | `backend/lcfs/web/api/other_uses/repo.py` | Drops "Other diesel fuel" pre-2024 ("Other" was already excluded) |
| Allocation agreement | `backend/lcfs/web/api/fuel_code/repo.py` | Exclude both pre-2024 (formatter used only by allocation agreement) |

- **2024+ behavior is unchanged** — both options remain available for 2024 and later.
- **No frontend change** — the UI renders whatever the API returns.
- **No DB/migration change** — filtering is the fix; existing saved reports are untouched.

## Acceptance criteria
- [x] Fuel options match correct year-based lookup tables
- [x] Invalid options removed from legacy reports 2019–2023

## Tests
- Added a regression test that compiles the actual query and asserts the `NOT IN` exclusion is present for 2023 and absent for 2024 (the existing repo tests mock the DB and don't exercise SQL).
- Full affected suite (fuel_supply, fuel_export, other_uses, allocation_agreement, fuel_code): **322 passed**.

### Manual verification
With the dev stack running, hit the table-options endpoints and diff the `fuel_types` lists:
- `GET /api/fuel-supply/table-options?compliancePeriod=2023` → must NOT contain "Other" or "Other diesel fuel"; `...=2024` → contains both.
- Repeat for fuel-export, other-uses, and allocation-agreement endpoints.
- In the UI, open a 2023 compliance report → each schedule's Fuel type dropdown omits both options; open a 2024 report → both present.

### DIFF
--- a/backend/lcfs/tests/fuel_supply/test_fuel_supplies_repo.py
+++ b/backend/lcfs/tests/fuel_supply/test_fuel_supplies_repo.py
@@ -285,6 +285,38 @@ async def test_get_fuel_supply_table_options_ghgenius_provision(
 
 
 @pytest.mark.anyio
+async def test_get_fuel_supply_table_options_excludes_other_fuel_types_pre_2024(
+    fuel_supply_repo, mock_db_session
+):
+    """Legacy reports (<2024) must exclude the 2024-era "Other" fuel types,
+    while 2024+ reports keep them."""
+    captured = {}
+
+    async def mock_execute(query, *args, **kwargs):
+        captured["query"] = query
+        result = MagicMock()
+        result.all = MagicMock(return_value=[])
+        return result
+
+    mock_db_session.execute = mock_execute
+
+    def compiled_sql():
+        return str(
+            captured["query"].compile(compile_kwargs={"literal_binds": True})
+        )
+
+    # Pre-2024: exclusion clause for both "Other" fuel types must be present
+    await fuel_supply_repo.get_fuel_supply_table_options("2023")
+    legacy_sql = compiled_sql()
+    assert "Other diesel fuel" in legacy_sql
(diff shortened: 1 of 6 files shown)


hamed-valiollahi (approved):
Looks good to me 👍  Thank you

PR #4499, opened 2026-06-04
Fix: LCFS - Verification and analyst columns of ci application

prv-proton:
Fixes #4464, #4465

github workflow fixes:

- Always runs npm ci in frontend shard jobs.
- Verifies installed Vitest matches frontend/package.json.
- Clears .vitest-reports before each shard run.
- Uploads only blob-*.json, not the whole report directory.
- Ignores missing blob files for empty shards.
- Adds version-aware blob validation before merge.


hamed-valiollahi (approved):
Looks good to me 👍  Thank you

PR #4502, opened 2026-06-04
feat: add Aggregator Issuance transaction type (#4492)

AlexZorkin:
## Summary

Adds a first-class **Aggregator Issuance** transaction type so credit issuances that were created for aggregator-related historical data entry surface as their own type instead of being lumped into the generic **"Legacy Transaction"** (`StandaloneTransaction`) bucket — and lets their compliance period / agreement date be corrected to the right year.

Relates to #4492.

## What's in this PR (structural only)

- **New `aggregator_issuance` table + model**, mirroring `initiative_agreement` / `admin_adjustment` (links a `transaction` row, carries `transaction_effective_date`, optional `recorded_date` / `gov_comment`).
- **Materialized views** (`mv_transaction_aggregate`, `mv_credit_ledger`) gain an `AggregatorIssuance` branch; rows that have an `aggregator_issuance` record are excluded from the standalone/legacy bucket, so genuine migrated legacy records keep their label.
- **Refresh trigger** on `aggregator_issuance` reusing `refresh_transaction_aggregate()`, matching the existing IA/AA triggers, so both MVs stay current automatically.
- **`AG` id prefix** (backend + frontend) and an **"Aggregator Issuance"** option in the Transactions grid type filter. The display label renders via the existing camelCase→spaces formatter.
- Reversible Alembic migration (`d2e4f6a8b0c1`).

Because the MV-derived compliance period comes from the agreement date's year, correcting the date into the right year corrects the period in one step. Credit `compliance_units` / `transaction_action` are never modified, so each org's **total** balance is unchanged — only the per-compliance-year bucket moves (the intended correction).

## Not in this PR

The record-level backfill of the affected transactions is applied **separately via manual SQL kept out of version control**, since it contains proprietary record identifiers.

## Verification

Full migration chain run against Postgres 14:
- `upgrade head` applies cleanly; a converted record surfaces as `AggregatorIssuance` with the corrected period, leaves the legacy bucket, and a control standalone record keeps its `StandaloneTransaction` label.
- Credit-ledger balance accumulates correctly (units preserved).
- `downgrade` drops the table + trigger and restores the prior view definitions; `re-upgrade` is clean.
- Backfill insert (transactional) auto-refreshes both MVs via the new trigger.


backend/lcfs/db/migrations/versions/2026-06-04-12-00_d2e4f6a8b0c1.py
```diff
@@ -0,0 +1,1290 @@
...
+        JOIN organization org
+            ON cr.organization_id = org.organization_id
+        JOIN compliance_report_status crs
+            ON cr.current_status_id = crs.compliance_report_status_id
+        JOIN compliance_period cp
+            ON cr.compliance_period_id = cp.compliance_period_id
+        JOIN "transaction" tr
+            ON cr.transaction_id = tr.transaction_id
+        AND cr.transaction_id IS NOT NULL
+        WHERE crs.status IN ('Assessed', 'Reassessed')
```
prv-proton, line 241: just a concern, if we are considering only Assessed and Reassessed transactions, then what about the transactions that are in "Reserve" status? would they be considered here or excluded? or They might fall into Standalone transactions.

backend/lcfs/db/migrations/versions/2026-06-04-12-00_d2e4f6a8b0c1.py
```diff
@@ -0,0 +1,1290 @@
...
+def _exec_script(ddl: str) -> None:
+    """Execute a multi-statement DDL script one statement at a time."""
+    for statement in ddl.split(";"):
+        statement = statement.strip()
+        if statement:
+            op.execute(statement + ";")
+
+
+def _create_aggregator_issuance_table() -> None:
+    op.create_table(
```
prv-proton, line 1164: It would be nice to include any missing indices on foreign keys or the key's that'll be used in joins like aggregator_issuance.transaction_id

prv-proton (approved):
Looks good 👍, Please consider suggestions and concerns highlighted in the comments.

backend/lcfs/db/migrations/versions/2026-06-04-12-00_d2e4f6a8b0c1.py
AlexZorkin: I updated the indexes to be consistent across all of these now.

backend/lcfs/db/migrations/versions/2026-06-04-12-00_d2e4f6a8b0c1.py
AlexZorkin: The transactions view currently is only supposed to show assessed transactions which is why we exclude in reserve here. we may want to clarify with PO

PR #4516, opened 2026-06-08
fix: restrict CI application page to CI Applicant role (#4504)

AlexZorkin:
## Summary

Restricts the CI application page (under the **Fuel codes** tab) so it is only accessible to users with the **CI Applicant** role, plus Government (IDIR) reviewers. Closes #4504.

Previously the page was reachable by BCeID users holding the **Signing Authority** role in addition to CI Applicants. This change removes Signing Authority from the frontend access guards.

CI-workflow participants who sign and submit are expected to also hold the CI Applicant role, so the backend submit guard (which requires Signing Authority) is unaffected — signers reach the page via their CI Applicant role and the submit check still passes.

## Changes

- **CIApplications** (list page): route guard `[ci_applicant, signing_authority, government]` → `[ci_applicant, government]`; "New CI application" button → `[ci_applicant]`
- **EditViewCIApplication** (add/edit page): route guard → `[ci_applicant, government]`
- **FuelCodesTabs**: the CI applications tab is shown only to CI Applicant or Government users (removed the now-unused signing-authority check)
- Updated the tab-visibility and route-redirect tests for the tightened access

## Acceptance criteria

- [x] Page accessible only to CI Applicant (and Government reviewers) — guards on both the list and add/edit pages
- [x] Non-CI-applicants blocked via navigation (tab hidden) and direct URL (redirect to dashboard)
- [x] Behaviour consistent across all entry points (nav tab, list page, add/edit page) and sessions
- [x] Aligns with the existing `withRole` / `Role` / tab RBAC patterns

## Testing

- `vitest run` on the three affected CI application test files — **38 tests pass**


hamed-valiollahi (approved):
Looks good to me 👍  Thank you

Issue #4458, opened 2026-05-26
LCFS - Comment Log filter bar, active chips, and client-side find


hamed-valiollahi:
**Description:**
Add the full filter bar to the Comment Log UI, wired to `useOrganizationComments` and URL query params. Render active filter chips with accessible remove buttons and a clear-all control. Add a client-side find input that highlights and navigates matches within the currently loaded results.

Controls (per wireframe):
- Keyword search input (calls server-side `search`)
- Category select (agreed list plus `All`)
- Compliance year select
- Date range picker
- Visibility segmented control (IDIR only)
- Active filter chips with remove buttons
- Clear filters button
- Result count
- Find input with next/previous buttons

**Wireframe:** https://preview.uxpin.com/7264e34749e7af420559b8cfc0ca8001c76db3ec#/pages/223328163/simulate/no-panels

**Purpose and benefit to user:**
Lets users quickly narrow large comment histories and locate specific text within the visible results without re-querying the server on every keystroke.

**Acceptance Criteria:**
- [ ] Given I select a category, When results update, Then an active chip appears and the URL reflects the selection.
- [ ] Given I click remove on a chip, When the chip is removed, Then the filter is cleared and results refresh.
- [ ] Given I click `Clear filters`, When the action completes, Then all filter params are removed from the URL and results reset.
- [ ] Given I type in the find input, When matches exist in the loaded results, Then matches are highlighted and I can navigate via next/previous with a visible current-match indicator.
- [ ] Given I am a BCeID user, When the filter bar renders, Then the visibility segmented control is not visible.
- [ ] Given I use only the keyboard, When I interact with the filter bar, chips, and find controls, Then all controls are reachable and operable.

**Development Checklist:**
- [ ] Category select bound to `category` param
- [ ] Compliance year select bound to `compliance_year` param
- [ ] Date range picker bound to `date_from`/`date_to`
- [ ] Visibility segmented control (IDIR only) bound to `visibility`
- [ ] Active filter chips with accessible remove labels (e.g. `Remove Year: 2025 filter`)
- [ ] Clear-all filters button
- [ ] Result count display
- [ ] Client-side find with next/previous navigation
- [ ] Highlight matches in `plainTextComment` only (not raw HTML)
- [ ] Real form labels on all inputs (no placeholder-only labels)
- [ ] axe scan on filter bar and chip interactions

**Notes:**
- Find is client-side only and operates on the loaded page of results.
- Do not inject highlight markup into the rich-text comment body.

Issue #4457, opened 2026-05-26
LCFS - Add PostgreSQL full-text search to the Comment Log endpoint


hamed-valiollahi:
**Description:**
Add keyword search to `GET /organizations/{id}/comments` via PostgreSQL full-text search over the `comment_search_vector` column. Apply `websearch_to_tsquery('english', :search)` when the `search` parameter is non-empty.

Category-specific search expansion:
- `category=Organization details` → also match against organization name fields.
- `category=Person` → also match against person-related fields (name, email).

**Wireframe:** https://preview.uxpin.com/7264e34749e7af420559b8cfc0ca8001c76db3ec#/pages/223328163/simulate/no-panels

**Purpose and benefit to user:**
Users can find specific comments by keyword across an organization's entire history, which is impractical via per-entity browsing.

**Acceptance Criteria:**
- [ ] Given I pass `search=transfer`, When the endpoint responds, Then only comments whose `comment_search_vector` matches the term are returned.
- [ ] Given I pass `search` with special characters or Unicode, When the endpoint responds, Then no `500` error occurs and results are sensible.
- [ ] Given `category=Organization details` and a `search` term, When the endpoint responds, Then results also include comments linked to organizations whose name matches the term.
- [ ] Given `search` is empty or omitted, When the endpoint responds, Then all comments matching other filters are returned.
- [ ] Given a `search` query against a populated dataset, When `EXPLAIN ANALYZE` is run, Then the GIN index is used.

**Development Checklist:**
- [ ] Add `search` parameter to the query schema
- [ ] Apply `websearch_to_tsquery` in the repository
- [ ] Implement category-specific expansion for `Organization details` and `Person`
- [ ] Sanitize `search` input at the API boundary
- [ ] Tests for keyword match, empty search, special characters, and Unicode
- [ ] Confirm GIN index usage with `EXPLAIN ANALYZE`

**Notes:**
- No semantic/embedding search in this ticket; full-text search is sufficient for the acceptance criteria.

Issue #4456, opened 2026-05-26
LCFS - Build the Comment Log read-only result list


hamed-valiollahi:
**Description:**
Build the read-only Comment Log list inside the new `CommentLog` component. Each row shows: author full name, create date, edited indicator (with update date), visibility badge (`Internal` / `external`), category, compliance year, entity context label, and the comment body rendered from sanitized HTML.

The list must be fully keyboard-accessible and meet WCAG 2.2.

**Wireframe:** https://preview.uxpin.com/7264e34749e7af420559b8cfc0ca8001c76db3ec#/pages/223328163/simulate/no-panels

**Purpose and benefit to user:**
Users can read every comment thread for an organization in one place, rather than navigating into each entity individually.

**Acceptance Criteria:**
- [ ] Given the endpoint returns results, When the page renders, Then each row shows author, date, visibility badge, category, compliance year, entity context, and comment body.
- [ ] Given results span multiple pages, When I use the pagination controls, Then the correct subset is displayed.
- [ ] Given I am a BCeID user, When the page renders, Then no `Internal` badge appears.
- [ ] Given a comment contains HTML, When it renders, Then the HTML is sanitized (no script execution).
- [ ] Given I use only the keyboard, When I navigate rows and pagination, Then focus is visible and order is logical with no scroll trap.
- [ ] Given an axe scan is run, When the page is loaded, Then no critical or serious accessibility violations are reported.

**Development Checklist:**
- [ ] Build comment row component
- [ ] Sanitize HTML via DOMPurify before render
- [ ] Reuse visibility badge component (or build one consistent with existing UI)
- [ ] Pagination control
- [ ] Loading skeleton and empty state
- [ ] Error state with retry
- [ ] Keyboard tab order and focus styles validated
- [ ] Run axe scan and resolve violations
- [ ] Contrast check for badges, metadata text, and edited indicator

**Notes:**
- Do not inject highlight markup into raw HTML. Highlights use `plainTextComment` (Phase 2).
- Edited indicator must not be hover-only.

Issue #4455, opened 2026-05-26
LCFS - Add Comment Log organization tab


hamed-valiollahi:
**Description:**
Add a new `Comment log` tab on the organization view with its own route. Add the `useOrganizationComments` React Query hook that calls `GET /organizations/{id}/comments` and keeps filter state in URL query params so views are shareable and browser navigation behaves as expected.

Relevant files:
- `frontend/src/routes/routes`
- `frontend/src/routes/routeConfig/organizationRoutes.tsx`
- `frontend/src/views/Organizations/OrganizationView/OrganizationView.jsx`

New component:
```
frontend/src/views/Organizations/OrganizationView/components/CommentLog/
```

**Wireframe:** https://preview.uxpin.com/7264e34749e7af420559b8cfc0ca8001c76db3ec#/pages/223328163/simulate/no-panels

**Purpose and benefit to user:**
Gives users a bookmarkable entry point to an organization's full comment history.

**Acceptance Criteria:**
- [ ] Given I view an organization, When I click the `Comment log` tab, Then the route updates and the Comment Log component renders.
- [ ] Given I apply a filter, When the request fires, Then the URL query params reflect the active filter.
- [ ] Given I share a filtered URL, When another authorized user opens it, Then the same filters are pre-applied.
- [ ] Given I am a BCeID user viewing my organization, When the page loads, Then the `Comment log` tab is visible.

**Development Checklist:**
- [ ] Add route entry in `organizationRoutes.tsx`
- [ ] Add tab label and `org:tabs` translation key
- [ ] Scaffold `CommentLog` component directory
- [ ] Implement `useOrganizationComments` with React Query + URL param sync
- [ ] Loading and error states
- [ ] Route-level test for tab selection and render

**Notes:**
- URL-driven filter state from day one. Do not retrofit later.

PR #4519, opened 2026-06-09
Feat: Add Comment Log

hamed-valiollahi:
This PR introduces the Comment Log feature.

Closes #4451, #4452, #4453, #4454, #4455, #4456, #4457, #4458


backend/lcfs/web/api/internal_comment/repo.py
```diff
@@ -344,3 +370,446 @@ async def get_internal_comment_ids_for_entity(
...
+            stmt = select(CIApplication.organization_id).where(
+                CIApplication.ci_application_id == entity_id
+            )
+        else:
+            return (None, None)
+
+        org_id = (await self.db.execute(stmt)).scalar_one_or_none()
+        return (org_id, None)
+
+    def _entity_meta_subquery(self):
```
AlexZorkin: The org-comments query is doing a lot of work per request. The entity-meta subquery runs a DISTINCT ON across the whole internal_comment table with no org filter pushed in — we only filter by org in the outer query. On top of that we're running three separate statements per call (two counts + the data query). We should push the org predicate into the subquery and drop the redundant row_total count. This could become a scaling problem as comments grow.

frontend/src/views/Organizations/OrganizationView/components/CommentLog/CommentLogFilters.jsx
```diff
@@ -0,0 +1,228 @@
...
+            <BCTypography variant="body2" sx={{ fontWeight: 500 }}>
+              {t('internalComment:log.filters.searchLabel')}:
+            </BCTypography>
+            <TextField
+              id={searchId}
+              data-test="comment-log-search"
+              size="small"
+              placeholder={t('internalComment:log.filters.searchPlaceholder')}
+              value={filters.search}
+              onChange={(e) => setFilter('search', e.target.value)}
```
AlexZorkin: Search hits the backend on every keystroke. The search field writes straight to the URL params/query key on each onChange, so we're firing the full-text query per character. Let's debounce it (~300ms) before committing to the filter.

backend/lcfs/web/api/internal_comment/repo.py
```diff
@@ -344,3 +370,446 @@ async def get_internal_comment_ids_for_entity(
...
+                CIApplicationInternalComment,
+                CIApplicationInternalComment.internal_comment_id
+                == InternalComment.internal_comment_id,
+            )
+            .outerjoin(
+                CIApplication,
+                CIApplication.ci_application_id
+                == CIApplicationInternalComment.ci_application_id,
+            )
+            .distinct(InternalComment.internal_comment_id)
```
AlexZorkin, line 572: The DISTINCT ON needs a tiebreaker. Right now it only orders by internal_comment_id, so for a comment linked to multiple entities (compliance report chains via copy_internal_comments) the entity link/year we show is arbitrary and can flip between requests. Add a deterministic secondary sort.

frontend/src/hooks/useOrganizationComments.ts
```diff
@@ -0,0 +1,309 @@
...
+      visibility
+    }: {
+      commentId: number
+      comment: string
+      visibility?: string
+    }) => {
+      const payload: Record<string, any> = { comment }
+      if (visibility) payload.visibility = visibility
+      const response = await client.put(
+        `/internal_comments/${commentId}`,
```
AlexZorkin: lets use an apiRoutes constant here instead of raw text.

AlexZorkin (changes requested):
Looking really good. Just a few query concerns. ty

AlexZorkin (approved):
Great updates 👍

PR #4533, opened 2026-06-10
chore: upgrade vitest to v4 and update tests for breaking changes (#4529)

AlexZorkin:
## Summary

Upgrades the frontend test stack from **vitest 3.2.4 to 4.1.8** (`vitest`, `@vitest/coverage-v8`, `@vitest/ui` — these must move together), resolving the **critical severity Dependabot alert** on vitest, and migrates the test suite for the v4 breaking changes.

Closes #4529. Supersedes #4518 (the automated bump, which failed 14+ of 32 frontend test shards because of the breaking changes below).

## Breaking changes handled

**1. Mocks follow real JS constructor semantics** ([migration guide](https://vitest.dev/guide/migration.html))
`vi.fn(() => obj)` / `.mockImplementation(() => obj)` can no longer be invoked with `new` — arrow functions aren't constructible. Converted constructor-style mocks to regular functions in:
- `Date` mocks — `BCPaginationActions`, `CreditCalculator`, `EditViewComplianceReport` tests
- `ResizeObserver` — `cellRenderers` tests
- Keycloak default export — `keycloak` and `logout-cache-clearing` tests (these previously crashed at module level)
- leaflet `Control` / `Icon` — `MapComponents` and FSE `utils` tests

**2. `vi.restoreAllMocks()` no longer resets `vi.fn()` module mocks** (spies only)
Per-test `mockImplementation`/`mockReturnValue` overrides leaked into subsequent tests. Switched the affected suites (`OrganizationDetailsCard`, `BCGridBase`) to `vi.resetAllMocks()`, which in v4 restores the factory implementations.

**3. Default `exclude` slimmed to `node_modules`/`.git`**
v4 no longer excludes `cypress/`/`dist/` by default, and this repo has `*.test.js` files under `cypress/e2e/` that vitest started collecting. Scoped `test.include` to `src/` in `vite.config.ts`.

**4. Nested `vi.mock()` deprecation**
Moved the `vi.mock()` calls in `testSetup.js` from inside `beforeAll` to top level — they were always hoisted anyway, and nested calls will become an error in a future vitest version.

**5. Timing fix**
One `BCGridBase` test relied on an empty `act()` to flush a `setTimeout(0)`-based `onGridReady`; it now awaits the timer explicitly.

## Verification

- Full local suite: **5,542 passed, 0 failed** (11 pre-existing skips), exit 0
- 43 more tests run than on `develop`'s baseline — tests in suites that previously crashed during collection under v4
- Coverage (`v8` provider, now AST-based remapping) verified working with the existing config
- CI shard/blob/merge pipeline audited: all flags used by the workflows (`--shard`, `--reporter=blob`, `--reporter=github-actions`, `--merge-reports`) still exist in v4, and shard + merge jobs both install from the lockfile so versions stay in sync
- Lockfile diff is vitest-ecosystem only (e.g. `vite-node` removed — v4 uses Vite's Module Runner)

### DIFF
--- a/frontend/package-lock.json
+++ b/frontend/package-lock.json
@@ -90,8 +90,8 @@
         "@types/react-dom": "^18.3.7",
         "@typescript-eslint/eslint-plugin": "^8.34.0",
         "@typescript-eslint/parser": "^8.34.0",
-        "@vitest/coverage-v8": "3.2.4",
-        "@vitest/ui": "3.2.4",
+        "@vitest/coverage-v8": "4.1.8",
+        "@vitest/ui": "4.1.8",
         "axe-core": "^4.10.3",
         "cross-env": "^7.0.3",
         "cypress": "^14.2.1",
@@ -121,7 +121,7 @@
         "storybook": "^8.6.17",
         "typescript": "^5.8.3",
         "vite-plugin-terminal": "^1.3.0",
-        "vitest": "3.2.4"
+        "vitest": "4.1.8"
       },
       "optionalDependencies": {
         "@rollup/rollup-linux-x64-gnu": "4.6.1"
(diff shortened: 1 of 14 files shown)


hamed-valiollahi (approved):
Looks great! Thanks for the upgrade. 👍

PR #4546, opened 2026-06-15
Feat: Add Release Notes Page - 4506

hamed-valiollahi:
This PR adds release notes generation support.

* Add release notes page
* Generate release notes via GitHub Actions
* Display historical release notes
* Support automated release note publishing

Closes #4506


frontend/src/components/Footer.jsx
```diff
@@ -53,10 +53,10 @@ const Footer = (props) => {
   )
   const repoDetails = useMemo(
     () => ({
-      href: 'https://github.com/bcgov/lcfs/releases/tag/v1.3.3',
-      name: 'v1.3.3',
-      id: 'footer-about-version',
-      label: 'LCFS repository changelog'
+      route: '/release-notes',
+      name: 'v1.3.3 — Release Notes',
```
AlexZorkin: I wondering if we can automate the increment of the version in this system as well instead of manually bumping the version?

AlexZorkin (approved):
This is a very cool feature, great addition. Approved, just one idea.

frontend/src/components/Footer.jsx
hamed-valiollahi: Great idea! It would be nice to automate that as well so we don't need to update it manually. I'll coordinate with Kuan on the Copilot part this week and make sure we have the required variable in place before merging. 👍

PR #4548, opened 2026-06-15
Fix: Reset comment type after submit - 4539

hamed-valiollahi:
This PR resets comment type to Internal after submitting a Public comment.

- Reset comment form after submit
- Default to Internal comment
- Apply to IDIR users

Closes #4539


AlexZorkin (approved):
👍

Issue #3598, opened 2025-12-10
LCFS - Update Summary Compare Mode to Disable Empty Fuel Categories


airinggov:
**Description:**
- Enhance the summary compare mode feature so that fuel category columns (Gasoline, Diesel, Jet Fuel) are dynamically enabled or disabled based on content availability.
- Current behavior defaults to Gasoline even if the supplier only supplied Diesel, requiring manual selection. New behavior should:
  - Automatically select the fuel category that has content.
  - Disable radio buttons for categories where Lines 3 or 9 have null or zero values.

**Wireframe:**

**Purpose and benefit to user:**
- Reduce user confusion and unnecessary clicks by automatically showing relevant data and disabling irrelevant options.

**Acceptance Criteria:**
- [ ] If Lines 3 or 9 for a fuel category are null or zero, disable its radio button.
- [ ] Automatically select the first fuel category with non-zero content.
- [ ] Disabled radio buttons are visually distinct and non-interactive.
- [ ] Behavior works consistently across all supported browsers.
- [ ] WCAG 2.2 AA compliance for disabled states and empty message.
- [ ] Unit tests for logic determining enabled/disabled states.
- [ ] Integration tests for UI behavior under different data conditions.

**Development Checklist:**
- [ ] Update backend logic to provide content availability flags for each fuel category.
- [ ] Modify frontend rendering to disable radio buttons based on flags.
- [ ] Implement auto-selection logic for first available category.
- [ ] Update accessibility attributes for disabled controls.
- [ ] Write unit tests for backend and frontend logic.
- [ ] Add integration tests for UI scenarios (single category, multiple categories, no categories).
- [ ] Update documentation and help text for new behavior.

**Notes:**
- Confirm which data source determines Lines 3 and 9 values.
- Validate performance impact for large datasets.
- Ensure audit logging captures which category was auto-selected.

PR #4549, opened 2026-06-15
fix: reconcile FSE kWh total and supplier export with the reporting table (#4540)

AlexZorkin:
## Summary

The **total kWh** shown above the FSE (Final Supply Equipment) table on a compliance
report did not match the sum of the FSE rows in the Excel export. The header total and
the table rows were produced by **two different queries** selecting different effective
row sets, so they disagreed on reports involving re-versioned, decommissioned-but-reported,
or supplemental-carried equipment.

Closes #4540.

## Root cause

The header total (`totalKwhUsage`) was computed by `get_total_kwh_usage_for_report_group`,
an independent aggregate that deduped by equipment `group_uuid` across the whole report
group. The table rows — and the government Excel export — come from
`get_fse_reporting_list_paginated(mode="summary")` reading the `v_fse_reporting_base_pref`
view, scoped to a single `compliance_report_id`. The two used different scoping, dedup,
`is_active`, and decommissioned-handling logic, so the displayed total could diverge from
the rows it summarized.

The supplier (BCeID) Excel export used a third query,
`get_effective_fse_reporting_rows_for_export`, which excluded all decommissioned equipment
(even already-reported), applied no `is_active` filter, and selected reporting rows by a
different priority — so a supplier download could also diverge from the on-screen total.

## Fix

- **Single source of filters.** Extract `_fse_reporting_base_conditions` so the reporting
  list and the total apply identical WHERE conditions on the view.
- **Header total.** Replace `get_total_kwh_usage_for_report_group` with
  `get_total_kwh_usage_for_report`, which sums `kwh_usage` over the same view rows and
  filters the table uses — so the total equals the sum of the rows by construction.
- **Supplier export.** Rewrite `get_effective_fse_reporting_rows_for_export` to take its
  row set + kWh + reporting data from the same view (summary filters), while still
  enriching equipment metadata (serial, manufacturer, model, level, ports, notes, intended
  uses/users) from the latest equipment version via `COALESCE`. Outer joins keep rows 1:1
  with the view, so the supplier Excel sum matches the total.

Result: the UI table, the header total, the government export, and the supplier export all
agree on the FSE kWh set and total.

**Out of scope (intentionally unchanged):**
- **Frontend** — the UI already renders the total and rows from one API response.
- **The `v_fse_reporting_base_pref` view / migrations** — the canonical source, unchanged.
- **Site/address columns** — already resolve to the latest site version in the view.

## Acceptance criteria

- [x] Header total equals the sum of the FSE rows shown in the table
- [x] Government Excel export sum equals the header total
- [x] Supplier (BCeID) Excel export sum equals the header total
- [x] Decommissioned-but-reported equipment is counted consistently everywhere
- [x] Latest-version equipment metadata preserved on supplier draft exports

## Testing

- Updated service tests for the new total method.
- Added repo tests asserting the total sums the same view + summary filters, and that the
  supplier export draws its rows/kWh from the same view + filters while enriching metadata
  via `COALESCE`.
- `backend/lcfs/tests/final_supply_equipment/` + `compliance_report/`: **719 passed**;
  CI-application and charging-equipment exporter suites also green.
- Tests assert query structure (mocked DB); runtime row-set equality on report #3598 is not
  yet verified against live data.

### DIFF
--- a/backend/lcfs/tests/final_supply_equipment/test_final_supply_equipment_repo.py
+++ b/backend/lcfs/tests/final_supply_equipment/test_final_supply_equipment_repo.py
@@ -528,6 +528,85 @@ async def test_get_fse_reporting_list_paginated(repo, fake_db):
 
 
 @pytest.mark.anyio
+async def test_get_total_kwh_usage_for_report_sums_same_view_and_filters(repo, fake_db):
+    """
+    The header total must sum kwh_usage over the exact same pref-view rows
+    (and identical filters) that the reporting list / Excel export returns, so
+    the displayed total cannot diverge from the rows it summarizes.
+    """
+    fake_db.scalar.return_value = 1234.5
+
+    result = await repo.get_total_kwh_usage_for_report(1, 10, "summary")
+
+    assert result == 1234.5
+
+    stmt = fake_db.scalar.call_args[0][0]
+    compiled = str(stmt.compile(compile_kwargs={"literal_binds": True}))
+    # Same source view as the list query, summing the kWh column.
+    assert "v_fse_reporting_base_pref" in compiled
+    assert "sum(v_fse_reporting_base_pref.kwh_usage)" in compiled
+    # Same scoping/filters as the list query, including the summary is_active
+    # filter so active-only rows are counted exactly as the table shows them.
+    assert "compliance_report_id = 10" in compiled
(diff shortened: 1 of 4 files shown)


prv-proton (approved):
Looks good 👍

PR #4550, opened 2026-06-15
Fix: Expand CNG Unit Widths - 4541

hamed-valiollahi:
This PR expands CNG unit column widths.
- Update fuel supply table
- Update allocation table
- Update export table

Closes #4541


AlexZorkin (approved):
👍

PR #4551, opened 2026-06-15
fix: CI application wizard refinements (#4531, #4532, #4534, #4536, #4537)

AlexZorkin:
Refinements to the Carbon Intensity (CI) application wizard, covering four tickets.

## #4531 — Step 1 (Application information)
- Relabel the field to **Facility nameplate capacity** (matches its validation copy).
- **Unit of measure** now lists physical capacity units (**L, kg, kWh, Gj, m³**) instead of energy-density units (MJ/L, gCO²e/MJ…). Root cause: the CI application stored the nameplate-capacity unit as an FK to the `unit_of_measure` table — whose own model comment is *"Units used to measure energy densities"*. A nameplate capacity is a physical quantity; fuel codes (and the allocation-agreement / other-uses schedules) already model it with `QuantityUnitsEnum`. Switched `CIApplication.facility_nameplate_capacity_unit_id` (int FK) → `facility_nameplate_capacity_unit` (`Enum(QuantityUnitsEnum)`), mirroring `FuelCode`.
  - **Migration `f3a8c1d2e4b5`** adds the enum column (reusing the existing `quantityunitsenum` Postgres type) and drops the FK column. Existing values referenced energy-density units that can't map to a physical unit, so they are left NULL (the field is optional).

## #4532 — Step 2 (Proposed fuel pathways)
- Removed the duplicate in-body section heading (the accordion already renders it).
- **Proposed CI** is now editable for both New and Renewal applications (it was locked on renewal).
- Fixed **"Add row" inserting two rows** and the resulting duplicate-on-save: the grid was driven by both a controlled `rowData` prop and ag-grid transactions. Made ag-grid the single source of truth for adds (matching the allocation-agreement grid).
- Fixed the **empty height gap after deleting a row**: delete now updates the controlled `rowData` (read from the live grid) so the grid, the prop, and the `autoHeight` calc stay in agreement.

## #4534 — Step 3 (Documents & GHGenius modelling)
- Removed the duplicate in-body heading.
- Removed the **Download GHGenius Excel template** button and the modelling input/output instructions (provided earlier in the process). Kept the Upload button. Pruned the now-unused locale keys and imports.

## #4536 — Step 4 (Sign & submit)
- Removed the duplicate in-body heading.
- **Submit application** is now disabled until the three required declarations are checked (the optional consultant-consent checkbox does not gate submission).

## Testing
- Backend: all CI application tests updated and passing (service, views, repo, list, validation).
- Frontend: full CI application suite passing (142 tests); step 1–4 test expectations updated.
- Migration applied and verified against a local dev DB (column type is now `quantityunitsenum`).

## Notes for reviewers
- This PR carries a **DB migration**; run migrations on deploy. Existing CI applications' nameplate-capacity unit is reset to NULL (the prior values were the wrong lookup).

---

## #4537 — Submitted application view (BCeID)
- BCeID/supplier users now keep the **full 5-step wizard progress bar** after submission instead of the short government workflow timeline (supplier steps 1-4 marked complete, Government decision pending). Government users still get the internal workflow timeline.
- The step 5 comments heading is now role-aware: government sees "Comments to the organization", BCeID sees **"Comments to the Government of B.C."**
- Added a progress-bar test for the BCeID-submitted case.

### DIFF
--- a/backend/lcfs/db/migrations/versions/2026-06-15-00-00_f3a8c1d2e4b5.py
+++ b/backend/lcfs/db/migrations/versions/2026-06-15-00-00_f3a8c1d2e4b5.py
@@ -1,0 +1,60 @@
+"""CI application facility nameplate capacity unit -> QuantityUnitsEnum
+
+Switches ci_application.facility_nameplate_capacity_unit_id (an FK to the
+energy-density `unit_of_measure` table — the wrong lookup, surfacing MJ/L,
+gCO²e/MJ, etc. in the form) to facility_nameplate_capacity_unit storing the
+QuantityUnitsEnum (L, kg, kWh, Gj, m³), matching how fuel codes model a
+facility nameplate capacity unit.
+
+Revision ID: f3a8c1d2e4b5
+Revises: e5f7a9b1c3d4
+Create Date: 2026-06-15 00:00:00.000000
+"""
+
+from alembic import op
+import sqlalchemy as sa
+from sqlalchemy.dialects import postgresql
+
+# revision identifiers, used by Alembic.
+revision = "f3a8c1d2e4b5"
+down_revision = "e5f7a9b1c3d4"
+branch_labels = None
+depends_on = None
+
+
+def upgrade() -> None:
+    # Reuse the existing `quantityunitsenum` Postgres type (created for fuel
+    # codes); do not recreate it.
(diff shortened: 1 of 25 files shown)


prv-proton (approved):
Looks good 👍

PR #4569, opened 2026-06-18
fix: sum Line 14 issued units over the transaction window (#4556)

AlexZorkin:
## Summary

Fixes #4556. Line 14 (compliance units issued under Initiative Agreements) was summed over the fixed compliance period (Jan 1 – Dec 31) instead of the Apr 1 → Mar 31 transaction window already used for Lines 12 & 13 (Jan 1 → Mar 31 for first/original reports). This change hands Line 14 the same `transaction_start_date`/`transaction_end_date` window so issuances are attributed to the correct reporting year by effective date.

## Scope

Lines 12 & 13 already use the correct window via `calculate_transaction_period_dates`. Line 14 was the only line still passing the calendar-year compliance period dates, so this is a one-line production change in `LowCarbonFuelTargetCalculator`.

## Impact

- Line 14 is a **display-only** summary line. It does not feed the Line 17 available balance or the Line 21 non-compliance penalty, both of which are derived independently from the transaction ledger. **No balance, penalty, or assessment outcome changes.**
- An impact review against current data identified **16 organization-year reports (24 versions)** whose Line 14 value changes; all carry a $0 penalty. First/original reports (Jan 1 start) are unaffected, since their window already begins January 1.
- Existing assessed reports will be corrected via a separate data patch.

## Testing

- Updated the summary-service assertion so Line 14 expects the transaction window.
- Added a regression test asserting Line 14 is summed over Apr 1 → Mar 31 (and not the compliance period).
- Ran the `compliance_report` and `transaction_windows` suites: **102 passed**.

### DIFF
--- a/backend/lcfs/tests/compliance_report/test_summary_service.py
+++ b/backend/lcfs/tests/compliance_report/test_summary_service.py
@@ -39,13 +39,13 @@ def _assert_repo_calls(
     Args:
         mock_repo: Mock summary repository
         mock_trxn_repo: Mock transaction repository
-        start_date: Compliance period start date (for Line 14)
-        end_date: Compliance period end date (for Line 14)
+        start_date: Compliance period start date (used for the Line 17 balance period year)
+        end_date: Compliance period end date
         organization_id: Organization ID
-        transaction_start_date: Transaction period start date (for Lines 12 & 13). If None, uses start_date.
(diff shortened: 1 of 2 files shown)


prv-proton (approved):
👍 , just for info, will we have transaction_start and end dates for all transactions including legacy ?

PR #4573, opened 2026-06-18
Feat: LCFS - Change log feature on the Fuel pathways-4389

prv-proton:
Closes #4389


AlexZorkin (changes requested):
Looks good, just will need a migration order fix.

Also add a direct test for the change-log diff logic _pathway_change_logs_from_versions.

AlexZorkin (approved):
looks good 👍

PR #4574, opened 2026-06-18
feat: add document attachments to internal comments (#4514)

AlexZorkin:
## Summary

Implements [#4514](https://github.com/bcgov/lcfs/issues/4514) — internal users can attach one or more documents to an internal comment. Attachments are stored via a new reference between the comment and document models, and are displayed with (and reload alongside) the comment.

The approach reuses the existing `Document`/S3 attachment machinery by introducing a new `internal_comment` parent type, rather than building a parallel system.

## Backend

- New `internal_comment_document_association` table + relationships on `InternalComment`/`Document`, with an Alembic migration off the current head.
- `DocumentService` handles the `internal_comment` parent type in upload/list/delete, with access that **matches comment permissions**: write (add/remove) = comment author only; read/download = government staff see all, others only Public comments.
- `/documents/{parent_type}/{parent_id}` routes validate the new parent type; the upload role gate is widened to include Government and CI Applicant so any comment author can attach.
- Comment GET responses now embed each comment's `documents` (single batched query, no N+1) so the list renders and reloads attachments in one round-trip.

## Frontend

- Paperclip button added to the comment editor (Quill) toolbar, with staged-file chips and the shared 50 MB / file-type validation.
- A new comment has no id until saved, so files are staged locally and uploaded after the comment is created; edit mode applies adds/removals on save.
- Saved attachments render as download links beneath each comment.

## Tests

- **Backend:** new coverage for the `internal_comment` parent type through the document routes, the access-control matrix, and embedded attachments in the comment list. Full document + S3 + compliance-report suites pass with no regressions; the migration applies via the test harness's `alembic upgrade head`.
- **Frontend:** new tests for the attach control, file staging, attachment download links, and feature gating. Existing Comments/editor tests pass.

## Notes

- The wireframe's `@`-mention icon is a separate feature and is intentionally out of scope here.
- `react-quill` exposes `Quill` as a static property of its default export (not a reliable named export), so it's accessed via `ReactQuill.Quill` to stay correct under production bundling.

## Acceptance criteria

- [x] Attachment option in the internal comment editor
- [x] Attach one or more documents to a single comment
- [x] Stored via a new comment↔document reference
- [x] Attachments remain linked after save and reload
- [x] Attachments displayed with the related comment
- [x] Add and delete attachments in both add and edit flows
- [x] Existing internal comment functionality unaffected


prv-proton (approved):
👍

PR #4575, opened 2026-06-18
Fix: Allow Editing Empty Rows in FSE Export - 4557

hamed-valiollahi:
This PR updates FSE export protection rules to support user calculations in editable cells.
- Unprotect empty FSE rows
- Allow edits in added rows

Closes #4557


AlexZorkin (approved):
👍

PR #4597, opened 2026-06-24
fix(comment-log): paginate organization comments by comment, not conversation

AlexZorkin:
## Summary

Fixes the organization **Comment Log** pagination count. The list paginated by *conversation* (entity group) but rendered every comment in the paged groups, so `pagination.total` counted conversations while the page showed comments. An entity with more than one comment produced a mismatched count — e.g. **"1 to 2 of 2"** displayed above **three** comment rows.

## Fix

Paginate by **comment**: `total` is the comment count and each page is a slice of comments. Ordering still clusters a conversation's comments together (by entity) with the most active conversations first, so threads stay visually grouped within a page. A long thread may straddle a page boundary, which the frontend renders as adjacent partial groups.

This also simplifies the query (removes the separate conversation-group counting/paging subqueries).

## Tests

Existing org-comment tests use one comment per entity, so conversations == comments and they were blind to this. Adds a regression test with a 2-comment entity + a 1-comment entity and asserts `total == 3` (not 2) with all three rows returned. Full `internal_comment` suite passes.

## Note

Trade-off: pagination no longer guarantees a conversation stays entirely on one page. If keeping threads un-split across pages is preferred over a comment-accurate count, the alternative is to keep conversation paging and relabel the count as conversations — happy to switch if that's the desired behavior.


prv-proton (approved):
👍

PR #4610, opened 2026-06-25
fix(supply-history): stop organization fuel supply filters from returning 500 (#4601)

AlexZorkin:
## Summary

Fixes [#4601](https://github.com/bcgov/lcfs/issues/4601): applying **any** filter on the Organization Dashboard → **Supply History** tab returned a 500.

The endpoint runs two queries (analytics charts + paginated table). "Any filter → 500" turned out to be **three** distinct bugs in those queries:

1. **Duplicate join (the reported `DuplicateAliasError`).** The analytics query re-`join()`ed `fuel_type` / `fuel_category` / `provision_of_the_act` / `fuel_code` in the filter loop, on top of the base `outerjoin`s. SQLAlchemy inferred the second join off `fuel_code`, so Postgres rejected it with *"table name `fuel_type` specified more than once."* This fired for every filter field. Fixed by filtering with `WHERE` only (the base query already joins those tables — the table query was already doing this correctly).

2. **`ILIKE` on an enum.** `fuel_category.category` is a Postgres enum, so `ILIKE` cast the pattern to the enum type and failed. Fixed by casting the column to text first.

3. **Filtering a Python property.** `FuelCode.fuel_code` is a `@property` (`prefix + suffix`), not a column, so `.ilike()` on it isn't valid SQL. Fixed by joining the prefix table and matching `concat(prefix, fuel_suffix)`.

Bugs 2 and 3 also affected the paginated table query (they were masked by bug 1 dying first).

## Tests

The existing org fuel-supply tests mocked the DB session, so they could never catch a SQL-construction error — which is how this shipped. Adds a repo-level regression test that runs all four filter fields through **both** the analytics and paginated queries against the real test DB. Full `fuel_supply` suite passes (67).

## Out of scope

Sorting the table by the `fuel_code` column hits the same property issue, but the default sort is by year and filtering doesn't trigger a `fuel_code` sort, so it's unrelated to this ticket. Flagged for a possible follow-up.

### DIFF
--- a/backend/lcfs/tests/fuel_supply/test_fuel_supplies_repo.py
+++ b/backend/lcfs/tests/fuel_supply/test_fuel_supplies_repo.py
@@ -716,3 +716,43 @@ async def test_get_effective_fuel_supplies_empty_result(
 
     # Verify mock was called
     assert mock_db_session.execute.call_count > 0
+
+
+@pytest.mark.anyio
+async def test_get_organization_fuel_supply_analytics_filters_no_duplicate_join(
+    dbsession,
+):
+    """Regression (#4601): filtering the org fuel-supply analytics query must
+    not emit a duplicate, unaliased join.
+
+    The base query already outerjoins fuel_type / fuel_category /
+    provision_of_the_act / fuel_code; re-joining them in the filter loop made
+    Postgres raise ``DuplicateAliasError: table name ... specified more than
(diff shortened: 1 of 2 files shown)


hamed-valiollahi (approved):
Looks good to me 👍  Thanks for the fix

PR #4613, opened 2026-06-26
Deterministic Summary for compliance report

prv-proton:


AlexZorkin (changes requested):
This is an awesome feature. We will need a set of tests to cover all of this new logic.

AlexZorkin (commented):
Looks great, thanks for the tests. 👍

AlexZorkin (approved):
👍

PR #4615, opened 2026-06-26
Resume draft CI application on first incomplete step (#4588)

AlexZorkin:
Closes #4588.

When a BCeID applicant reopens a draft CI application, the wizard now opens on the first incomplete step instead of always starting at Step 1.

## Approach
The resume step is derived client-side from the saved draft (data the detail endpoint already returns) — a step counts as complete only when its required fields have been persisted:
- **Step 1:** facility country, positive nameplate capacity, and unit
- **Step 2:** at least one pathway
- **Step 3:** technical report + GHGenius model uploads

The first step missing those is opened; when Steps 1–3 are all saved, the applicant resumes on Step 4 (Sign & submit). Because the calculation reads persisted data, progress survives logout/login and is identical regardless of which user opens the draft. An explicit `?step=N` in the URL still wins, so deep links, browser back/forward, and free accordion navigation are unaffected.

No backend or schema change is required.

## Acceptance criteria
- Determines the most recent completed step ✓
- Opens directly on the next incomplete step ✓
- A step is complete only if required fields were saved ✓
- Free navigation between steps preserved ✓
- Opens on Step 1 when nothing is completed ✓
- Persists across sessions (logout/login) ✓
- Correct regardless of which user opens it ✓

## Tests
- New `ciResumeStep` unit tests (per-step completion + resume index)
- Two integration tests in `EditViewCIApplication` asserting the resumed accordion expands
- Full CarbonIntensity suite passes (155 tests)


prv-proton (changes requested):
Looks good 👍

prv-proton (approved):
👍 , tested working fine.

PR #4616, opened 2026-06-29
Fix: LCFS - Generated fuel code table

prv-proton:
Fixes #4599


frontend/src/constants/config.ts
```diff
@@ -135,9 +135,9 @@ export const CONFIG: AppConfig = {
     // CI applications is in active development — default ON for dev/test,
     // OFF in prod until the feature is signed off and the tenant repo flag
     // flips it on for production.
-    ciApplications:
-      window.lcfs_config.feature_flags.ciApplications ??
-      !isProductionEnvironment,
+    ciApplications: true,
```
AlexZorkin: are we hard coding this on purpose?

AlexZorkin (approved):
Looks good just one comment to review.

frontend/src/constants/config.ts
prv-proton: set it up for testing, will revert back.

PR #4627, opened 2026-07-02
Fix: LCFS - Fuel produced in Canada select list and for gasoline

prv-proton:
Closes #4563, #4564


hamed-valiollahi (approved):
Looks good to me 👍  Thank you

PR #4628, opened 2026-07-02
fix: prevent default background flash on login page

AlexZorkin:
## Problem

On first load, the login page briefly flashes the bundled default background image (`summer.webp`) before the selected (admin-configured) background loads in.

## Cause

`bgUrl` was initialized to the bundled default image, which is a webpack asset and paints instantly. The selected image is fetched from `/login-bg-images/active` and streamed from `/login-bg-images/{id}/stream`, so it only appears after the network round-trip — the default was always shown first, then replaced (the flash).

## Fix

- Initialize with **no** background image and a neutral fill color, so nothing is painted eagerly.
- **Preload** the active (or fallback) image off-screen via `new Image()` and only commit it to state once it has fully decoded — the swap is seamless.
- The bundled default is used **only** when no active image is configured or the fetch/stream fails, and it's preloaded the same way so it never flashes either.
- Added a `cancelled` guard so an unmount mid-fetch doesn't set state on an unmounted component.

## Trade-off

For the brief moment before the active image decodes, users see a neutral color instead of the default photo. That's intended — it eliminates the default-image flash. On a warm cache the image commits almost immediately.

## Files

- `frontend/src/components/Login.jsx`


hamed-valiollahi (approved):
Looks good to me 👍  Thank you

PR #4631, opened 2026-07-02
Feat: LCFS - Consolidate Credit Market Report email attachments

prv-proton:
Fixes #4584


backend/lcfs/db/migrations/versions/2026-07-02-12-00_b7c8d9e0f1a2.py
```diff
@@ -0,0 +1,49 @@
...
+Revision ID: b7c8d9e0f1a2
+Revises: f2c8a9d1b3e5
+Create Date: 2026-07-02 12:00:00.000000
+"""
+
+from alembic import op
+
+
+revision = "b7c8d9e0f1a2"
+down_revision = "f2c8a9d1b3e5"
```
AlexZorkin: **Blocker:** `down_revision = "f2c8a9d1b3e5"` forks the alembic chain — develop's head moved to `b1c2d3e4f5a6` on Jun 29 (CI application fuel code association). Merging as-is creates two heads and `alembic upgrade head` fails in every backend CI shard (same failure mode as #4530/#4573).

Please re-point to `b1c2d3e4f5a6`. Heads up: #4632 also chains off `f2c8a9d1b3e5` and will be re-pointed the same way — whichever PR merges second will need one more re-point onto the other's revision.

backend/lcfs/db/migrations/versions/2026-07-02-12-00_b7c8d9e0f1a2.py
```diff
@@ -0,0 +1,49 @@
...
+            WHERE name = '{NOTIFICATION_TYPE}'
+        );
+        """
+    )
+
+
+def downgrade() -> None:
+    op.execute(
+        f"""
+        DELETE FROM notification_type
```
AlexZorkin, line 52: Downgrade will hit an FK violation once any user subscribes to this type (`notification_channel_subscription.notification_type_id`, no cascade). Delete the subscriptions for this type before deleting the type row.

backend/lcfs/services/jobs/jobs.py
```diff
@@ -245,3 +250,47 @@ async def reindex_compliance_report_tables(app: FastAPI):
...
+
+async def send_monthly_credit_market_report(app: FastAPI):
+    """
+    Sends the monthly Metabase credit market dashboard report through CHES.
+    """
+    logger.info("Starting monthly credit market report job")
+    session_factory = app.state.db_session_factory
+
+    try:
+        report = MetabaseClient().fetch_credit_market_report()
```
AlexZorkin: **Event-loop blocking:** this coroutine runs on the app's event loop (AsyncIOExecutor), but `MetabaseClient` is synchronous `requests` — auth + one POST per dashcard, each with a 30s timeout. A slow Metabase stalls every API request on this worker for potentially minutes.

One-line fix: `report = await asyncio.to_thread(lambda: MetabaseClient().fetch_credit_market_report())` — or port the client to `httpx.AsyncClient` (already a dependency).

backend/lcfs/services/scheduler/scheduler.py
```diff
@@ -87,6 +93,41 @@ def start_scheduler(app: FastAPI):
                 "Added one-time startup job: 'reindex_compliance_report_tables_startup'"
             )
 
+        if settings.credit_market_report_enabled:
```
AlexZorkin, line 96: **Duplicate mass emails:** the scheduler starts in every uvicorn worker (`workers_count`, default 2) and every pod replica, each with its own `MemoryJobStore` — there's no cross-process coordination. When this flag is enabled, external subscribers receive the report once per worker × replica.

This is a pre-existing pattern for the other jobs, but this one emails the public. It needs a distributed guard (pg advisory lock / Redis lock, or leader-only scheduling) before `credit_market_report_enabled` is ever turned on.

backend/lcfs/web/api/email/templates/credit_market_report.html
```diff
@@ -0,0 +1,89 @@
...
+    </td>
+    {% endfor %}
+  </tr>
+</table>
+
+{% for section in chart_sections %}
+<div style="border:1px solid #e6e8ee;border-radius:8px;background:#ffffff;padding:14px 18px;margin:8px 0 0 0;box-shadow:0 1px 3px rgba(15,23,42,0.08);">
+  <div style="font-family:BC Sans,Aptos,sans-serif;color:#4f5a78;font-size:15px;font-weight:700;margin-bottom:12px;">{{ section.name }}</div>
+  {% if section.image_data_uri %}
+  <img src="{{ section.image_data_uri }}" alt="{{ section.name }}" width="100%" style="display:block;width:100%;max-width:980px;height:auto;border:0;" />
```
AlexZorkin, line 54: `data:image/png;base64` URIs in `` are stripped or blocked by Gmail, Outlook, and most major clients — the chart sections will render blank for the majority of recipients. Consider CID inline attachments via CHES, or drop the images in favour of the table fallback. Worth validating in real mail clients before enabling.

backend/lcfs/services/metabase/client.py
```diff
@@ -0,0 +1,1450 @@
...
+                table
+                for table in report.tables
+                if self._is_monthly_report_table(table) and table.rows
+            ),
+            None,
+        )
+        if not monthly_table:
+            return []
+
+        current_row = monthly_table.rows[0]
```
AlexZorkin: This assumes `rows[0]` is the current month, i.e. it depends on the dashboard's sort staying newest-first. If someone reorders the card, the KPI cards silently show a stale month. Safer to select the row with the max parsed month.

backend/lcfs/services/metabase/client.py
```diff
@@ -0,0 +1,1450 @@
...
+            if "volume" in lower_name
+            else chart.render_line_chart(table)
+        )
+        if not image:
+            return None
+        encoded = base64.b64encode(image).decode("ascii")
+        return f"data:image/png;base64,{encoded}"
+
+
+class SimplePngChart:
```
AlexZorkin: `SimplePngChart` (~450 lines of hand-rolled PNG encoding + a 3×5 bitmap font) only executes when Pillow is absent — and Pillow is currently only a transitive dependency (via weasyprint). If that ever changes, this untested fallback silently becomes the production renderer.

Suggest declaring `pillow` as a direct dependency in `pyproject.toml` and deleting the fallback.

backend/lcfs/settings.py
```diff
@@ -58,6 +58,21 @@ class Settings(BaseSettings):
...
+    credit_market_report_hour: int = 0
+    credit_market_report_minute: int = 0
+    credit_market_report_run_on_startup: bool = False
+    credit_market_report_notification_type: str = "PUBLIC__CREDIT_MARKET_MONTHLY_REPORT"
+
+    metabase_base_url: str = ""
+    metabase_api_key: str = ""
+    metabase_username: str = ""
+    metabase_password: str = ""
+    metabase_session_token: str = ""
```
AlexZorkin: nit: Metabase session tokens expire (~14 days), so a static `metabase_session_token` setting will go stale in any long-lived environment — the API key path is the durable option. Consider dropping this setting to avoid a config trap.

backend/lcfs/services/metabase/client.py
```diff
@@ -0,0 +1,1450 @@
...
+            points = self._points(values)
+            color = self.line_colors[index % len(self.line_colors)]
+            for start, end in zip(points, points[1:]):
+                self._line(canvas, start, end, color)
+            for point_index, point in enumerate(points):
+                self._circle(canvas, point[0], point[1], 4, color)
+                if self._should_label_point(point_index, len(points)):
+                    self._draw_centered_text_canvas(
+                        canvas,
+                        self._format_chart_number(values[point_index], currency=True),
```
AlexZorkin: nit: `currency=True` is hardcoded for line-chart point labels (here and in the area-chart variant below) — a non-price trend series would get spurious `CA$` prefixes.

AlexZorkin (commented):
Reviewed — the report/XLSX builder core is solid and well-tested, but there are a few things to address before this can merge / be enabled, plus one hard blocker on the migration chain.

**Must fix before merge:**
1. Migration `down_revision` forks the alembic chain (inline comment) — CI backend shards will fail on `alembic upgrade head`.

**Must fix before the job is ever enabled:**
2. Sync Metabase I/O on the event loop (inline).
3. Duplicate mass emails — scheduler runs per worker × per replica with `MemoryJobStore` (inline).
4. `data:` chart images are stripped by Gmail/Outlook (inline).

**Also worth addressing:** downgrade FK issue, `rows[0]` month assumption, the Pillow-fallback dead weight, and there is no retry/alerting — if Metabase or CHES is down at fire time the month's report silently skips (log-only). The column/card name matching in `CreditMarketReportBuilder` is heuristic — a dashboard rename silently empties the email; consider logging a warning when an expected table is missing.

**Checked and fine:** secrets are env-loaded with empty defaults and never logged; Jinja autoescape is on (no injection from Metabase strings); all HTTP calls have timeouts; recipients are deduped and BCC-only; migration upgrade is idempotent; global recipient scope is appropriate for public market data; i18n keys are complete across all four notification panels.

Test coverage is strong on the builder and job orchestration; the MetabaseClient HTTP/auth paths and the send function itself are untested.

AlexZorkin (changes requested):
Comments attached.

backend/lcfs/settings.py
prv-proton: Yes, using API key.

AlexZorkin (approved):
great updates, thank you. 👍

PR #4632, opened 2026-07-02
Feat: LCFS - LNG End-use new UCI values effective May 6, 2026 (#4571)

AlexZorkin:
Closes #4571

## What
The LNG End-use UCI values change effective May 6, 2026. This adds a second "after May 6, 2026" set of LNG marine End-use options carrying the new UCI values.

Because UCI (`additional_carbon_intensity`) and the End-use dropdown options (`energy_effectiveness_ratio`) are already scoped by **compliance year**, this is a **data-only migration** — the dropdown options and the CI calculation follow the per-year rows automatically, so no frontend or service code change is needed (in either the compliance-report fuel supply or the public calculator).

## Behaviour
| Reporting year | Options / values |
|---|---|
| ≤ 2025 | existing ("before") options + values only |
| 2026 | **both** the before and the after options |
| ≥ 2027 (through 2030) | new ("after") options + values only |

After UCI values (per your spec): general 22.2 · 51–75% 12.7 · 76–100% 7.1 · slip-general 5.5 · slip-26–75% 3.3 · slip-76–100% 2.9 · unknown/avg 22.2 · Unknown engine 22.2 · Other 0.

## Design notes
- The after rows are **new `end_use_type` rows** (required: UCI is unique per period+fuel+end-use, so 2026 needs distinct rows to hold both values). Their labels are the existing label text **verbatim** + `" (after May 6, 2026)"` (derived in SQL; existing labels are left untouched — no relabeling). EER is unchanged (copied from the matching before end-use).
- Also removes the stray, incomplete LNG rows previously seeded for 2027–2030 (EER without UCI), so those years show only the after set.

## Verification
Ran the full migration chain against a real Postgres and confirmed 2025 → before only, 2026 → both, 2027+ → after only, with the exact values above; downgrade + re-upgrade reverse cleanly. Added a regression test (`lcfs/tests/fuel_code/test_lng_uci_effective_dates.py`) covering all nine end-uses across the three year regimes — **3/3 passing** against a freshly migrated test DB.


prv-proton (approved):
👍 . Just a query, aren't we truncating the labels?

PR #4635, opened 2026-07-03
fix: include all FSE records in compliance report export (#4619)

AlexZorkin:
Fixes #4619

## Problem

Compliance report downloads were capped at 1,000 Fuel Supply Equipment (FSE) records. Reports with more than 1,000 FSE records (e.g. 1,300+) were silently truncated to the first 1,000 rows in the downloaded file, producing incomplete data.

## Cause

The government (IDIR) export path in `sheet_exporters/fse.py` fetched FSE records via `get_fse_reporting_list_paginated` with a **hardcoded** `size=1000`, so only the first page was ever included. (The supplier/BCeID path already fetched all rows via `get_effective_fse_reporting_rows_for_export`, so only the government export was affected.)

## Fix

- Fetch the first page to learn the true `total` (which the paginated query already returns).
- Re-fetch with `size=total` **only when** the report has more rows than the initial page, so nothing is capped.
- Uses the query's own count, so it stays self-consistent, and the second fetch happens only when truncation would actually occur.

## Verification

- Backend syntax-checked.
- Not exercised against a live stack (requires the backend running plus a report with >1,000 FSE records); verified by inspection.

## Files

- `backend/lcfs/web/api/compliance_report/sheet_exporters/fse.py`


prv-proton (approved):
👍

PR #4636, opened 2026-07-03
fix: roll over early issuance reporting year on Jan 1 (#4602)

AlexZorkin:
Fixes #4602

## Problem

The compliance year shown in the early issuance reporting labels did not roll over on January 1. After Jan 1 2026 it still displayed **2025** instead of 2026:

- "Enable early issuance reporting for {year}" (org add/edit form)
- "Early issuance reporting enabled for {year}" (org profile)

## Cause

Both labels interpolated `CURRENT_COMPLIANCE_YEAR`, a fixed legislation-based constant (`LEGISLATION_TRANSITION_YEAR + 1` = 2025). It never advances with the calendar.

## Fix

Early issuance tracks the **current calendar year** — the backend already does this via `EarlyIssuanceYear.get_current_compliance_year()` (`str(datetime.now().year)`).

- Added a matching frontend helper `getCurrentEarlyIssuanceYear()` that returns `new Date().getFullYear()`.
- Used it for the org add/edit form and profile labels so the year rolls over automatically.
- **Left `CURRENT_COMPLIANCE_YEAR` unchanged on purpose** — it also drives the fixed reporting year used elsewhere (e.g. the Lookup Table's default compliance period), which should stay 2025. Repurposing it would have shifted that dropdown incorrectly.
- Updated the `OrganizationProfile` test mock to expose the new helper.

## Verification

- `OrganizationProfile.test.jsx` — all 34 tests pass.

## Files

- `frontend/src/constants/common.ts`
- `frontend/src/views/Organizations/AddEditOrg/AddEditOrgForm.jsx`
- `frontend/src/views/Organizations/OrganizationView/OrganizationProfile.jsx`
- `frontend/src/views/Organizations/OrganizationView/__tests__/OrganizationProfile.test.jsx`


prv-proton (approved):
Looks good, just one concern regarding from when should we start considering current early issuance i.e., is it from 1st Jan or 1st April of the year?

PR #4640, opened 2026-07-06
fix: include zero-dollar Category D transfers in summary Lines 12 and 13 (#4556)

AlexZorkin:
## Description

Follow-up to #4569 / #4556: zero-dollar Category D transfers were missing from compliance report summary Lines 12 and 13.

### Root cause

Category D is assigned during review, before the director records the transfer. `director_record_transfer` only stamps `transaction_effective_date` in the auto-categorization branch, so transfers with a pre-assigned category are recorded with a NULL effective date. The Line 12/13 queries filter with `transaction_effective_date BETWEEN ...`, which never matches NULL — those transfers silently drop out of the totals. All recorded transfers with a NULL effective date in the dev snapshot are zero-dollar Category D, confirming this is the population affected.

### Changes

- **`transfer/services.py`** — stamp the Pacific record date in `director_record_transfer` when the category is pre-assigned too (guarded so an existing date is never overwritten), preventing new NULLs.
- **`compliance_report/summary_repo.py`** — the Line 12/13 queries now use `COALESCE(transaction_effective_date, <Pacific date of the Recorded transfer_history row>)`, the same fallback convention as `mv_transaction_aggregate`. Existing NULL-date transfers are included as soon as a summary recalculates — no data patch required for open reports.
- Timezone detail: `transfer_history.create_date` is `timestamptz`, so the fallback uses a single `AT TIME ZONE 'America/Vancouver'` conversion (the double UTC hop idiom is only correct for timezone-naive columns and shifts evening records to the next day).

### Scope note

Assessed/locked reports read stored summary values, so this corrects open and newly recalculated reports; patching already-assessed reports remains a separate data exercise, consistent with #4556.

### Testing

- New DB-backed repo tests: NULL-date transfer counted in Line 12 and Line 13 when recorded inside the window; Pacific boundary case (03:00 UTC Apr 1 = Mar 31 Pacific) excluded from an Apr 1 window; transfer with no Recorded history row excluded without error.
- New service tests: effective date stamped when category is pre-assigned; pre-existing effective date preserved.
- Full `compliance_report` suite (497) and transfer suite pass.


prv-proton (approved):
👍

PR #4670, opened 2026-07-09
Disable supplemental 30-day warning after first submission (#4630)

AlexZorkin:
## Summary

Fixes #4630. The 30-day supplemental submission notice now only applies before the *initial* submission of a supplemental report, and is permanently retired once the report has been submitted for the first time.

## Changes

- **Disable after first submission.** A `Submitted` entry in the report's history permanently suppresses the notice. Returning the report to Draft (which flips the same report row back to Draft while keeping its history) does not reactivate it, so any number of analyst return/resubmit cycles stay warning-free.
- **Creation-based countdown.** The deadline is now `creationDate + 30 days`, matching the notice text ("30 days from the date of creation"), instead of `updateDate + 30` which drifted on every edit. The detail payload has no report-level create date, but a Draft history record is written when a supplemental is created, so the earliest history entry supplies the creation date — no backend change required.
- **Covers analyst-requested supplementals.** The notice now applies to both `Supplier Supplemental` and `Government Initiated` (ministry-initiated, for the supplier to edit) supplementals; `Government Reassessment` (analyst adjustment) is excluded.
- **Removed dead duplicate notice.** A second draft-supplemental notice was dead code — it read a non-existent `createTimestamp` field and referenced an unimported `<Alert>`/`<AlertTitle>`, so it never rendered (and would have thrown if it had). Also removed a stray `console.log` that ran on every render.

## Acceptance criteria
- [x] Notice shown for newly created supplementals that have never been submitted
- [x] Notice shown for analyst-requested supplementals that have never been submitted
- [x] Permanently disabled once submitted by a BCeID user for the first time
- [x] Returning to Draft does not reactivate the notice
- [x] Countdown no longer calculated after the first submission
- [x] Multiple analyst return/resubmission cycles supported without the notice
- [x] Existing supplemental workflows and statuses unaffected

## Testing
- `vitest` on `EditViewComplianceReport` — 105 passed, including new cases: shows for supplier-created and `Government Initiated`, excluded for `Government Reassessment`, hidden after first submission, and stays hidden after return-to-Draft.
- The return-to-supplier flow (same report row, history retains `Submitted`) was confirmed against the backend `update_service`.


prv-proton (approved):
👍

PR #4675, opened 2026-07-10
Fix: Restore fuel code date filtering - 4650

hamed-valiollahi:
This PR fixes date filtering on the Fuel Codes index.

- Allow Effective Date filtering
- Allow Last Updated filtering

Closes #4650


AlexZorkin (approved):
👍

PR #4618, opened 2026-06-30
Feat: LCFS - Add Fuel codes card to BCeID dashboard for CI applicants (#4579)

AlexZorkin:
Closes #4579

## What
Adds a "Fuel codes" card to the BCeID dashboard for **CI applicant** users (the IDIR version already existed). The card surfaces the organization's carbon intensity applications and links into the CI application workflow, matching the wireframe.

## Card contents (per wireframe)
- **Carbon intensity application(s) in draft** — count → CI applications list
- **Carbon intensity application(s) submitted for government review** — count → CI applications list
- **View all carbon intensity applications** → CI applications list
- **Start a new application** → add CI application
- Empty state when nothing is in progress

## Backend
New org-scoped endpoint `GET /dashboard/org-fuel-code-counts`, gated to `CI_APPLICANT` / `SIGNING_AUTHORITY`, resolving the org from the caller. Counts the org's `CIApplication`s in `Draft` and `Submitted` status in a single query (mirrors `org-compliance-report-counts`). Counts come from `CIApplication.organization_id` (always set) rather than `FuelCode.organization_id` (nullable/often-NULL).

## Frontend
Role-gated card (`withRole([roles.ci_applicant])`, mirrors `OrgComplianceReportsCard`), `useOrgFuelCodeCounts` hook, `apiRoutes` entry, `Dashboard.jsx` placement under `<Role roles={[roles.ci_applicant]}>`, i18n strings, and a unit test; updated the `Dashboard` test's cards mock.

## Acceptance criteria
CI-applicant-only visibility · below existing cards · consistent counters/layout (reuses `BCWidgetCard`/`CountDisplay`) · org-filtered counts · navigation to the CI application pages · empty state · shared accessible card components.

## Tests
Frontend: card unit test + Dashboard suite — full dashboard suite passes (256). Backend: a views test (role gating + count) added; runs in CI.

PR #4676, opened 2026-07-10
Gate Fuel codes dashboard card behind ciApplications flag (#4579)

AlexZorkin:
## Summary

Follow-up to the merged #4579 Fuel codes dashboard card (PR #4618). Two changes:

1. **Feature-flag gating (primary).** The card was gated by the `ci_applicant` role only, so it rendered for CI-applicant users even where the CI applications feature is turned off — including prod. It's now wrapped in `withFeatureFlag(FEATURE_FLAGS.CI_APPLICATIONS)` (no redirect — it simply renders nothing when the flag is off), matching how the CI application routes are already gated. This keeps it off prod until the feature is released, while still visible on test where the flag is on.

2. **Layout fix (review nit).** The "View all" / "Start a new" links faked their indent with a fixed `pl` and had no leading element, so they sat at a different indent and row height than the counted rows above. They now render the same `CountDisplay` element, invisibly (`visibility: hidden` + `aria-hidden`), so indent and line spacing match by construction.

## Testing
- `vitest` on `OrgFuelCodeCard` — 12 passed, including a new test asserting the card renders nothing when `ciApplications` is off.
- Not browser-verified: the card sits behind a `ci_applicant` Keycloak login + the flag, so it isn't reachable in a quick preview. The flag-off/on behaviour is covered by unit tests.

### DIFF
--- a/frontend/src/views/Dashboard/components/cards/bceid/OrgFuelCodeCard.jsx
+++ b/frontend/src/views/Dashboard/components/cards/bceid/OrgFuelCodeCard.jsx
@@ -10,13 +10,18 @@ import { roles } from '@/constants/roles'
 import { ROUTES } from '@/routes/routes'
 import { useOrgFuelCodeCounts } from '@/hooks/useDashboard'
 
-const CountDisplay = ({ count }) => (
+// When `hidden`, the number is rendered invisibly but still occupies layout,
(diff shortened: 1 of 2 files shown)


hamed-valiollahi (approved):
Looks good to me! 👍  Thanks

PR #4684, opened 2026-07-13
Fix: Refine Organization Comment Log Search - 4629

hamed-valiollahi:
This PR allows consistent, case-insensitive partial-text searches across all Organization Comment Log records.

Closes #4629


AlexZorkin (approved):
👍

PR #4685, opened 2026-07-13
Fix: CI application reference number always shows CI#

AlexZorkin:
## Summary
The Reference number on the CI application summary sometimes displayed a fuel code number instead of the system-generated CI# increment.

## Root cause
In `ApplicationSummary.jsx`, `referenceNumber` preferred a referenced fuel code number and only fell back to `CI${ciApplicationId}`. Any application containing a Renewal pathway (which carries a `fuelCode`) rendered the fuel code number, producing the inconsistent behaviour reported.

## Fix
Always source the reference number from `CI${ciApplicationId}`, matching how the CI Applications grid already renders it (`_schema.tsx`). The `referencedFuelCode` variable is retained since it is still used to derive the previous fuel code expiry date.

## Testing
- Reference number now always shows the CI# on the summary step regardless of pathway type.
- No existing test asserts this value; the component is not unit-tested in isolation.

Fixes #4657


hamed-valiollahi (approved):
Looks good to me 👍

PR #4686, opened 2026-07-13
Fix: approve generated fuel codes when CI application is approved

AlexZorkin:
## Summary
Generated fuel codes created for a CI application were persisted in Draft status and never advanced, so they remained "Draft" even after the application was approved (Completed). This left their status inconsistent and skewed reporting.

## Root cause
Generated fuel codes are created as real `FuelCode` rows in Draft status, but `record_decision`'s `Completed` branch only transitioned the application itself — nothing advanced the associated fuel codes out of Draft.

## Fix
Added `_approve_generated_fuel_codes`, invoked when a CI application transitions to `Completed`. It advances each associated Draft fuel code to Approved and stamps `approval_date` to mirror the application's approval date. Codes already Approved are left untouched.

Notifications and version history are intentionally not triggered here, since these are system-generated codes auto-approved as part of the CI decision rather than the manual fuel-code review workflow.

The associated fuel codes are eager-loaded via `get_by_id` (used by `validate_access`), so the mutations are in-session and flushed by the `@service_handler` transaction.

## Note on the ticket wording
The ticket says "Analyst approves", but terminal approval (`Completed`) is Director-only in the workflow — Analysts recommend. The transition is placed on the actual approval step (Director → Completed). If the intent is to approve codes at the recommend step, that is a one-line move.

## Testing
- Added `test_step5_completed_approves_generated_fuel_codes`.
- `poetry run pytest lcfs/tests/ci_application/test_ci_application_service.py` — 21/21 relevant tests pass; existing approval/generation tests unchanged.

Fixes #4654


prv-proton (approved):
👍 , thanks

PR #4687, opened 2026-07-13
Fix: clearing a required generated fuel code field no longer 500s

AlexZorkin:
## Summary
Clearing the **Application date** in the CI generated fuels grid returned a 500 ("Blocked: server is in error state").

## Root cause
Clearing the field sends `applicationDate: null`. `update_generated_fuel_code` wrote it straight onto the `FuelCode` via `setattr`, and because `application_date` is a `NOT NULL` column, the flush raised an `IntegrityError` that surfaced as a generic 500.

## Fix
Guard `update_generated_fuel_code`: if an update would clear a **non-nullable** `FuelCode` column to null/empty, raise the project's `ValidationErrorException` so the grid receives a structured field-level validation error (422) and renders it inline, instead of hitting the database and 500ing.

- The check is derived from the model's own column nullability, so it also covers the same class of bug for the other required non-nullable fields (`edrms`, `company`, `feedstock`, etc.), not just the application date.
- Genuinely nullable required fields (`approval_date`, `effective_date`, `expiration_date`) still fall through to the existing save-and-flag path.
- Verified end-to-end: `@view_handler` re-raises `ValidationErrorException` to the 422 handler, and the grid's existing error handling reads the `{errors:[{fields,message}]}` shape to highlight the field.

## Design note
Showing a *cleared* required date as a saved-but-flagged row (the grid's warning UX) would require persisting `null`, which the `NOT NULL` column forbids without a schema migration. Rather than widen a shared core column, this treats required as required and returns a clean field-level error. If a migration to make `application_date` nullable is preferred instead, that's a straightforward alternative.

## Testing
- Added `test_update_generated_fuel_code_clearing_required_date_raises_validation` (asserts the cleared date raises a validation error and never reaches `repo.update`).
- `poetry run pytest lcfs/tests/ci_application/test_ci_application_service.py` — 116/116 mock-based service tests pass. (View tests requiring a live Postgres were not run in this environment.)

Fixes #4653


hamed-valiollahi (approved):
Looks good to me 👍

PR #4695, opened 2026-07-14
Fix: keep Request further documentation button enabled after modal cancel

AlexZorkin:
## Summary
After clicking "Request further documentation" and cancelling the modal, the button became disabled and stayed greyed out until a full page refresh.

## Root cause
The button latched a local `requestedDocumentation` flag to `true` the moment it was clicked (on modal open) and never reset it. The modal's Cancel/Close handlers live in the parent (`EditViewCIApplication.jsx`) and only close the modal — they never reset the child's state, so the button stayed disabled until the component remounted.

## Fix
Opening the modal is not a completed action, so the premature disable is dropped: the button is now `disabled={readOnly}` only. The modal overlay already prevents re-triggering while it is open, and cancelling now leaves the button clickable — matching the sibling workflow buttons, which disable only during in-flight mutations.

## Testing
- Updated the existing `GovernmentDecisionStep` test (which asserted the buggy disable-on-click) to assert the button stays enabled after opening the modal.
- `vitest run GovernmentDecisionStep.test.jsx` — 26/26 pass.

Fixes #4651


prv-proton (approved):
👍

PR #4696, opened 2026-07-14
Fix: BCeID pipeline reflects Completed status on final step

AlexZorkin:
## Summary
When a CI application is approved, the grid shows a green "Completed" chip, but the BCeID (applicant) pipeline left the final "Government decision" step pending — an inconsistent status between the two views.

## Root cause
The BCeID applicant pipeline capped `wizardActiveStep` at `CI_APPLICATION_STEPS.length - 1` for every non-Draft status, and the completion predicate is `index < wizardActiveStep`. So the final step (index 4, "Government decision") was never marked complete — including when the status is `Completed`. The IDIR timeline path already handled `Completed`; only the BCeID wizard path did not.

## Fix
When the application reaches its terminal `Completed` state, advance the wizard past the final step so "Government decision" is marked complete and the BCeID pipeline matches the grid's "Completed" status. Submitted/Recommended behaviour is unchanged; scoped to `Completed` (Withdrawn is left as-is, since no government decision is made there).

## Testing
- Added a test asserting the final step is complete for BCeID users once the application is `Completed`.
- `vitest run CIApplicationProgress.test.jsx` — 15/15 pass.

Fixes #4652


prv-proton (approved):
👍 , thanks.

PR #4697, opened 2026-07-15
Fix: View creation for annual fuel supply

prv-proton:


hamed-valiollahi (approved):
Looks good, thanks for the fix 👍

PR #4699, opened 2026-07-15
Fix: Improve pathway export column widths - 4661

hamed-valiollahi:
This PR improves default column widths in pathway Excel exports.

- Show column headers without resizing
- Minimize manual column adjustments

Closes #4661


AlexZorkin (approved):
Looks good ty 👍

PR #4705, opened 2026-07-15
Skip data-entry validation when deleting Fuels for other use rows (#4689)

AlexZorkin:
Closes #4689.

BCeID users could not delete rows from **Fuels for other use** in a draft supplemental report. The delete failed with:

> Unable to delete/undo row: required when Expected use is Other.

## Root cause
Deleting a row re-submits the stored row with `deleted: true` to `POST /other-uses/save`, whose body is typed as `OtherUsesCreateSchema`. That schema's `model_validator(mode="before")` runs **while FastAPI parses the request** — so it rejects the payload before the endpoint ever reaches its `if request_data.deleted:` branch (`views.py:151`).

Any row with expected use `Other` and no rationale was therefore impossible to delete. That state is reachable for rows saved before the rationale rule existed — and `OtherUsesSchema` already tolerates exactly those rows on read (see the pre-existing `test_response_schema_allows_legacy_other_without_rationale`), so the data model expects them to exist.

This matches the issue's own diagnosis: *"validation intended for data entry is preventing row deletion."*

## Fix
Skip the field-level rules when the payload is a delete.

Those rules only apply to data being *saved*. `delete_other_use` (`services.py:290-321`) reads only `group_uuid`, `other_uses_id` and `compliance_report_id`, and copies **every other field from the record it already has** — so the submitted field values are discarded regardless. Validating them gates the delete on data that is never used.

Saving is unchanged: the rules still apply to any non-delete payload.

## Acceptance criteria
- Users can delete rows in Fuels for other use of a draft supplemental report ✓
- Deleting bypasses validation rules only applicable to saving ✓
- No warning message when deleting a valid row ✓
- Row removed successfully ✓ (the existing delete path now runs)
- Calculations/validation still work after deletion ✓ (delete path itself untouched)
- Adding/editing rows unaffected ✓ (rules still enforced on save)

## Scope
Backend-only. The frontend already sends the delete correctly and does not block it client-side — `isOtherExpectedUseMissing` only runs in `onCellEditingStopped` (the save path), not in `onAction`. The 400 surfaced from the server.

The skip is deliberately scoped to `OtherUsesCreateSchema` rather than the shared validators in `schema_validators.py`, so `allocation_agreement` (the other consumer of `fuel_code_required_label`) is untouched.

## Tests
3 new cases in `test_other_uses_validation.py`:
- delete with `Other` + no rationale → accepted (fails without this fix)
- delete with a fuel-code-requiring provision + no fuel code → accepted
- save with `deleted=False` + `Other` + no rationale → still rejected (regression guard)

19 tests pass in `lcfs/tests/other_uses/`. The 8 `test_other_uses_view.py` tests error at **fixture setup** on a pre-existing seeding `ForeignKeyViolationError` unrelated to this change — untouched modules (e.g. `lcfs/tests/fuel_export/`) show the identical failure on develop.

## Note on verification
Verified via the test suite; reproducing the original symptom end-to-end needs the deployed environment plus BCeID auth and a draft supplemental containing a legacy `Other` row with no rationale.

## Formatting
`black` + `isort` (the documented backend convention) were run on the two touched files, which reorders a few pre-existing import lines. Functional change is confined to the validator.

### DIFF
--- a/backend/lcfs/tests/other_uses/test_other_uses_validation.py
+++ b/backend/lcfs/tests/other_uses/test_other_uses_validation.py
@@ -1,16 +1,19 @@
-import pytest
 from unittest.mock import MagicMock
+
+import pytest
 from fastapi import HTTPException, Request
 from fastapi.exceptions import RequestValidationError
 
 from lcfs.web.api.other_uses.schema import OtherUsesCreateSchema, OtherUsesSchema
 from lcfs.web.api.other_uses.validation import OtherUsesValidation
 
+
 @pytest.fixture
 def other_uses_validation():
     request = MagicMock(spec=Request)
     validation = OtherUsesValidation(request=request)
     return validation
+
 
 @pytest.mark.anyio
 async def test_validate_compliance_report_id_success(other_uses_validation):
@@ -31,7 +34,10 @@ async def test_validate_compliance_report_id_success(other_uses_validation):
         )
     ]
 
-    await validation.validate_compliance_report_id(compliance_report_id, other_uses_data)
+    await validation.validate_compliance_report_id(
+        compliance_report_id, other_uses_data
+    )
+
 
 @pytest.mark.anyio
 async def test_validate_compliance_report_id_failure(other_uses_validation):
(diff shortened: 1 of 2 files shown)


hamed-valiollahi (approved):
Looks good to me 👍

PR #4708, opened 2026-07-16
Fix: LCFS - Make city and province fields mandatory across forms

prv-proton:
Fixes #4647, #4642


hamed-valiollahi (approved):
Looks good to me 👍

PR #4709, opened 2026-07-16
Fix: Require new CI value for pathway renewal - 4668

hamed-valiollahi:
This PR requires applicants to enter a new CI value when renewing a pathway.
- Clears previously populated CI values
- Applies only to pathway renewals
- Keeps the CI field mandatory
- Enforces existing CI validation rules

Closes #4668


AlexZorkin (approved):
👍

PR #4713, opened 2026-07-16
Suppress supplemental 30-day list flag after first submission (#4630)

AlexZorkin:
## Purpose

Follow-up to #4630 / #4670. The 30-day supplemental warning has **two** surfaces, and the earlier fix only addressed one:

1. The **banner** on the compliance report page — retired after first submission by #4670. ✅
2. The **red flag icon in the IDIR compliance reporting grid** (Type column, tooltip _"Supplemental draft over 30 days old"_) — **not** addressed. ❌

The grid icon fired on any group whose latest version is a Draft supplemental older than 30 days, with no check for prior submission. A supplemental that a supplier submitted and an analyst then returned to Draft is back in Draft and still >30 days old, so the icon kept showing — which is what was reported on dev.

## Changes

Retire the grid flag once the supplemental has been submitted for the first time, matching the banner behaviour and the #4630 acceptance criteria (returning to Draft must not reactivate the warning; multiple return/resubmit cycles supported).

- **`repo.py`** — `get_reports_paginated` runs a single batched query to determine which report groups' _latest_ version has a `Submitted` history entry, and exposes `latest_supplemental_has_been_submitted` on each row. (Kept as a repo-level query rather than a `v_compliance_report` view change to limit blast radius, since that view also feeds two materialized views.)
- **`schema.py`** — add `latest_supplemental_has_been_submitted` to `ComplianceReportViewSchema`.
- **`_schema.tsx`** — hide the Type-column flag when the latest supplemental has been submitted before.
- **tests** — cover the returned-to-Draft-after-submission case.

## Testing

- Frontend: `_schema.test.jsx` passes (23 tests, incl. the new suppression case).
- Backend: the batched query validated against dev data — it flags exactly the returned-Draft-but-previously-submitted groups for suppression while leaving never-submitted drafts flagged.

## Acceptance criteria

- [x] Flag shown for supplementals that have never been submitted (>30 days old).
- [x] Flag permanently suppressed once submitted for the first time.
- [x] Returning to Draft does not reactivate the flag.
- [x] Multiple analyst return/resubmit cycles supported without the flag.

### DIFF
--- a/backend/lcfs/web/api/compliance_report/repo.py
+++ b/backend/lcfs/web/api/compliance_report/repo.py
@@ -483,6 +483,54 @@ class ComplianceReportRepository:
             self.db, query, offset, limit
         )
 
+        # Determine which groups' *latest* version has ever been Submitted. The
+        # IDIR "supplemental draft over 30 days old" flag only nudges the
+        # organization toward the *initial* submission of a supplemental; once
+        # it has been submitted the flag is permanently retired, even if the
+        # report is later returned to Draft for analyst-requested revisions
+        # (#4630 — matches the report-page banner suppression). A prior
+        # Submitted history entry on the latest version is the signal.
+        group_uuids = [report.compliance_report_group_uuid for report in query_result]
+        submitted_group_uuids: set = set()
+        if group_uuids:
+            latest_versions = (
+                select(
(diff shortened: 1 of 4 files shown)


hamed-valiollahi (approved):
Looks great! 👍  Thanks

PR #4715, opened 2026-07-16
Fix credit ledger transaction time: recreate mv_transaction_aggregate with raw UTC (#4634)

### LINKED ISSUE #4634: LCFS - Credit Ledger Displays Incorrect Time Zone for Transfer Transactions
**Describe the Bug**
Credit transfer transactions are displaying the incorrect time in the Organization Credit Ledger. Transfers #5517 and #5514 were recorded at approximately 11:00 AM PST, however the transaction entries shown in the credit ledger for both the buyer and seller organizations display a time of 6:00 PM PDT.

A recent fix was implemented to address date display issues in the credit ledger, but it appears the time component is still being converted or displayed incorrectly.

**Expected Behaviour**
- Transfer transactions should display the transaction time in PST.
- Transfers #5517 and #5514 should display as approximately 11:00 AM PST.
- Date and time values should be presented consistently across all credit ledger views.
- Buyer and seller organizations should see the same correct timestamp.

**Actual Behaviour**
- Transfers #5517 and #5514 appear as 6:00 PM PDT in the Organization Credit Ledger.
- The incorrect time is displayed in both the buyer and seller organization views.

**Implications**
- Creates confusion regarding when credit transfers actually occurred.
- Reduces confidence in the accuracy and auditability of ledger records.
- May result in users questioning whether the transaction was processed at the correct time.

**Steps to Reproduce**
1. Navigate to the Organization Credit Ledger for either organization involved in transfer #5517 or #5514.
2. Locate the transfer transaction entry.
3. Review the displayed transaction timestamp.
4. Compare the displayed time with the recorded transfer time.

**Acceptance Criteria**
- [ ] Credit transfer timestamps are displayed in PST.
- [ ] Transfers #5517 and #5514 display the correct transaction time.
- [ ] Date display functionality from the recent fix remains unchanged.
- [ ] Buyer and seller ledger views display identical and correct timestamps.
- [ ] Time zone handling is validated across all credit ledger transaction types.

**Additional Context**
- A recent fix corrected date display issues in the credit ledger (#4600 )
- This issue appears related only to the time component and may be caused by timezone conversion or formatting logic.
- Business expectation is that ledger transaction timestamps continue to display in PST.

AlexZorkin:
Closes #4634.

The Organization Credit Ledger displayed transfer timestamps offset by the Pacific–UTC gap. Transfers **#5517 / #5514**, recorded ~11:08 AM PDT, showed as 6:08 PM PDT.

## Why the earlier fix (#4600 / PR #4614) didn't take
Commit `c4e994b81` corrected `backend/lcfs/db/sql/materialized_views.sql` — but **nothing executes that file**. `prestart.sh` runs `manage_views.py`, which recreates only the `vw_*` views from `metabase.sql`; the `mv_*` materialized views are defined **solely by Alembic migrations**. The file's own header says as much. So no environment ever picked up the corrected definition, and a plain `REFRESH MATERIALIZED VIEW` just recomputes rows from the old, still-double-converting definition.

Confirmed live: `pg_get_viewdef('mv_transaction_aggregate')` still contains `AT TIME ZONE 'UTC' AT TIME ZONE 'America/Vancouver'` on `update_date`/`create_date`.

## Root cause
`mv_transaction_aggregate` builds the display columns with `AT TIME ZONE 'UTC' AT TIME ZONE 'America/Vancouver'`, which strips the zone and shifts the instant ~7h forward. The frontend (`timezoneFormatter`) then converts the already-shifted value to Pacific a second time.

| transfer #5517 | value |
|---|---|
| True instant | `2026-07-03 18:08 UTC` = **11:08 AM PDT** |
| MV stored (double-converted) | `2026-07-04 01:08 UTC` |
| Frontend renders | **6:08 PM PDT** ✗ |

This also explains why only the *time* looked wrong in #4634: the shift landed on the same Pacific calendar day, so the date appeared correct.

## Fix
Migration `2026-07-16-12-00_d4e5f6a7b8c9.py` recreates `mv_transaction_aggregate` (and its dependent `mv_credit_ledger`, which the ledger endpoint actually reads) with the corrected definition: display timestamps emit the **raw UTC `timestamptz`** and the frontend performs the single Pacific conversion. The `::date` / compliance-period extractions keep their Pacific conversion, so the #4600 date fix is preserved.

The new definition is lifted from `materialized_views.sql`; a diff against the live definition confirms the **only** change is the six `update_date`/`create_date` expression pairs — nothing else.

## Verification (against a live DB)
- `upgrade` → view has **0** timezone conversions on the timestamp columns; MV now emits the true instant (e.g. `2018-08-14 19:41` instead of the shifted `2018-08-15 02:41`).
- `::date` columns → **9** conversions preserved (date behavior unchanged).
- `downgrade` → old definition restored (reversible).
- `alembic heads` → single clean head chained off the current head.
- All 4 indexes (incl. the unique indexes used for concurrent refresh) recreated.

## Notes for reviewers
- **No frontend change needed** — `timezoneFormatter` already converts UTC→Pacific correctly; it just needs the DB to stop pre-shifting. Buyer and seller read the same underlying instant, so both now show the same correct time.
- **Cross-consumer:** the main Transactions grid (`TransactionView`) also maps to `mv_transaction_aggregate`. It formats `updateDate` with a UTC-date formatter, so this change moves it toward the true instant's date as well (e.g. `2018-08-15` → `2018-08-14`), consistent with #4614's intent. No regression.
- Recreating `mv_transaction_aggregate` cascades only to `mv_credit_ledger` (verified via `pg_depend`); both are recreated in the migration.

## Follow-up (not in this PR)
`materialized_views.sql` is a reference copy that no code path executes yet drifts from the migration-defined truth — it silently caused this bug to look "fixed." Worth either wiring it into a migration-backed recreate step or clearly quarantining it. Happy to file an issue.

### DIFF
--- a/backend/lcfs/db/migrations/versions/2026-07-16-12-00_d4e5f6a7b8c9.py
+++ b/backend/lcfs/db/migrations/versions/2026-07-16-12-00_d4e5f6a7b8c9.py
@@ -837,9 +837,22 @@ CREATE UNIQUE INDEX mv_credit_ledger_tx_org_idx ON mv_credit_ledger (transaction
 """
 
 
+def _exec_script(ddl: str) -> None:
+    """Execute a multi-statement DDL script one statement at a time.
+
+    The migrations run over asyncpg, which rejects multiple commands in a
+    single prepared statement, so each statement is sent on its own (matching
+    the helper in 2026-06-04-12-00_d2e4f6a8b0c1.py).
+    """
+    for statement in ddl.split(";"):
+        statement = statement.strip()
+        if statement:
+            op.execute(statement + ";")
+
+
 def upgrade() -> None:
-    op.execute(UPGRADE_SQL)
+    _exec_script(UPGRADE_SQL)
 
 
 def downgrade() -> None:
-    op.execute(DOWNGRADE_SQL)
+    _exec_script(DOWNGRADE_SQL)
--- a/backend/lcfs/db/migrations/versions/2026-07-16-12-00_d4e5f6a7b8c9.py
+++ b/backend/lcfs/db/migrations/versions/2026-07-16-12-00_d4e5f6a7b8c9.py
@@ -1,0 +1,845 @@
+"""Fix credit ledger timezone: emit raw UTC timestamps in mv_transaction_aggregate
+
+The credit ledger showed transfer timestamps offset by the Pacific-UTC gap
+(e.g. an 11:00 AM PST transfer rendered as 6:00 PM). mv_transaction_aggregate
+built update_date/create_date with
+`AT TIME ZONE 'UTC' AT TIME ZONE 'America/Vancouver'`, shifting the instant,
+and the frontend then converted the already-shifted value to Pacific a second
+time. Commit c4e994b81 corrected db/sql/materialized_views.sql, but that file
+is only a reference copy; nothing executes it, so no environment picked up the
+change. This migration recreates mv_transaction_aggregate (and its dependent
+mv_credit_ledger) with the corrected definition so the fix actually reaches the
+database. Display columns now emit the raw UTC timestamptz and the frontend
+performs the single Pacific conversion. The ::date / compliance-period
+extractions are intentionally unchanged.
+
+Revision ID: d4e5f6a7b8c9
+Revises: b7c8d9e0f1a2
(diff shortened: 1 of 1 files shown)


hamed-valiollahi (approved):
Nice fix! Thanks 👍

PR #4716, opened 2026-07-16
fix: redirect authenticated users from public landing into the app

AlexZorkin:
## Problem

Logging in via BCeID or IDIR redirects the user back to the public landing page instead of the dashboard. Clicking the login button a second time then takes them into the app.

## Root cause

The login buttons call `keycloak.login({ redirectUri: window.location.origin })`, so the OIDC callback returns to the app root `/`. `/` is a protected route wrapped in `RequireAuth`; `MainLayout` sends an unauthenticated root visit to the public landing page (`/public`). When the guard renders before the session is fully settled, it bounces the just-authenticated user to `/public` — which had **no logic to forward an already-authenticated user onward**, stranding them there until they click login again (instant now that the SSO session is live).

The recent public-landing redesign didn't create the bounce; it changed where it lands (`/login` → the full `/public` page), making it obvious.

## Fix

Forward authenticated users from both public entry points (`PublicDashboard` and `Login`) to the dashboard via a small effect:

```jsx
useEffect(() => {
  if (initialized && keycloak.authenticated && !currentUserError) {
    navigate(ROUTES.DASHBOARD, { replace: true })
  }
}, [initialized, keycloak.authenticated, currentUserError, navigate])
```

The `!currentUserError` guard matters: a valid IdP login with no LCFS account makes `RequireAuth` bounce to `/public` *with* an error, so without the guard we'd create an infinite `/` ↔ `/public` redirect loop. Those users stay put (current behaviour) while legitimate users pass straight through.

## Tests

Added unit coverage for the authenticated, current-user-error, and unauthenticated cases on both entry points. Full suite for the two files: 31/31 passing.

## Verification note

This is the real Keycloak → BCeID/IDIR redirect flow, which can't be exercised without the full stack + real IdP credentials, so it was verified via unit tests rather than a live login.

### DIFF
--- a/frontend/src/views/PublicDashboard/PublicDashboard.tsx
+++ b/frontend/src/views/PublicDashboard/PublicDashboard.tsx
@@ -1,4 +1,4 @@
-import { useMemo } from 'react'
+import { useEffect, useMemo } from 'react'
 import { Box, Button } from '@mui/material'
 import CalculateOutlinedIcon from '@mui/icons-material/CalculateOutlined'
 import TableChartOutlinedIcon from '@mui/icons-material/TableChartOutlined'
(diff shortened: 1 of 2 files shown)


hamed-valiollahi (approved):
Looks good to me 👍  Thanks for the fix

PR #4746, opened 2026-07-23
Move Company Overview to organization dashboard as searchable comment thread (#4608)

AlexZorkin:
## Summary

Closes #4608. Company Overview moves off its standalone organization subtab
and onto the **organization dashboard** (directly below Organization Details)
as a dedicated comment thread built on the existing internal-comment framework.
The notes are now **searchable in the organization Comment Log**, filterable by
the seeded `Company Overview` category.

The four legacy free-form fields (`company_details`,
`company_representation_agreements`, `company_acting_as_aggregator`,
`company_additional_notes`) are removed — the business area confirmed they were
unused.

## Backend

- New `ORGANIZATION` entity type + `OrganizationInternalComment` association
  table/model, wired through create, list, org/year derivation, and the Comment
  Log `_entity_meta_subquery` so org-scoped comments appear and are searchable.
- Migration `c9d1e2f3a4b5`: create the association table + index, seed the
  `Company Overview` category (idempotent), and drop the legacy `company_*`
  columns.
- Removed the `PUT /{organization_id}/company-overview` endpoint, its service,
  and schema. Org-scoped comment creation is government-only (BCeID is blocked).

## Frontend

- New `CompanyOverviewComments` dashboard section reusing `CommentForm`
  (create) and `CommentRow` (list + inline edit), shown for IDIR users below
  Organization Details.
- Added `useOrganizationCommentThread` / `useCreateOrganizationComment` hooks.
- Removed the Company Overview subtab, route, breadcrumb segment, old component,
  and the `useUpdateCompanyOverview` hook.

## Testing

- Backend: `internal_comment` (71 existing + 4 new org-comment tests) and
  `organizations` (123) suites pass; the new tests run through the real Alembic
  migration. Migration DDL additionally validated in a rollback transaction.
- Frontend: 103 tests across 7 suites pass (new component test + updated
  route/tab/dashboard-card suites). Prettier-clean.


backend/lcfs/db/migrations/versions/2026-07-22-10-00_c9d1e2f3a4b5.py
```diff
@@ -0,0 +1,156 @@
...
+            INSERT INTO comment_category (display_name, display_order)
+            VALUES (:name, :order)
+            ON CONFLICT (display_name) DO NOTHING
+            """
+        ).bindparams(name=COMPANY_OVERVIEW_CATEGORY, order=COMPANY_OVERVIEW_ORDER)
+    )
+
+    # 3. Drop the legacy free-form company_* columns ----------------------
+    for column_name, _comment in LEGACY_COMPANY_COLUMNS:
+        op.drop_column("organization", column_name)
```
prv-proton, line 278: Is it ok to drop these columns without taking a backup first? I noticed that some rows in the production database still have values populated in these fields.

prv-proton (approved):
Looks good to me👍 , Thanks. 

Just need to fix the migration order before merge. and few concerns below.

frontend/src/views/Organizations/OrganizationView/components/CompanyOverviewComments.jsx
```diff
@@ -0,0 +1,126 @@
...
+                variant="body2"
+                sx={{ mb: 2 }}
+                data-test="company-overview-empty"
+              >
+                {t('org:sections.companyOverview.empty')}
+              </BCTypography>
+            ) : (
+              <BCBox data-test="company-overview-list">
+                {comments.map((comment, index) => (
+                  <CommentRow
```
prv-proton: the styling seems to vary slightly for the commented rows in comparison to the standard styling used elsewhere in the app

backend/lcfs/db/migrations/versions/2026-07-22-10-00_c9d1e2f3a4b5.py
AlexZorkin: Good catch — updated in e033fd8aa. The migration now snapshots all populated rows into `organization_company_overview_backup` before dropping anything, and additionally converts each organization's values into a migrated "Company Overview" comment so the existing data remains visible in the new framework. Downgrade restores the columns from the backup. Verified with an upgrade/downgrade round-trip against a database with populated values (including HTML-unsafe and multi-line text).

frontend/src/views/Organizations/OrganizationView/components/CompanyOverviewComments.jsx
AlexZorkin: Fixed in e033fd8aa — the rows were already the shared CommentRow component, but the section wasn't using the Comment Log's bordered white container around them, which is what the row striping is designed against. It now uses the identical wrapper (bordered white panel + semantic list) and the same info-alert empty state as the Comment Log, so CommentRow renders identically in both places.

PR #4750, opened 2026-07-23
Feat: Add CI application location autocomplete - #4717

hamed-valiollahi:
This PR supports predictive location entry in CI Application Step 1.

- Suggest existing cities during Step 1
- Populate province/state and country automatically
- Allow manual entry for unmatched cities

Closes #4717


backend/lcfs/web/api/ci_application/repo.py
```diff
@@ -184,6 +184,102 @@ async def get_transport_modes(self) -> Sequence[TransportMode]:
         )
         return result.scalars().all()
 
+    @repo_handler
+    async def get_all_facility_locations(self) -> List[Dict[str, str]]:
```
AlexZorkin: Should this be filtering per organization? currently its getting all locations for all organizations rather than just the user's org that is querying.

AlexZorkin (changes requested):
Just one concern about the lookup. ty

backend/lcfs/web/api/ci_application/repo.py
hamed-valiollahi: Good point! Thanks for catching that. I should’ve checked with the PO before pushing the code. 👍

I spoke with the PO, and he suggested using all locations without filtering by organization. He also said using georeferenced data is more preferable if possible. That’s what this implementation now does by using BC Geocoder and Nominatim instead of the previous approach.

Thanks again!

AlexZorkin (approved):
Thanks for checking with the PO. solution looks good, just a few merge conflicts to address first.

PR #4754, opened 2026-07-24
Keep Verification 2 available for Medium risk applications (#4741)

AlexZorkin:
## Summary

Closes #4741. Verification 2 was gated on **High risk only**, so once
Verification 1 was completed on a Medium risk application the Verification 2
section disappeared — taking the **Risk Assessment** and **Priority Score**
with it, because those fields render inside the active verification panel.

Medium now follows the same verification path as High: Verification 2 remains
available after Verification 1, and fuel code generation / recommendation to
the director require it to be completed. **Low risk is unchanged** and still
completes after Verification 1.

## Root cause

The High-only rule was duplicated across five call sites. A frontend-only fix
would not have worked: `complete_verification_2` hard-rejected anything but
High with a 400, so the restored button would have failed on click.

The three backend checks now share a single `_requires_verification_2` helper
rather than repeating the comparison.

## Changes

**Backend** (`ci_application/services.py`)
- Add `VERIFICATION_2_RISK_LEVELS` + `_requires_verification_2`
- Apply it to fuel code generation, Verification 2 completion (error message
  updated to "Medium and High"), and recommend-to-director gating

**Frontend**
- `GovernmentDecisionStep.tsx` — include Medium in the Verification 2 rule,
  which also restores the Risk Assessment and Priority Score fields
- `CIApplicationProgress.jsx` — show the Verification 2 step for Medium,
  accepting the legacy `Moderate` value

## Behaviour change to be aware of

Because Medium now *requires* Verification 2, any in-flight Medium application
that already cleared Verification 1 can no longer generate fuel codes or be
recommended until Verification 2 is completed. Worth confirming with the
business area before release.

## Testing

- Backend: 186 CI application tests pass. Added coverage for Verification 2
  completion (allowed for Medium and High, rejected for Low, requires
  Verification 1 first), generation blocked for Medium until Verification 2 and
  allowed after it, plus a Low-risk regression guard.
- Frontend: 187 CarbonIntensity tests pass across 16 files. Added assertions
  that the Verification 2 action, Risk Assessment, and Priority Score all
  remain visible for Medium after Verification 1.
- Two existing tests encoded the previous High-only rule and were updated to
  the new behaviour.


prv-proton (approved):
Looks good 👍

PR #4757, opened 2026-07-24
Fix: Remove obsolete Fuel Code text - #4718

hamed-valiollahi:
This PR removes obsolete text from affected Fuel Code views

Closes #4718


AlexZorkin (approved):
👍

PR #4763, opened 2026-07-24
Multi-file upload + shared CI upload component + applicant terminology (#4739, #4740, #4743)

AlexZorkin:
Three related tickets delivered in one branch.

## #4739 — Multi-file upload utility

`DocumentTable` (the shared upload component) now accepts multiple files in a
single select or drop action, using the native OS selection (Shift-click,
Ctrl/Cmd-click). Each file is validated and uploaded independently, so existing
size/type rules apply per file. Optimistic rows use functional state updates so
a batch accumulates instead of overwriting, and a file's scanning row now
transitions in place (fixing a latent duplicate-row case). Single-file upload is
unchanged, so every existing consumer (compliance reports, charging sites,
transactions) keeps working.

## #4740 — CI application document upload uses the common component

CI Step 3's upload control previously opened the OS file browser immediately.
It now opens the shared `DocumentUploadDialog` — the common drag-and-drop
component that lists the allowed file types — matching the `ChargingSiteDocument`
pattern. `DocumentUploadDialog` gains a `ci_application` variant + locale.
Existing file validation rules are unchanged, and it inherits the #4739
multi-file support.

## #4743 — "supplier" → "applicant" in the CI process

The CI progress bar now reads **"With applicant"** and **"days with applicant"**
instead of "supplier". The change is limited to the CI application process;
supplier references elsewhere in the portal are untouched. Internal
identifiers/state field names (e.g. `supplierRequestDate`) are intentionally
left as-is to avoid any behaviour change.

## Testing

- 514 frontend tests pass across the Documents, CarbonIntensity, ChargingSite,
  and Transactions suites.
- New coverage: multi-file select + multi-file drop + the `multiple` attribute
  on the input; CI Step 3 opens the shared modal (wired to `ci_application`)
  rather than the OS browser, and stays closed while read-only; the progress bar
  shows applicant wording and no longer shows "With supplier".
- All upload consumers and the CI resume-step logic were re-run to confirm no
  regressions.


prv-proton (approved):
Looks good 👍

PR #4782, opened 2026-07-28
Rebuild credit ledger as a compliance-period ledger (#4714)

AlexZorkin:
## Summary

Closes #4714. The Credit Ledger tab (IDIR and BCeID) is rebuilt from a
paginated all-time grid with a year dropdown into **one ledger page per
compliance period**: previous/next period navigation, a transaction table with
a within-period running balance, Show totals / Show pending toggles, and an
assessed-balance section below the ledger.

## Backend

New endpoint `GET /credit-ledger/organization/{id}/period/{year}` built
directly on `mv_transaction_aggregate` (no migration, no changes to the
materialized views or their refresh triggers):

- Completed transactions only by default; `include_pending=true` adds
  in-flight transactions (transfers Submitted/Recommended, agreements and
  adjustments Recommended), each flagged `isPending`.
- Units In / Units Out derived from the organization's perspective (transfer
  direction; signed units elsewhere), with the running balance accumulating
  within the selected period only.
- Totals grouped by transaction type (in / out / net) plus a grand total.
- Assessed balances for the previous and current compliance year via a
  non-floored March 31 cut-off sum, so deficits surface as negatives.

## Frontend

`CreditLedger.jsx` rebuilt as a plain per-period table: prev/next navigation,
ID / Effective date / Type / Units in / Units out / Running balance columns
(negatives highlighted in red), both toggles, the assessed-balance section,
and the existing Excel export retained. New `usePeriodCreditLedger` hook.

## Two defects found during local verification (both fixed + regression-tested)

1. **Mixed date types in the aggregate view** — `mv_transaction_aggregate`
   emits `date`, tz-aware, and naive `datetime` effective dates; ordering them
   together crashed. Fixed with a normalizing sort key.
2. **"Show pending" was structurally dead for transfers** — pending transfers
   carry `compliance_period = 'N/A'` until recorded, so a period filter never
   matched them. The period query now falls back to the transaction's
   effective-date year for rows without a numeric period.

## Deviations flagged for review

- **Ordering:** transactions are listed in **ascending** effective-date order
  so the running balance accumulates from the start of the period, matching
  the wireframe; the written AC says descending. One-line change if the
  literal AC wording is preferred.
- **Pending compliance reports** are not in `mv_transaction_aggregate` until
  assessed, so the pending toggle surfaces in-flight transfers/agreements but
  compliance reports appear only once assessed — consistent with the rest of
  the system.

## Testing

- Backend: 29 credit_ledger tests, including repo-level integration tests that
  insert source rows and `REFRESH MATERIALIZED VIEW` inside the rolled-back
  test transaction to cover the pending-period fallback and the non-floored
  assessed balance against the real view.
- Frontend: 91 tests across the ledger component, hook, org routing, and
  OrganizationView suites.
- Manually verified in a local environment against a develop-schema database,
  including a rich multi-type scenario exercising negative running balances,
  grouped totals, the pending toggle, and the assessed-balance section.


hamed-valiollahi (approved):
The credit ledger rebuild looks fantastic! 👍

It makes the portal more user friendly and improves the overall experience. I tested it locally with both user types and everything worked as expected. Thank you

PR #4790, opened 2026-07-31
Fix: LCFS - Date picker in operating date field exits after year selection

prv-proton:
Fixes #4768

Date editor changes:

- Prevented the calendar from closing when a user selects a year; it now continues through month and day selection.
- Improved manual date entry so typed dates save correctly on Enter, Tab, blur, picker accept, and clear.
- Avoided unnecessary rewrites when a date cell is focused and left unchanged.

 Autocomplete editor changes:

- Changed Enter behavior to commit the currently highlighted option instead of only toggling the dropdown.
- After selecting with Enter, the grid now advances to the next editable cell and begins editing there, matching Tab behavior.


AlexZorkin (approved):
Looks good 👍

PR #4822, opened 2026-08-04
Fix credit ledger transaction ID prefixes by type

AlexZorkin:
## Problem

The compliance-period credit ledger rendered transaction IDs with a two-way ternary — `CR` for compliance reports, `CT` for everything else. Initiative agreements and administrative adjustments therefore displayed a **correct number under the transfer prefix** (e.g. `CT3113` where it should read `IA3113`).

The numbers themselves were never wrong: `mv_credit_ledger` selects each type's own id (`initiative_agreement_id`, `admin_adjustment_id`, …), so only the prefix was incorrect. The bug was duplicated at both render sites — the transaction list and the grouped-totals view.

## Fix

The correct prefix map already existed, but was private to the transactions grid — which is precisely how the ledger drifted from it. Moved it to `utils/formatters` as `TRANSACTION_ID_PREFIXES` + `formatTransactionId()`, and pointed both consumers at it so they cannot diverge again.

| Type | Prefix |
| --- | --- |
| Transfer | `CT` |
| InitiativeAgreement | `IA` |
| AdminAdjustment | `AA` |
| ComplianceReport | `CR` |
| AggregatorIssuance | `AG` |
| StandaloneTransaction | *(none)* |

Legacy (standalone) transactions stay unprefixed, matching the transactions grid — they carry a raw `transaction` table id rather than a domain id, so the previous code was labelling those `CT` as well.

## Also fixed

The transactions grid sort comparator sliced a **single** character off a **two**-character prefix (`parseInt(value.slice(1), 10)`), so every comparison evaluated to `NaN` and the ID column never actually sorted. Now uses a shared `parseTransactionIdNumber()` that strips a prefix of any length — or none — and falls back to `0` rather than `NaN`.

## Tests

- `_schema.test.js` was a bare `describe.todo()`; it now covers the transaction ID column — prefix per type, unprefixed legacy rows, and a descending-sort assertion that fails against the old comparator.
- `CreditLedgerPeriod.test.jsx` stubbed the entire formatters module, so it would have passed a broken prefix silently. Switched to `importOriginal` so only `dateFormatter` is stubbed, plus assertions that `IA43`/`CT444` render and `CT43` does not — in both the list and totals views.
- Unit tests for both new helpers.

87 passing across the three affected test files; full `views/Transactions` suite at 338 (up from 335). `type-check` clean for all changed files.

Verified via tests rather than a running stack — the assertions cover the exact rendered cell text, which is what was wrong.

## Note

Originally fixed on `release-v1.3.7` (`81215b766`) and cherry-picked here; the diff is byte-identical and the six touched files were identical across both branches.


prv-proton (approved):
👍

PR #4694, opened 2026-07-14
Chore: Automate and backfill release notes - 4663

hamed-valiollahi:
This PR automates production release notes and backfills missing releases.

- Generate release notes on production deployment
- Backfill missing historical release notes
- Keep footer version and release link current

Closes #4663

PR #4824, opened 2026-08-04
Generate release notes before the prod image is built (#4663)

AlexZorkin:
## Problem

The v1.3.7 release notes never reached the Release Notes page. The workflow ran ([run 30936940663](https://github.com/bcgov/lcfs/actions/runs/30936940663)), generated the entry correctly (111 insertions), and then failed on the last step:

```
remote: error: GH006: Protected branch update failed for refs/heads/main.
remote: - Changes must be made through a pull request.
remote: - Cannot change this locked branch
```

Two independent problems, either of which alone would have broken it:

1. **The workflow pushed directly to `main`.** `main` has `lock_branch: true` plus required PR reviews, so `GITHUB_TOKEN` can never push to it. This push could not have succeeded on any release.
2. **Generation ran too late to matter.** The page reads `frontend/public/release-notes.json`, a static asset baked into the frontend image at build time. prod-ci builds that image from `main` when the release PR merges — for v1.3.7 at 17:36, finishing ~17:47. The GitHub release was published at 18:05, 18 minutes later, and the notes commit carried `[skip ci]` to avoid a redeploy. Even a successful push would only have shipped the entry a release late, permanently.

This was the first time the workflow had ever run — v1.3.6 had no GitHub release, so the trigger never fired. Everything through 1.3.6 was hand-backfilled in #4694, which is why the automation looked healthy.

## Changes

**Generate on the release branch, not after the release.** Notes are now generated on pushes to `release-v*`. The entry is written while the release branch is still open, merges to `main` with the rest of the release, and is present in the tree the prod image is built from. `generateReleaseNotes` is called with `target_commitish` since the tag does not exist yet.

**Never push to `main`.** The `release: published`/`edited` trigger is kept as a safety net for releases published without a release branch, but it now commits to `chore/release-notes-<tag>` and opens a PR. It reuses an existing branch when one is already open, so no force-push is needed. When the entry is already present, the run is a no-op and no PR is opened.

**Fix the summary step.** The Models API call was pointed at `models.inference.ai.azure.com`, which returns 404. Every run silently fell back to raw GitHub markdown — so the polished, plain-English summary the page expects was never actually being produced. Now points at `models.github.ai/inference` with the `models: read` permission, keeping the existing graceful fallback.

**Pick the previous tag correctly.** Previous-tag resolution took the highest *other* semantic tag, so a hotfix on an older line would diff against a newer release. It now takes the highest tag strictly below the one being released.

**Warn on untagged releases.** v1.3.6 was deployed but its tag was never pushed, so v1.3.7's generated notes silently covered the whole `v1.3.5...v1.3.7` range and repeated content already published under v1.3.6. The workflow now emits a warning when the newest entry in `release-notes.json` has no matching git tag.

## Also included

The v1.3.7 entry itself, scoped by hand to what actually shipped after the v1.3.6 prod release (the v1.3.6 overlap described above removed).

## Note on rollout

Because the JSON is baked into the image, merging this does not update the live page on its own — the entry appears with the next prod build. #4826 carries these same changes to `main` (bundled with the credit ledger prefix fix) to get v1.3.7 onto the page sooner.

This PR is the `develop` counterpart and is **not** redundant with #4826: the workflow that runs for a release is the one on the release branch, which is cut from `develop`. Without this on `develop`, `release-v1.3.8` would be cut with the broken workflow and fail the same way.

## Verification

- Workflow YAML parses; tag/mode resolution simulated across push, release, `workflow_dispatch`, non-semantic tag, and empty-tag cases.
- Previous-tag resolution checked against the repo's real tag set (`v1.3.7`→`v1.3.5`, `v1.3.3`→`v1.3.2`, correcting the old behaviour).
- PR-body heredoc rendering simulated to confirm no markdown code-block indentation.
- All 20 entries in `release-notes.json` validated against the `Release` interface; `useReleaseNotes` tests pass (6/6).

`ReleaseNotes.test.tsx` could not be executed locally — `@dnd-kit/*` is declared in `package.json` but missing from the local `node_modules`, which is a pre-existing environment issue unrelated to this change.

## Follow-up worth considering

The underlying fragility is that release notes are a build-time asset, so any correction needs a deploy. There is already a `release_notes` API, DB model, and admin override table — widening that to hold the generated entry would decouple the page from image builds entirely. Out of scope here.


hamed-valiollahi (approved):
Thanks, Alex! Appreciate you digging into this and getting it resolved. 🙏

PR #4826, opened 2026-08-04
Fix credit ledger ID prefixes, release notes automation, and ledger assessed balances/envelopes (#4831, #4832)

AlexZorkin:
Hotfix against `main`, covering four separate changes. Consolidates #4823, #4824 and #4825 (supersedes #4823 and #4825; #4824's content is carried here in full), and now also carries the credit ledger work for #4831 and #4832.

> **Scope note for reviewers.** This started as a two-commit hotfix for the ID prefixes and the release-notes workflow. Sections 3–5 are newer and are *behaviour changes*, not cherry-picks: they change how every compliance year is grouped and where assessed balances come from. Please review them as such rather than as part of a routine hotfix.

**Companion PRs — this must not merge alone.** The same three ledger commits go to `develop` in #4843 and to `release-v1.3.7` in #4844. Without #4843, the next release cut from `develop` silently reverts sections 3–5.

---

## 1. Credit ledger transaction ID prefixes (was #4823)

The compliance-period credit ledger rendered transaction IDs with a two-way ternary — `CR` for compliance reports, `CT` for everything else — so initiative agreements and administrative adjustments displayed a **correct number under the transfer prefix** (e.g. `CT3113` where it should read `IA3113`). Only the prefix was wrong; `mv_credit_ledger` already selects each type's own id.

The correct prefix map existed but was private to the transactions grid, which is how the ledger drifted from it. It moves to `utils/formatters` as `TRANSACTION_ID_PREFIXES` + `formatTransactionId()`, with both consumers pointed at it.

| Type | Prefix |
| --- | --- |
| Transfer | `CT` |
| InitiativeAgreement | `IA` |
| AdminAdjustment | `AA` |
| ComplianceReport | `CR` |
| AggregatorIssuance | `AG` |
| StandaloneTransaction | *(none)* |

Also fixes the transactions grid sort comparator, which sliced one character off a two-character prefix and so compared `NaN` on every row, leaving the ID column unsorted.

## 2. Release notes automation (was #4824)

The v1.3.7 notes never reached the Release Notes page. The workflow ran ([run 30936940663](https://github.com/bcgov/lcfs/actions/runs/30936940663)), generated the entry correctly, then failed to push to `main`.

Generation now runs on pushes to `release-v*`, so the entry is written while the release branch is open and merges to `main` in time for the image build (`generateReleaseNotes` gets `target_commitish`, as the tag does not exist yet). The `release` trigger stays as a safety net but opens a PR instead of pushing to `main`, reusing an existing branch so no force-push is needed, and no-ops when the entry is already present. The endpoint and tag resolution are fixed, and the workflow now warns when the newest entry in `release-notes.json` has no matching git tag.

## 3. Assessed balances come from Line 22 (#4831)

The ledger *derived* each year's assessed balance by summing transactions up to a March 31 cut-off. It now reads Line 22 (available compliance unit balance at period end) from the most recent assessed report for that year, so the displayed value is the one that was actually assessed rather than a recalculation that can drift from it.

- **"Most recent" is the highest version in the report chain** — a supplemental supersedes the original, regardless of when either was assessed. Drafts, submitted and unassessed reports are never a source.
- **A year with no assessed report shows nothing.** The API returns `null`, not `0`, and the cell renders blank. A zero there reads as a real end-of-year balance of nil.
- Status matching accepts `Assessed` and `Reassessed`, comparing as text: `Reassessed` exists in `compliance_report_status` but not in `ComplianceReportStatusEnum`, so binding it as an enum value would raise before the query ran.
- The old derived helper is **removed**, not left in place. Leaving it would keep a second source for the same number, which is the thing this issue is trying to eliminate.

## 4. Compliance year envelopes are April – March (#4832)

Transactions were grouped by the `compliance_period` column, effectively a calendar year. They are now grouped by the compliance reporting period.

| | Envelope |
| --- | --- |
| Compliance year Y | April 1 (Y) → March 31 (Y+1) |
| First year with an assessed report | January 1 (Y) → March 31 (Y+1) |

The first year is widened because no earlier envelope exists to hold that January–March, so those transactions would otherwise fall out of the ledger entirely.

**Compliance reports are exempt from the date rule** and stay with the year they report on, whatever date they were assessed — a 2024 report assessed in 2026 is 2024 ledger.

This also subsumes the pending-transfer special case that #4714 needed: pending transfers carry no compliance period until they are recorded, and a date-based envelope picks them up without a separate fallback.

> **Assumption worth confirming.** The issue does not say which side of the year April 1 falls on. This reads it as April 1 (Y) → March 31 (Y+1), because that is the only reading where the "first year also includes January–March" clause does any work, and it matches the existing March 31 (Y+1) assessment cut-off. If the intent was April 1 (Y-1) → March 31 (Y), every transaction shifts a year — it is a one-line change to `compliance_year_envelope`.

## 5. The Excel export matches the screen

The Download Excel button went to the all-time export, which selects by calendar compliance period — so a downloaded ledger could disagree with the screen exactly at the year boundaries where the two now diverge.

A new period export endpoint builds the spreadsheet from `get_period_ledger`, the same call that renders the screen, so the envelope, rows, ordering and running balance come from one place and cannot drift apart. The sheet's columns are the table's columns, ids and type labels included. The **pending toggle is passed through**, which the old export ignored entirely.

The all-time export is unchanged and still serves the legacy grid, which groups by calendar year and whose export therefore already matches it.

**One bug caught while reviewing this:** `openpyxl` refuses timezone-aware datetimes, and `mv_transaction_aggregate` emits a mix of `date`, naive and tz-aware values — so the download failed outright for any organization whose ledger contained one. Effective dates are now reduced to a plain date on the way into the sheet. The regression test builds a real spreadsheet rather than mocking the builder, since a mocked builder is what let it through in the first place.

## Testing

- **Backend:** 199 tests pass. Integration tests run against a real Postgres and the refreshed materialized view — a February transfer lands in the *previous* compliance year, a May transfer starts the next, the first-year envelope picks up February, a pending transfer is caught by date, Line 22 comes from the latest assessed version and not a newer draft, and an unassessed year returns nothing.
- **Frontend:** 372 tests pass, including the blank assessed cell, the id prefixes, and the download following the period and toggle on screen.
- Sections 3–5 have not been exercised in a running stack. Worth checking a real organization at a year boundary, and opening one downloaded sheet, before this ships.

## Also worth a look

#4836 / #4837 (Excel download the entire credit ledger, #4808) also touch the credit ledger export and may need reconciling with the new period endpoint.


prv-proton (approved):
Looks good to me 👍

prv-proton (approved):
Looks good to me👍

PR #4834, opened 2026-08-06
Fix: Advance CI application after step one - 4767

hamed-valiollahi:
This PR advances the CI application after step one submission.

- Advance users to step two
- Prevent unexpected scroll to page top

Closes #4767


prv-proton (approved):
👍

PR #4835, opened 2026-08-06
Feat: LCFS - Allow multiple transport modes and distances

### LINKED ISSUE #4775: LCFS - Allow multiple transport modes and distances
**Describe the task**
Update distance for each selected transport modes to allow multiple km values. Add an input value to selected multi-select fields. (We don't need icons)

Image

**Purpose**
This task supports applications where both transport mode fields uses more than one mode of transport.

**Acceptance Criteria**
- [ ] Users can select more than one transport mode.
- [ ] Users can enter a separate distance for each selected transport mode.
- [ ] The interface clearly associates each distance with the correct mode.

**Additional context**
- This should align with the feedstock transport mode approach.
- Validate keyboard navigation, screen-reader behaviour, and WCAG 2.2 accessibility requirements.
- Test across supported browsers and devices.

prv-proton:
Fixes #4775


backend/lcfs/db/migrations/versions/2026-08-06-10-00_f9a0b1c2d3e5.py
AlexZorkin, line 1: One issue to confirm before this migration goes live: 

Migration backfill can silently lose data, then drops the source columns in the same transaction. We need to be 100% sure of the following.

The backfill is JOIN transport_mode tm ON tm.transport_mode = btrim(mode_name). Any mode name in the CSV that doesn't match a transport_mode row exactly (case, punctuation, a renamed/deleted mode) produces zero rows — and then line ~226 drops feedstock_transport_mode / feedstock_transport_distance, so the original is gone. Those columns were nullable=False, so every existing pathway has data to lose.

frontend/src/components/BCDataGrid/components/Editors/TransportModeDistanceCellEditor.tsx
```diff
@@ -0,0 +1,263 @@
...
+    () => options.map((option) => getOptionLabel(option)).filter(Boolean),
+    [options]
+  )
+  const [selectedModes, setSelectedModes] = useState(() =>
+    normalizeValue(value, fallbackDistance)
+  )
+  const firstInputRef = useRef(null)
+  const rootRef = useRef(null)
+
+  const getCurrentValue = () =>
```
AlexZorkin, line 111: The editor emits distance: ''; the schema requires int → raw 422

getCurrentValue() in the new editor maps blank to ''. TransportModeDistanceSchema.distance: int = Field(..., ge=0) rejects '', and neither union arm matches, so the request 422s at Pydantic. Consequences:

The friendly validation_errors[frontend_key] = "Distance is required." added to _validate_generated_fuel_code_row_values is unreachable for the blank case — the request never gets there. (toUpdatePayload in GeneratedFuelCodesSection.tsx sends the row verbatim.)
AddEditFuelCode.jsx has no client-side distance validation at all, so checking a mode and tabbing past the km field → opaque 422 on save.

FastAPI's 422 detail is an array; getErrorMessage falls back to error.response.data.detail and feeds it to a message string — likely renders as [object Object] or throws.

Pick one contract: either coerce '' → omit client-side and let the row fail row-validation, or make distance: Optional[int] and enforce required-ness in the existing validators.

backend/lcfs/web/api/ci_application/services.py
```diff
@@ -205,13 +201,115 @@ def _json_value(value: Any) -> Any:
...
+    values = value if isinstance(value, list) else [value]
+    names = []
+    for item in values or []:
+        name = _transport_mode_name(item)
+        if name:
+            names.append(name)
+    return list(dict.fromkeys(names))
+
+
+def _transport_mode_details_from_any(value: Any) -> List[Dict[str, Any]]:
```
AlexZorkin, line 243: This method silently drops modes without a distance

_transport_mode_details_from_selection filters on if name and distance is not None, and _from_any returns that filtered list whenever it's non-empty. So [{Truck,100},{Rail,None}] → Rail is dropped entirely, mode and all. This feeds _fuel_code_transport_mode_links and _sync_fuel_code_transport_mode_links, where distance is nullable by design — mixed lists are legitimate there. A user removing one distance loses the mode.

AlexZorkin (changes requested):
The comments in code are blockers that will need to be addressed. The following are additional finds that should be looked at.

**Should fix**
1. Dead legacy keys in the change-log field list
PATHWAY_LOG_FIELDS still lists feedstock_transport_mode / finished_fuel_transport_mode, but those columns no longer exist on Pathway. _pathway_snapshot does getattr(pathway, field, None) → always None, while _pathway_input_snapshot still writes ",".join(...). The supplemental_pathways_updated history record therefore always stores before.feedstock_transport_mode = None / after = "Truck,Rail". It doesn't surface in the change-log UI (that path uses _pathway_snapshot on both sides), but it writes permanently misleading audit rows. Drop the legacy keys, or derive them from the link rows in both snapshot functions.

2. Union[str, TransportModeDistanceSchema] on the input schema is accept-then-reject
PathwayInputSchema accepts bare strings, then _validate_row raises "Transport mode distance is required." for any of them. Just use List[TransportModeDistanceSchema] on the input schema and keep the union on the response/legacy schemas — you'd get a proper per-item Pydantic error instead of a hand-rolled one that fires on the first offender with no index in loc.

3. Four copies of the same normalization
_transport_mode_name + _transport_mode_distance verbatim in ci_application/services.py and fuel_code/services.py
TransportModeDistanceSchema + TransportModeSelection verbatim in ci_application/schema.py and fuel_code/schema.py
normalizeTransportModeDistances in _step2Schema.jsx and again in _schema.tsx
getModeName/getDistance in the editor
One BE helper module and one FE util. This will drift.

4. Speculative key fallbacks
transportMode || transport_mode || mode || label || name || value appears in five places. Only transportMode is ever produced by this codebase. Silent fallbacks like this hide contract drift rather than surfacing it — narrow to the real shapes.

5. Accessibility gaps vs. the issue's stated AC
Issue [#4775](https://github.com/bcgov/lcfs/issues/4775) explicitly lists WCAG 2.2 / keyboard / screen-reader as acceptance criteria. In TransportModeDistanceCellEditor.tsx:149:

Commit is Cmd/Ctrl+Enter only — undiscoverable, and there's no Done/Cancel button. Tab past the last field leaves the popup with no focus trap and no commit.
role="group" on a 460×420 scrolling overlay; role="dialog" + aria-modal, or a <fieldset>/<legend>, matches the interaction better.
Header strings ("Select transport modes and enter distance (km)", "Add a distance for each selected transport mode.", "Enter") are hard-coded English. The PR deletes two keys from carbonIntensity.json and adds none back.
9. // @ts-nocheck on a brand-new .tsx (line 1) — a new file shouldn't opt out of the type system. Type it, or name it .jsx and keep the PropTypes it already has.

**Minor**

Dead param: the editor reads data?.[colDef?.cellEditorParams?.distanceField] (line 70), but no colDef in this PR sets distanceField. fallbackDistance is always undefined.

Redundant popup wiring: useImperativeHandle exposes isPopup/getPopupPosition while the colDefs also set cellEditorPopup/cellEditorPopupPosition. AG Grid 32 uses the colDef; the instance methods are dead. Likewise afterGuiAttached and the mount useEffect both call firstInputRef.current?.focus().

Double normalization: hasCompleteTransportModeDistances(normalizeTransportModeDistances(x)) in validatePathwayRow — the callee normalizes internally.
Error-key collapse: "no mode selected" and "mode selected, distance blank" now both push feedstockTransportMode, so the user gets one message for two different fixes. The issue's third AC is specifically about clearly associating distance with mode.

Asymmetric downgrade: max(distance) collapses per-mode values, and the restored columns come back nullable=True where they were nullable=False. Acceptable for a downgrade, but worth a comment saying so.

Two contracts for one concept: pathway link distance is NOT NULL, fuel-code link distance is nullable. Presumably deliberate (legacy fuel codes have none) — worth a model comment so the next person doesn't "fix" it.

Test coverage: the existing tests were updated to keep passing, but nothing new asserts distances round-trip through update_step2 → link rows, nothing covers the new editor, and nothing covers the migration backfill.

Diff noise: a fair amount of unrelated black/prettier reformatting is mixed in (the errors.append({...}) blocks in schema.py, the reduce in GeneratedFuelCodesSection.tsx, several test files, a stray whitespace fix in repo.py). ~15% of the diff. Not blocking, but it makes the substantive changes harder to find.

AlexZorkin (approved):
Good updates, ty 👍

PR #4845, opened 2026-08-06
Fix: Auto-save Step 4 consultant details on a draft CI application (#4772)

AlexZorkin:
Closes #4772.

Step 4 was the only step of the CI application with no save path. An applicant who typed the optional consultant communication details and then left the draft instead of submitting lost them, and had to retype them when they reopened it.

## What changed

The three consultant fields now **save on blur** and confirm with a toast. The ticket asked for auto-save rather than a "Save and proceed" button, so the block persists as it is filled in, without a button that would read as "submit" on the final step.

A `savedRef` holds the last persisted values, so a blur that changed nothing does not fire a redundant request — or a second toast at the applicant.

## The draft endpoint is deliberately laxer than submit

Behind it is a new `PUT /ci-applications/{id}/step4` taking a draft schema separate from the submit schema:

| | Submit (`CIApplicationStep4Schema`) | Draft auto-save (`CIApplicationStep4DraftSchema`) |
| --- | --- | --- |
| Declarations | required | absent |
| Consultant fields | required when consent given | all optional |
| Email | `EmailStr` | plain string |

The email type is the important one: auto-save fires on blur while the applicant is still typing, and a half-finished address must not `422` and cost them the field. **Submission stays the enforcement point** and still demands a complete, valid consultant block whenever consent is given, so nothing is weakened at the gate that matters.

The endpoint is open to `CI_APPLICANT`, not only signing authorities, because applicants are the ones who fill this block in. Submission remains restricted.

## Consent and status

- **Withdrawing consent wipes the stored values**, mirroring what `submit_application` already does, so a draft never retains consultant details the applicant has un-consented to.
- Saves are refused on anything that is no longer a `Draft`.

## Testing

- **Backend:** 195 CI application tests pass, covering the new service method and endpoint.
- **Frontend:** 193 CarbonIntensity tests pass (16 files), including 16 new ones for the auto-save behaviour, plus 15 `useCIApplication` hook tests.

Not exercised in a running stack — worth typing into the three fields on a real draft, leaving, and reopening it before merge.

## Note

`EditViewCIApplication.jsx` reports a Prettier warning, but that is pre-existing on `develop` and untouched here.


hamed-valiollahi (approved):
The new feature looks great. Thanks for the solid work! 👍

PR #4855, opened 2026-08-10
Fix: Restore pathway changes confirmation and include legacy transactions in the Excel export - 4829 4809

### LINKED ISSUE #4809: LCFS - Excel excludes legacy transactions
**Describe the Bug**
In production, the transaction table download to Excel excludes legacy transactions on the IDIR side.

**Expected Behaviour**
Legacy transactions should appear in the transaction table download to Excel on both the IDIR side and the BCeID side.

**Actual Behaviour**
Legacy transactions are excluded from the Excel download on the IDIR side.

**Implications**
IDIR users may receive incomplete transaction exports from production. This can affect reconciliation, review, and recordkeeping if legacy transactions are missing from the downloaded Excel file.

**Steps To Reproduce**

User/Role: IDIR User

1. Go to 'Transaction Table' in production.
2. Click on 'Download to Excel'.
3. Open the downloaded Excel file.
4. Compare the exported transactions to the transactions displayed in the portal.
5. See error.

### LINKED ISSUE #4829: LCFS - Request pathways changes modal
**Describe the Bug**

The 'Request Pathways Changes' button no longer displays the confirmation modal before initiating the pathway change request process.

Users are able to proceed immediately after selecting the button without receiving the expected confirmation prompt.

**Expected Behaviour**

When a user selects the 'Request Pathways Changes' button, a confirmation modal is displayed.

The modal informs the user of the impacts of requesting pathway changes and requires the user to confirm their intent before proceeding.

The pathway change request process only continues after the user confirms the action.

**Actual Behaviour**

The confirmation modal is not displayed when the 'Request Pathways Changes' button is selected.

The pathway change request process proceeds immediately.

**Implications**

Users may unintentionally initiate a pathway change request without understanding the consequences of the action.

Removal of the confirmation step increases the risk of accidental requests and creates an inconsistent user experience if the confirmation modal was intentionally included as part of the workflow.

**Steps To Reproduce**

User/Role: IDIR Analyst

1. Open a CI Application.
2. Navigate to the pathway review section.
3. Select the 'Request Pathways Changes' button.
4. Observe that no confirmation modal is displayed.

**Expected Result**

- A confirmation modal is displayed when the 'Request Pathways Changes' button is selected.
- The user must explicitly confirm the action before the pathway change request is initiated.
- The pathway change request only proceeds after confirmation.

AlexZorkin:
Two independent bug fixes, one per commit.

Closes #4829
Closes #4809

---

## #4829 — 'Request pathway changes' skipped its confirmation

The button fired immediately, so an analyst could reopen an application for pathway edits with no warning and no way to back out.

Gated behind a confirmation modal, following the request-documentation flow already in the same component: clicking opens the confirm and leaves the button enabled; only confirming performs the request and disables it.

**Tests** — the existing happy-path case now clicks through the modal, plus two new cases: click opens the confirm without firing the request or disabling the button, and cancel leaves the button enabled. 37 passing in `GovernmentDecisionStep.test.jsx`.

---

## #4809 — Excel export dropped legacy transactions

**Root cause: the grid and its export scoped by organisation through different code paths.**

| | organisation filter | `organization_id` to repo | visibility branch |
| --- | --- | --- | --- |
| Grid | applied in the service | `None` | government |
| Export | — | passed through | **supplier** |

The supplier branch restricts non-transfer rows to `Approved`/`Assessed`. Legacy transactions are `Recorded`, so an analyst viewing an organisation saw them in the grid and lost them in the download — exactly the reported symptom.

`export_transactions` now takes an `is_government` flag, and the two government endpoints scope the way the grid does: organisation filter applied as an explicit condition, `organization_id` withheld from the repository so the government branch still applies.

Suppliers were missing the same rows for the same reason, and the issue asks for both sides, so `Recorded` is now accepted in the supplier branch too. That also covers aggregator issuances, which are likewise settled on arrival. Rows remain restricted to the receiving organisation.

**Note for review:** this is a visibility change on the BCeID side. It is additive and organisation-scoped, but worth a careful read rather than a rubber stamp.

**Tests** — `test_transactions_in_have_correct_visibilities` had encoded the bug, asserting a supplier sees 9 transactions; it is 12 with the fix, and the additional rows are exactly `StandaloneTransaction`/`Recorded`. Updated, plus:

- an assertion that a supplier's own legacy transaction is visible;
- an assertion that every non-transfer row a supplier sees is still scoped to their organisation, so widening the status list cannot quietly become an exposure;
- three service tests pinning the call contract — government-with-org withholds `organization_id` and applies an explicit org filter, supplier keeps role visibility, government-all-orgs passes nothing.

40 passing across `lcfs/tests/transaction/`; 148 organization tests unaffected.

---

## Formatting

`black`/`isort` were run and backed out. isort rewrote imports across four files including `transaction/schema.py`, which neither fix touches — roughly 80 lines of churn on a bug fix. `black --check` also flags both changed backend files, but `black --diff` shows every complaint is pre-existing code (compliance-report conditions, trailing whitespace), so these files are not clean on `develop` either. The diff here is substantive changes only.


prv-proton (approved):
👍

PR #4858, opened 2026-08-10
Compose the release notes summary instead of dumping raw markdown (#4663)

AlexZorkin:
`develop` counterpart of #4857 — same commit, cherry-picked. All three files byte-identical to that branch.

The v1.3.7 entry conflicted: `develop` still carried the hand-written entry from #4824, while `release-v1.3.7` and `main` have the one the automation regenerated after v1.3.6 was tagged. Resolved onto the regenerated entry so all three branches converge on the same content. The other 19 entries were already identical — verified entry-by-entry.

## Problem

The "What's in this release" summary on the [live Release Notes page](https://lowcarbonfuels.gov.bc.ca/release-notes) is currently **9,154 characters of raw GitHub markdown**, starting with an HTML comment and running through every `* chore(deps): bump ... by @dependabot[bot] in https://...` line.

The summary was set to whatever the GitHub Models call returned, falling back to the unmodified native notes on failure. That call now returns:

```
410 {"error":{"code":"github_models_retirement_brownout",
     "message":"GitHub Models is temporarily unavailable as part of a scheduled retirement brownout."}}
```

on every run — GitHub Models is being retired. So the fallback *is* the behaviour, and raw GitHub markdown was never an acceptable thing to render on a public government page. (This is distinct from the earlier `404`, which was a dead endpoint host and has already been corrected; the service itself is what is going away now.)

## Fix

**Compose the summary from the parsed sections.** The script already parses everything into features/fixes/security/dependencies/other, so the summary is built from that: a sentence of counts plus up to four highlights. No external service, so it cannot silently degrade the way the Models call did.

Before — 9,154 chars:
```
<!-- Release notes generated using configuration in .github/release.yml at v1.3.7 --> ## What's Changed ### 📦 Dependencies * chore(deps-dev): bump ws from 8.18.3 to 8.21.0 in /frontend by @dependabot[bot] in https://github.com/bcgov/lcfs/pull/4666 * chore(deps): bump axios ...
```

After — 274 chars:
> This release includes 7 new features, 28 bug fixes, 10 dependency updates and 25 other changes. Highlights: Add YoY supply change visualization by fuel type; Add Credit Market login feature flag; Allow Resetting IDIR IA User Roles; Allow system admins to edit release notes.

`--enhanced-notes` becomes optional and is still honoured if a writer is wired up later — but it is ignored when the text handed back is really just the raw notes, so this exact failure cannot recur silently.

**Drop the retired Models step** from the workflow rather than leave it warning on every run. That also removes the `models: read` permission, which is no longer needed.

**Filter two kinds of noise** now visible on the page:
- Release bookkeeping — `Release v1.3.6` (twice) and `Prod Release v1.3.7` are merges of the release process itself, not changes a fuel supplier or ministry analyst would recognise.
- Duplicate bullets, from a change landing more than once (`Wrap organization menu tabs so all sections stay visible` currently appears twice).

`other` goes 29 → 25 as a result. No other section changes.

## Regenerated entry

The v1.3.7 entry is regenerated in this PR so the fix reaches the page with this merge rather than waiting for the next release. The JSON diff touches **only** the v1.3.7 entry — verified entry-by-entry against `HEAD`; the other 19 are untouched.

## Verification

- Ran the fixed script exactly as the workflow invokes it, against live `generate-notes` output for `v1.3.6...v1.3.7`. Output above is the real result, not an illustration.
- Diffed the result field-by-field against what is currently deployed: `version`, `tag`, `releaseUrl`, `fullChangelogUrl` and `contributors` unchanged; `features` 7, `fixes` 28, `dependencies` 10 unchanged; only `summary` and the four filtered `other` items differ.
- Workflow YAML parses; step list and the trimmed `build-release-notes.mjs` invocation confirmed.

## Note

`scripts/` has no test coverage — this script has now failed silently twice (dead endpoint, then retired service), and both times the failure was only visible on the production page. Worth wiring a test for `parseNativeNotes`/`composeSummary`, but it needs a test runner at repo root that does not exist today, so it is out of scope here.


hamed-valiollahi (approved):
Nice improvement, everything looks great from my side! 👍

PR #4860, opened 2026-08-11
Feat: Add YoY fuel supply percentage change - 4807

hamed-valiollahi:
This PR adds percentage change to year-over-year fuel supply data and fixes tooltip visibility.

- Show percentage change by fuel type
- FIx tooltip visibility
- Recalculate values when supply data changes

Closes #4807


AlexZorkin (approved):
👍

PR #4861, opened 2026-08-11
Display CI priority score and risk assessment after verification

### LINKED ISSUE #4797: LCFS - Display priority score and risk assessment as text after VX2
**Description:**
As an IDIR user, I need the priority score and risk assessment to remain visible as read-only text after VX2 so that I can continue to reference the review information after verification is complete.

**Wireframe:**
<link>

Keeping the priority score and risk assessment visible supports review continuity and preserves important application context after verification.

**Acceptance Criteria:**
- [ ] Given I am an authorized IDIR user, when VX2 is complete, then I can view the saved priority score as read-only text.
- [ ] Given I am an authorized IDIR user, when VX2 is complete, then I can view the saved risk assessment as read-only text.
- [ ] Given fuel codes have been generated, when I return to the application, then the priority score and risk assessment remain visible.
- [ ] The read-only values match the values saved before VX2 was completed.
- [ ] The values are displayed according to existing IDIR permissions.

**Development Checklist:**
- [ ] Display the priority score after VX2.
- [ ] Display the risk assessment after VX2.
- [ ] Confirm where the read-only values appear in the application.
- [ ] Confirm permissions for viewing these values.
- [ ] Validate the display at each post-VX2 workflow state.
- [ ] Validate keyboard navigation and responsive display.
- [ ] Validate the display against WCAG 2.2 accessibility requirements.

**Notes:**
Testing identified that the priority score may disappear from the application and remain visible only in the index after fuel codes are generated.

AlexZorkin:
Closes #4797

## Problem

The risk assessment and priority score render only inside the active Verification 1 / Verification 2 panel in `GovernmentDecisionStep`. As soon as the last verification completes that panel unmounts, so both values disappear from the application — they remained readable only from the CI application index grid and the progress tracker's step meta. This matches the behaviour called out in the issue's Notes.

## Change

Frontend only — the detail endpoint already returns `priorityScore`, `preliminaryRiskAssessment`, `verification2PriorityScore` and `verification2RiskAssessment`.

A read-only summary now renders inside the decision panel when neither verification panel is active and Verification 1 has completed:

- Values prefer the Verification 2 pair and fall back to the Verification 1 pair, mirroring the backend's `_effective_priority_score` helper so the text matches what the index shows.
- The heading reads "Verification 2" or "Verification 1" so it is clear which pass produced the values.
- `Medium` is displayed as "Moderate" to match the wording on the radio labels; a value that was never recorded shows "Not recorded".
- Permissions are unchanged — the summary sits inside the existing role-gated panel, so all IDIR users see it and BCeID users do not.
- Markup is a semantic description list (`dl` / `dt` / `dd`). No interactive controls are added, so the tab order is unchanged, and the summary uses the same column-to-row breakpoint as the editable panel.

## Testing

12 new vitest cases in `GovernmentDecisionStep.test.jsx` cover the post-VX2 state, post-fuel-code-generation, Recommended and Completed, the low-risk VX1-only path, the Medium-to-Moderate label, IDIR vs BCeID visibility, and the description-list markup.

`npx vitest run src/views/CarbonIntensity/__tests__` — 16 files, 207 tests passing.

Not exercised against a running stack; the local environment requires an IDIR login.

## Note for reviewers

`npm run lint` currently fails on every file in the repo, independent of this branch: `frontend/.eslintrc.cjs` extends `'@typescript-eslint/recommended'` where ESLint expects `'plugin:@typescript-eslint/recommended'`. Reproduce on an untouched file with `npx eslint src/main.tsx` from `frontend/`. That also means the husky pre-commit lint step is a no-op failure today. Prettier passes on all changed files, and `tsc --noEmit` reports nothing new. Worth fixing separately, since correcting it will surface a backlog of previously-hidden warnings against `--max-warnings 0`.


prv-proton (approved):
Looks good to me 👍

PR #4862, opened 2026-08-11
Feat: LCFS – Add Design Data Yes/No selection to pathways

### LINKED ISSUE #4798: LCFS - Add Design Data Yes/No selection to pathways
**Description:**
As a BCeID applicant, I need pathway data to include a 'Design Data' Yes/No selection so that the system can correctly determine whether an operational data date range is required.

**Wireframe:**
https://preview.uxpin.com/7264e34749e7af420559b8cfc0ca8001c76db3ec#/pages/222448667

Operational data requires a date range, while design data does not. The selection will allow the portal to apply the correct date range requirements for the type of data being submitted.

**Acceptance Criteria:**
- [ ] Given I am completing pathway data, when I select 'Yes' for 'Design data', then the Operational data date range is optional.
- [ ] Given I am completing pathway data, when I select 'No' for 'Design data', then the Operational data date range is mandatory.
- [ ] Given I change the selection from 'No' to 'Yes', when validation runs, then the operational data date range is no longer treated as mandatory.
- [ ] Given I change the selection from 'Yes' to 'No', when the date range is incomplete, then the system displays the applicable validation message.
- [ ] The selected value is saved and remains visible when the user returns to the pathway.
- [ ] The same validation logic applies in authorized BCeID and IDIR workflows.

**Development Checklist:**
- [ ] Add the 'Design Data' Yes/No Boolean field.
- [ ] Add the 'Design Data' Yes/No selection to pathways.
- [ ] Update operational data date range validation.
- [ ] Confirm whether a default value is permitted.
- [ ] Confirm how existing pathway records are handled.
- [ ] Validate the control for BCeID and IDIR users.
- [ ] Validate keyboard navigation and screen reader support.
- [ ] Validate responsive and supported browser behaviour.
- [ ] Validate the control against WCAG 2.2 accessibility requirements.

**Notes:**
Testing indicated that design data should be identified through a Yes/No selection and that the operational data date range should only be mandatory when design data is not used.

prv-proton:
Fixes #4798


AlexZorkin (changes requested):
One thing to review:

services.py:933 sets expiration_date=pathway.operating_data_to when minting the draft FuelCode. That column is nullable, so nothing crashes — but a design-data pathway now produces a fuel code with expiration_date = NULL, which also means the expiry-notification scheduler will never fire for it. Should probably review how the fallback dates work.

AlexZorkin (approved):
Good to go 👍

PR #4865, opened 2026-08-11
Fix FSE map showing valid BC locations as outside BC

### LINKED ISSUE #4852: LCFS - FSE Map Incorrectly Displays "Outside of BC" Pin Colours for Valid BC Locations
**Describe the Bug**
Within the Compliance Report FSE Map, some Fuel Supply Establishments (FSEs) that are physically located in British Columbia are displaying the pin colour associated with "Outside of BC". Multiple examples have been identified across different organizations and reports, suggesting this is a systemic issue rather than a data issue affecting a specific supplier. 

One example of such records can be viewed in compliance report ID 3453

**Expected Behaviour**
FSE map pins should display the correct colour based on the FSE's validated geographic location. FSEs located within British Columbia should be displayed using the appropriate in-BC pin colour.

**Actual Behaviour**
Certain FSEs located within British Columbia are displayed with the "Outside of BC" pin colour on the Compliance Report FSE Map.

**Implications**
- Users may incorrectly interpret FSE locations.
- Compliance reviewers may spend additional time validating locations manually.
- Reduces confidence in the map visualization and location validation process.
- May result in inaccurate reporting or compliance review decisions.

**Steps to Reproduce**
1. Open a Compliance Report containing FSE records.
2. Navigate to the FSE Map view.
3. Review FSE pin colours.
4. Observe instances where known BC-based FSEs are displayed using the "Outside of BC" pin colour.

**Additional Context**
The issue appears to be related to the Compliance Report map visualization or location classification logic rather than the underlying FSE records themselves, as the same FSEs do not exhibit incorrect status indicators in the FSE Maintenance System.

AlexZorkin:
Closes #4852

## Root cause

Two compounding defects in the BC boundary check (`backend/lcfs/services/geocoder/client.py`), confirmed against the live Nominatim API rather than inferred:

**1. The failure response was read as a success.** `_nominatim_reverse_geocode` only treated a falsy body as an error. Nominatim answers unresolvable coordinates with **HTTP 200** and a body of `{"error":"Unable to geocode"}` — truthy — so it returned `success=True` carrying an `Address` whose every field was `None`.

**2. `check_bc_boundary` then failed in the wrong direction.** It tested `"british columbia" in (province or "").lower()`. An empty province is `False`, and because `result.success` was `True` the bounding-box fallback was never reached, so an *unresolvable* coordinate was reported as *outside BC* rather than *unknown*. The same substring test also rejects the `BC` and `CA-BC` spellings other providers return — the BC Geocoder path populates `province` from `provinceCode`.

Live probes used to confirm this:

```
lat=49.2827&lon=-123.1207  ->  200, address.state = "British Columbia", ISO3166-2-lvl4 = "CA-BC"
lat=0&lon=0                ->  200, {"error":"Unable to geocode"}
```

Reverse-geocode results are cached by coordinate for an hour, which is why the same sites stay wrong across different organizations and reports — matching the "systemic rather than a data issue" observation in the ticket.

A second contributor sits in the map itself. `GeoMapping.jsx` keyed geofencing results by equipment id (`chargingEquipmentId_serialNumber`) even though geofencing is a property of the coordinates, so one FSE could inherit another's verdict when the same id appeared at two positions. `MapComponents.jsx` then chose the pin with `if (!isInBC)`, which makes a *missing* result indistinguishable from a genuine out-of-province one.

## Change

Backend:
- Treat an `{"error": ...}` body, or a response carrying no address details, as a failed lookup.
- Read the province from `state`, then `province`, `region`, and the ISO subdivision code.
- Accept every spelling of BC (`BC`, `B.C.`, `CA-BC`, `British Columbia`, `Colombie-Britannique`).
- Only trust the lookup when it actually resolved a province; otherwise fall back to the bounding box instead of assuming out of province. A resolved non-BC province still wins over the bounding box, so Calgary stays out.

Frontend:
- Key geofencing results by coordinate (`getCoordinateKey`), so a result can never be attributed to the wrong pin.
- Use the Outside of BC pin only for an explicit `false`; an unresolved location stays grey.
- Give the grey pin a legend entry in both states, since it is no longer shared with out-of-province locations.

## Testing

Backend — 56 geocoder tests pass. The 9 new cases were checked against the unfixed client to confirm they are not vacuous: 8 fail there, and the one that passes is the `"British Columbia"` spelling that already worked.

```
python -m pytest lcfs/tests/services/geocoder -q   # 56 passed
```

Frontend — 196 tests pass across `src/views/FinalSupplyEquipments`. The existing `MapComponents` marker tests keyed results by equipment id and were updated to the coordinate keying.

```
npx vitest run src/views/FinalSupplyEquipments   # 10 files, 196 tests
```

Not exercised in a browser; the FSE map needs an IDIR login and a report carrying FSE records.

## Worth checking before this ticket closes

`transformApiData` does `parseFloat(row.latitude) || 0`, so an FSE with a missing or unparseable latitude plots at `0,0` — which is exactly the ocean coordinate Nominatim cannot geocode. Those pins will now render grey rather than red, but they are still plotted off the coast of Africa. If some of the bad pins in report 3453 are that case, it is a separate data/validation problem rather than the boundary check, and worth confirming against the report's rows.

## Note for reviewers

`npm run lint` fails on every file in the repo, independent of this branch: `frontend/.eslintrc.cjs` extends `'@typescript-eslint/recommended'` where ESLint expects `'plugin:@typescript-eslint/recommended'`. Separately, `GeoMapping.jsx` and `MapComponents.jsx` already fail Prettier on `develop`, so they were left unformatted here rather than reformatted wholesale, which would have buried the fix. Black reports the backend changes clean.


hamed-valiollahi (approved):
Nice, looks good to me 👍 Thanks for sorting that out!

PR #4870, opened 2026-08-12
Fix: Auto-save CI assessment values - 4796

### LINKED ISSUE #4796: LCFS - Auto-save priority score and risk assessment
**Describe the Task**
Priority score and risk assessment values only save when Verification 1 or 2 buttons are clicked. They do not save when an IDIR user exits the application or selects workflow actions such as requesting further documentation or pathway changes.

**Expected Behaviour**
Priority score and risk assessment values should auto-save when entered or changed. The user should not be required to complete a verification action to save the values.

Saved values should remain available when the user changes views, exits and returns to the application, or performs another workflow action.

**Actual Behaviour**
Priority score and risk assessment values may be erased when exiting the application, requesting further documentation, or requesting pathway changes.

**Implications**
Analysts may lose review information and need to re-enter values. Lost information may affect application tracking, review continuity, and handoff between analysts.

**Steps To Reproduce**

User/Role: IDIR Analyst

1. Go to 'Carbon Intensity Application'.
2. Enter or change the 'Priority Score' or 'Risk Assessment'.
3. Exit the application or select 'Request Further Documentation' or 'Request Pathway Changes'.
4. Return to the application.
5. See that the entered value was not retained.

hamed-valiollahi:
This PR auto-saves CI application assessment values.

- Auto-save priority and risk assessment
- Available to IDIR analysts

Closes #4796


backend/lcfs/web/api/ci_application/schema.py
```diff
@@ -523,6 +523,11 @@ class CIApplicationVerification2Schema(BaseSchema):
     priority_score: Optional[StrictInt] = Field(default=None, ge=1, le=999)
 
 
+class CIApplicationRiskAssessmentDraftSchema(BaseSchema):
+    preliminary_risk_assessment: Optional[CIRiskAssessmentEnum] = None
+    priority_score: Optional[StrictInt] = None
```
AlexZorkin: do we need to add the bounds protection like the above class? ge=1 le=999?

backend/lcfs/web/api/ci_application/services.py
```diff
@@ -1373,6 +1374,38 @@ async def assign_analyst_to_application(
...
         return await self._to_full_schema_with_user(ci)
 
+    @service_handler
+    async def update_risk_assessment_draft(
+        self,
+        ci_application: CIApplication,
+        data: CIApplicationRiskAssessmentDraftSchema,
+        user: UserProfile,
+    ) -> CIApplicationSchema:
+        self._require_submitted_workflow(ci_application)
```
AlexZorkin: No verification-state guard. _require_submitted_workflow only checks status. If verification_2_date is already set but status is still Submitted, the else branch overwrites preliminary_risk_assessment/priority_score — i.e. the completed verification-1 record. The frontend guards this; the API doesn't.

AlexZorkin (changes requested):
a couple things to review:

- No backend tests for update_risk_assessment_draft.
- Autosave races: two in-flight PUTs both call setQueryData(detail) on success, so an out-of-order response can restore stale values.
- Every debounce tick writes update_user + action_type = UPDATE.

AlexZorkin (approved):
great updates ty 👍

PR #4885, opened 2026-08-17
Feat: Allow renaming uploaded files - 4819

### LINKED ISSUE #4819: LCFS - Add a rename function to the file upload utility
**Describe the task**
Add a rename function to the file upload utility that allows users to update the display name of uploaded files directly from the uploaded file list without changing the original file.

**Purpose**
Users may upload files with unclear, inconsistent, or system-generated names. Allowing the display name to be updated improves usability and helps reviewers identify documents more easily while preserving the original uploaded file.

**Acceptance Criteria**
- [ ] Users can rename an uploaded file from the uploaded file list.
- [ ] Updating the display name does not modify the original file content or file extension.
- [ ] The updated display name is displayed everywhere the file appears within the application.
- [ ] Users can save the updated display name without re-uploading the file.
- [ ] Existing file permissions and visibility rules continue to apply.
- [ ] The system validates invalid characters and prevents duplicate display names where required by business rules.
- [ ] The rename functionality is accessible by keyboard and meets WCAG 2.2 accessibility requirements.

**Additional context**
- The rename function should be available from the uploaded file list.
- The feature should update the user-facing display name only.
- Original file metadata should be retained for audit and recordkeeping purposes.
- Consider using the common file upload utility so the functionality is available consistently throughout the portal.

hamed-valiollahi:
This PR allows users to rename uploaded file display names.

- Allow renaming from uploaded file lists
- Preserve original files and extensions

Closes #4819


backend/lcfs/services/s3/client.py
AlexZorkin: Blocker: IDOR in rename_file. s3/client.py does db.get_one(Document, document_id) and never checks the document is actually associated with parent_id/parent_type. The view validates access to the parent, so any CI applicant who owns application #5 can rename any document in the system by passing an arbitrary document_id. delete_file at least resolves through the association table; rename needs the same scoping.

AlexZorkin (changes requested):
One blocker to address then a few suggestions here to review:

- Allowlist duplicated in s3/client.py and frontend/src/constants/common.ts. Two sources of truth.
- _normalize_display_name uses os.path.splitext on the requested name, so a name like 2026.01 summary becomes 2026.pdf. Only bites when the original file has no extension, but it silently discards input.
- Duplicate-name check is a read-then-write race with no DB constraint.
- RenameableFileName uses defaultProps — deprecated for function components in React 18.3+.

AlexZorkin (approved):
thanks for the updates 👍

PR #4889, opened 2026-08-18
Fix: Updated FSE Records Are Excluded from Compliance Reports

### LINKED ISSUE #4884: LCFS - Updated FSE Records Are Excluded from Compliance Reports After Re-Validation
**Describe the Bug**
There appears to be an issue with Final Supply Equipment (FSE) records that are updated through the maintenance system and subsequently re-validated by an analyst. After the update and validation process, the FSE record's kWh usage appears to be reset to zero in a submitted compliance report. Additionally, these updated FSE records may be excluded altogether from the compliance report despite still appearing in the FSE edit page as included and checked marked.

**Expected Behaviour**
FSE records that are updated in the maintenance system and re-validated by an analyst should retain their correct kWh usage values and continue to be included in the compliance report if they remain marked as included. The FSE edit page and the main compliance report should display consistent FSE counts and totals. This remains editable until the report is recommended by an analyst.

**Actual Behaviour**
After an FSE record is updated and re-validated:
- The kWh usage appears to be reset to zero.
- The affected FSE record continues to appear in the compliance report FSE edit page.
- The record appears as checked and included in the report.
- The affected FSE record is excluded from the main compliance report view and associated totals.
- The FSE edit page and the main compliance report display inconsistent FSE counts (e.g., 30 FSE records visible in the edit page, but only 20 appearing on the main report page).

**Implications**
- Compliance reports may contain incomplete or inaccurate Final Supply Equipment data.
- FSE totals displayed on the main report may not match the records shown on the FSE edit screen.
- Users may incorrectly assume all selected FSE records are being included in report calculations.
- The issue is particularly problematic for original compliance reports, where analysts do not want to require a supplemental report simply to correct or update Final Supply Equipment information.
- Updated and re-validated FSE records are currently being excluded from the report altogether, despite appearing as included within the FSE edit page.
- For supplemental reports, there is at least a temporary workaround whereby the analyst can return the report and request that the BCeID user re-enter the kWh usage and resubmit. However, this workaround is not available or appropriate for original reports.
- Compliance reporting outputs may be impacted if qualifying FSE records are unintentionally excluded.
- The issue creates uncertainty regarding whether the root cause is the kWh value being reset to zero, the re-validation process, or the report inclusion logic.

**Steps To Reproduce**

User/Role: BCeID User and Analyst

1. Go to the maintenance system and update an existing Final Supply Equipment record that is part of a submitted supplemental compliance report.
2. Submit the updated Final Supply Equipment record.
3. Have an analyst review and validate the updated record.
4. Open the associated compliance report.
5. For a supplemental return to draft, navigate to the FSE edit page and confirm the updated FSE record appears selected/included.
6. Return to the main compliance report view.
7. Observe that the updated FSE record is no longer included in the displayed FSE records and report totals.
8. Verify whether the affected FSE record's kWh usage has been reset to zero following re-validation.
9. Compare the total number of FSE records shown in the edit page versus the main compliance report page.

prv-proton:
Fixes #4884


AlexZorkin (approved):
solid fix 👍

AlexZorkin (approved):
Looks good 👍

PR #4906, opened 2026-08-20
Fix: Approved Current/Archived Carbon Intensities Table Fuel Code Filter Returns Network Error

### LINKED ISSUE #4881: LCFS - Approved Current/Archived Carbon Intensities Table Fuel Code Filter Returns Network Error
**Describe the Bug**
The "Fuel code" column filter appears to be broken on both the Approved Carbon Intensities - Current page and the Approved Carbon Intensities - Archived page. When a user enters a value into the Fuel Code filter field, the application returns a "Network Error" message instead of filtering the results.

**Expected Behaviour**
Users should be able to enter a fuel code or partial fuel code into the Fuel Code filter field and receive filtered results without errors on both the Current and Archived Approved Carbon Intensities pages.

**Actual Behaviour**
Entering any value into the Fuel Code filter field causes the page to display a "Network Error" message and prevents the filter from returning results.

Image

**Implications**
- Users are unable to filter Approved Carbon Intensities records by fuel code.
- Searching for specific fuel pathways becomes more difficult and time-consuming.
- The issue affects both Current and Archived Approved Carbon Intensities data.
- The error may indicate a problem with the filter query, API request, or backend processing associated with the Fuel Code field.
- User confidence may be reduced when the application returns an unexpected error message.

**Steps To Reproduce**

User/Role: Any user with access to Approved Carbon Intensities

1. Go to the Approved Carbon Intensities - Current page or Approved Carbon Intensities - Archived page.
2. Locate the "Fuel code" column filter.
3. Enter any value into the filter field.
4. Observe that a "Network Error" message is displayed instead of filtered results.
5. Confirm the same behaviour occurs on both pages.

prv-proton:
Fixes #4881

### DIFF
--- a/backend/lcfs/tests/fuel_code/test_fuel_code_repo.py
+++ b/backend/lcfs/tests/fuel_code/test_fuel_code_repo.py
@@ -443,6 +443,41 @@ def test_get_fuel_code_bulletin_pagination_params_accepts_fuel_code_filter(
     assert "%bclcf%" in compiled_condition
 
 
+def test_get_fuel_code_bulletin_pagination_params_filters_effective_date_as_date(
+    fuel_code_repo,
+):
+    pagination = PaginationRequestSchema.model_validate(
+        {
+            "page": 1,
+            "size": 25,
+            "filters": [
+                {
+                    "field": "effectiveDate",
+                    "filterType": "date",
+                    "type": "equals",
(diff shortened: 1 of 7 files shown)


AlexZorkin (approved):
Everything checks out 👍

PR #4928, opened 2026-08-21
Fix: Dev bugs contact name and multi-transport modes.

prv-proton:
- Multi-Transport modes data persistance for multiple rows in case of generated fuel codes.
- contact Name changes fix.


AlexZorkin (approved):
Looks good 👍

PR #4937, opened 2026-08-24
Fix: Investigate bug in summary section of report 2772

### LINKED ISSUE #4866: LCFS - Investigate bug in summary section of report 2772
**Describe the Bug**  
The Low Carbon Fuel Target Summary for Supplemental Compliance Report 2772 is displaying incorrect assessment carry-forward values.

The original report was not assessed. However, the supplemental report shows:
- Line 15: 212 compliance units previously issued
- Line 20: 0 compliance units as the balance change from assessment

This results in an incorrect value on Line 22.

Despite the summary displaying these values, the organization still received a transaction for the correct issuance amount of 212 compliance units (CR2772).

**Expected Behaviour**  
If the original report was never assessed, the supplemental report should not display previously issued units or assessment-related balance changes that imply a prior assessment occurred.

Line 15, Line 20, and the resulting Line 22 values should accurately reflect the report's assessment history and issuance records.

**Actual Behaviour**  
The supplemental report displays:
- 212 units previously issued on Line 15
- 0 units on Line 20 as a balance change from assessment

These values appear inconsistent with the report's assessment status and result in an incorrect calculation on Line 22.

**Implications**  
- Users may see incorrect compliance unit balances in the report summary.
- Assessment history may be represented inaccurately.
- Report calculations and auditability may be impacted. Check 2025 impact too?
- Creates confusion when transaction records show a correct issuance while the report summary does not align.

**Steps to Reproduce**  
1. Open Supplemental Compliance Report 2772.
2. Navigate to the Low Carbon Fuel Target Summary.
3. Review values on Lines 15, 20, and 22.
4. Compare the displayed values against the original report's assessment status and the compliance unit transaction record (CR2772).

**Additional Context**  
The discrepancy appears isolated to the report summary calculations or assessment carry-forward logic, as the associated compliance unit transaction was created for the correct issuance amount of 212 units.

prv-proton:
Fixes #4866

For historical/government-initiated supplementals when the original same-period report was never assessed and if there are no previously issued units to carry forward, stored Line 15/16 values from the prior submitted version will be turned 0.

### DIFF
--- a/backend/lcfs/tests/compliance_report/test_summary_service.py
+++ b/backend/lcfs/tests/compliance_report/test_summary_service.py
@@ -3103,6 +3103,53 @@ async def test_line_15_16_zero_when_no_assessed_report_for_supplemental(
 
 
 @pytest.mark.anyio
+@pytest.mark.parametrize("compliance_year", ["2024", "2025"])
+async def test_locked_supplemental_summary_recomputes_carry_forward_when_no_assessed_report(
+    compliance_report_summary_service,
+    mock_repo,
+    compliance_year,
+):
+    """
+    Locked supplemental reports without an assessed same-period baseline should
+    not display previous-version issued units as Line 15/16 carry-forward.
+
+    This covers historical/migrated records such as CR2772 where the final
+    transaction was created for the correct issuance, but the stored summary
+    still showed Line 15 = current issuance and Line 20 = 0.
+    """
+    summary = make_summary(locked=True)
+    summary.summary_id = 1
+    summary.compliance_report_id = 2772
+    summary.line_15_banked_units_used = 212
+    summary.line_16_banked_units_remaining = 0
+    summary.line_17_non_banked_units_used = 1000
+    summary.line_18_units_to_be_banked = 212
+    summary.line_19_units_to_be_exported = 0
+    summary.line_20_surplus_deficit_units = 0
+    summary.line_21_non_compliance_penalty_payable = 0
+    summary.line_22_compliance_units_issued = 1000
+
+    report = make_report(1, ComplianceReportStatusEnum.Assessed, compliance_year)
+    report.summary = summary
+    report.compliance_report_id = 2772
+
+    mock_repo.get_compliance_report_by_id = AsyncMock(return_value=report)
+    mock_repo.get_assessed_compliance_report_by_period = AsyncMock(return_value=None)
+
+    result = (
+        await compliance_report_summary_service.calculate_compliance_report_summary(
+            report_id=2772
+        )
+    )
+
+    line_values = _get_line_values(result.low_carbon_fuel_target_summary)
+    assert line_values[15] == 0
+    assert line_values[16] == 0
+    assert line_values[20] == 212
+    assert line_values[22] == 1212
+
+
+@pytest.mark.anyio
 async def test_line_15_16_uses_assessed_report_values(
     compliance_report_summary_service,
     mock_repo,
--- a/backend/lcfs/web/api/compliance_report/summary_service.py
+++ b/backend/lcfs/web/api/compliance_report/summary_service.py
@@ -162,6 +162,67 @@ class ComplianceReportSummaryService:
         )
         return prev_compliance_report is not None
 
+    async def _normalize_locked_supplemental_without_assessed_baseline(
+        self,
+        summary: ComplianceReportSummarySchema,
+        compliance_report: ComplianceReport,
+    ) -> None:
+        """Correct stale low-carbon carry-forward rows for locked supplementals.
+
+        Some historical/government-initiated supplementals can exist even when
+        the original same-period report was never assessed. In that case there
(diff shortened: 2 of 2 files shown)


backend/lcfs/web/api/compliance_report/summary_service.py
```diff
@@ -162,6 +162,67 @@ async def _should_lock_lines_7_and_9(
...
+        the original same-period report was never assessed. In that case there
+        are no previously issued units to carry forward, so stored Line 15/16
+        values from the prior submitted version must not affect the displayed
+        balance.
+        """
+        report_version = getattr(compliance_report, "version", None)
+        if not isinstance(report_version, int) or report_version <= 0:
+            return
+
+        assessed_report = await self.cr_repo.get_assessed_compliance_report_by_period(
```
AlexZorkin: get_assessed_compliance_report_by_period returns the latest assessed report in the org and period excluding self, with no version constraint. Once a later version of 2772 is assessed, this finds it, returns early, and v1 goes back to showing 212 / 0 / 1000. Constrain to version < current within the same compliance_report_group_uuid.

backend/lcfs/web/api/compliance_report/summary_service.py
```diff
@@ -162,6 +162,67 @@ async def _should_lock_lines_7_and_9(
...
+        if not rows_by_line:
+            return
+
+        line_17 = rows_by_line.get(17).value if rows_by_line.get(17) else 0
+        line_18 = rows_by_line.get(18).value if rows_by_line.get(18) else 0
+        line_19 = rows_by_line.get(19).value if rows_by_line.get(19) else 0
+        line_20 = int(line_18 or 0) + int(line_19 or 0)
+        line_22 = max(int(line_17 or 0) + line_20, 0)
+
+        for line, value in ((15, 0), (16, 0), (20, line_20), (22, line_22)):
```
AlexZorkin: This only mutates the response schema; the stored line_15_banked_units_used and line_20_surplus_deficit_units stay wrong, so anything reading the table directly still sees the inconsistency. Can we fix the values themselves here?

backend/lcfs/web/api/compliance_report/summary_service.py
```diff
@@ -162,6 +162,67 @@ async def _should_lock_lines_7_and_9(
...
+            if row.line is not None
+        }
+        if not rows_by_line:
+            return
+
+        line_17 = rows_by_line.get(17).value if rows_by_line.get(17) else 0
+        line_18 = rows_by_line.get(18).value if rows_by_line.get(18) else 0
+        line_19 = rows_by_line.get(19).value if rows_by_line.get(19) else 0
+        line_20 = int(line_18 or 0) + int(line_19 or 0)
+        line_22 = max(int(line_17 or 0) + line_20, 0)
```
AlexZorkin: Line 22 is recomputed as line_17 + line_20, but the calculator uses available_balance + deferred_prior_issuance + line_20. This diverges if the group has assessed adjustment transactions.

backend/lcfs/web/api/compliance_report/summary_service.py
```diff
@@ -162,6 +162,67 @@ async def _should_lock_lines_7_and_9(
...
+
+        for line, value in ((15, 0), (16, 0), (20, line_20), (22, line_22)):
+            if line in rows_by_line:
+                rows_by_line[line].value = value
+
+        if 21 in rows_by_line:
```
AlexZorkin: Row 21 of low_carbon_fuel_target_summary is rewritten, but non_compliance_penalty_summary row 21 and the total built at 367-382 aren't. Update both, or limit the shim to lines 15/16/20/22.

backend/lcfs/tests/compliance_report/test_summary_service.py
```diff
@@ -3102,6 +3102,53 @@ async def test_line_15_16_zero_when_no_assessed_report_for_supplemental(
     assert line_16.value == 0, f"Line 16 should be 0 when no assessed report, got {line_16.value}"
 
 
+@pytest.mark.anyio
```
AlexZorkin: Please add the "later version assessed" case and a penalty-section assertion; only the happy path is covered.

AlexZorkin (changes requested):
The core logic is there but need to make sure we're patching the core data not just the view response.

prv-proton:
> The core logic is there but need to make sure we're patching the core data not just the view response.

- Added a same-group, prior-version assessed baseline lookup in repo.py, so later assessed versions no longer suppress correction for earlier versions like v1.
- Added a prior-version-scoped deferred adjustment lookup in transaction/repo.py
- Changed summary_service.py so the correction updates the stored ComplianceReportSummary columns, flushes them, and then mirrors those corrected values into the response.
- Line 22 now includes deferred_prior_issuance, matching the calculator.
- Penalty Line 21 and total_non_compliance_penalty_payable are updated in both stored data and response rows.

AlexZorkin (approved):
Good updates thank you 👍

PR #4970, opened 2026-08-26
Fix: Persist priority score - 4796

### LINKED ISSUE #4796: LCFS - Auto-save priority score and risk assessment

hamed-valiollahi:
This PR ensures priority scores remain saved correctly. Closes #4796

Please refer to the issue reported by the Product Owner.
> This still isn't autosaving the way we had hoped in Test. In dev, it is working to a certain extent, but still reverts back to the originally saved priority score and risk assessment if I leave the page or refresh. I think we are hoping that it will auto save in the UI on entry and within the breadcrumb area of what stage the application is at. In dev, it looks like this if I navigate away from the page after changing the priority score to 200 and refresh

image

### DIFF
--- a/backend/lcfs/tests/ci_application/test_ci_application_service.py
+++ b/backend/lcfs/tests/ci_application/test_ci_application_service.py
@@ -2263,7 +2263,7 @@ async def test_request_documentation_enables_upload_for_workflow_roles(
 
 
 @pytest.mark.anyio
-async def test_request_documentation_clears_previous_government_review(
+async def test_request_documentation_resets_review_and_preserves_latest_assessment(
     service, repo, mock_user
 ):
     mock_user.role_names = {RoleEnum.ANALYST}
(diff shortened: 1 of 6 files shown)


AlexZorkin (changes requested):
Review the copilot suggestions.

hamed-valiollahi:
Thanks for the review! Both issues are now fixed. 👍

AlexZorkin (approved):
Thanks for the updates 👍

hamed-valiollahi:
Appreciate the review.

